PR #2, opened 2015-09-28
Add storage samples

bshaffer:


storage/app.php
```diff
@@ -0,0 +1,197 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Follow the instructions on https://github.com/google/google-api-php-client
+ * to download, extract and include the Google APIs client library for PHP into
+ * your project.
+ */
+require_once 'google-api-php-client/autoload.php';
```
tmatsuo: Is the client library available via Composer?

storage/app.php
```diff
@@ -0,0 +1,197 @@
...
+ * @param string $apiResponse The API response to process.
+ * @return string Markup for the specific Google Cloud Storage API request.
+ */
+function generateMarkup($apiRequestName, $apiResponse) {
+  $apiRequestMarkup = '';
+  $apiRequestMarkup .= "<header><h2>" . $apiRequestName . "</h2></header>";
+
+  if ($apiResponse['items'] == '' ) {
+    $apiRequestMarkup .= "<pre>";
+    $apiRequestMarkup .= print_r(json_decode(json_encode($apiResponse), true),
```
tmatsuo: Why json_encode then json_decode? What is the purpose of this?

storage/app.php
```diff
@@ -0,0 +1,197 @@
...
+?>
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8">
+  </head>
+  <body>
+    <header><h1>Google Cloud Storage Sample App</h1></header>
+    <div class="main-content">
+      <?php if(isset($listBucketsMarkup)): ?>
```
tmatsuo: It looks like good old days PHP programs. What about to rewrite this with silex?

tmatsuo:
I put some comments. Also can you write some tests?

storage/app.php
bshaffer: wow, there is no purpose and this is hilarious :D

storage/app.php
bshaffer: It's okay to do this, honestly. We could rewrite to use silex / a templating language, but PHP functions just fine as a templating language in itself. It's probably more trouble than it's worth.

bshaffer:
@tmatsuo this is just the initial merge from https://github.com/GoogleCloudPlatform/storage-getting-started-php. We can improve upon it (tests, composer) later but I think for now it's okay to just get it consolidated

storage/app.php
bshaffer: Yes, see [here](https://packagist.org/packages/google/apiclient). I agree it should be updated, especially once we move to the `2.0` version.

tmatsuo:
@bshaffer alright. Then I'll wait for the fix on json_encode & json_decode. Please file one or two bugs for bringing this sample up to our standard. Feel free to take those bugs or assign them to me.

storage/app.php
bshaffer: There could be some hackish reason for this. I think we should not worry about it until we have time to actually test. I will file a bug for it.

bshaffer:
see #3, #4, and #5

PR #7, opened 2015-10-08
Add a test.  Ran php-cs-fixer.  Add a .travis.yml file.  Add encrypte…

SurferJeffAtGoogle:
…d credentials.


.travis.yml
```diff
@@ -0,0 +1,30 @@
...
+language: php
+php:
+  - 5.5
+  - 5.6
+  - hhvm
+  - nightly
+
+before_install:
+  # Decrypt the credentials we added to the repo using the key we added with the Travis command line tool
+  - openssl aes-256-cbc -K $encrypted_a66730e1b199_key -iv $encrypted_a66730e1b199_iv -in credentials.tar.gz.enc -out credentials.tar.gz -d
```
bshaffer: Can we just encrypt the plaintext credentials and add that as a travis environment variable?

SurferJeffAtGoogle: > Can we just encrypt the plaintext credentials and add that as a travis environment variable?

Yes, we could do that.  But then we'd have to write the contents of the environment variable to a .json file so that Application Default Credentials can pick them up.  That feels more round-about.

bshaffer: works for me. I'd prefer to have it in the `.travis.yml` so it's all in one file instead of two, but that's just personal preference.

.travis.yml
```diff
@@ -0,0 +1,30 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+language: php
+php:
+  - 5.5
+  - 5.6
+  - hhvm
+  - nightly
```
bshaffer: I just noticed this. We'll want to remove this, as a nightly build will be more trouble than its worth to maintain. instead we should just add `7` for PHP 7.

PR #10, opened 2015-10-15
Store the application credential's raw json in a base64-encoded

SurferJeffAtGoogle:
environment variable so that travis tests can correctly auth in forked
repositories.  No matter of quote and newline escaping worked, so base64
encoding was the way to go.  It's also easy since base64 is a Linux
command-line tool.


dump.credentials.php
```diff
@@ -0,0 +1,10 @@
+<?php
```
bshaffer: This should be dump_credentials.php, unless "credentials" is considered part of the extension.

tmatsuo:
I would like to see our CONTRIBUTING.md having a section for running the test on travis.

bshaffer:
We need to write these tests so they skip when credentials aren't present.

SurferJeffAtGoogle:
I have accommodated all your comments.  The code is ready for another look!

dump_credentials.php
```diff
@@ -0,0 +1,9 @@
+<?php
+/**
+ * Dumps the contents of the environment variable GOOGLE_CREDENTIALS_BASE64 to
+ * a file.
+ *
+ * To setup Travis to run on your fork, read TRAVIS.md.
+ */
+file_put_contents(getenv('GOOGLE_APPLICATION_CREDENTIALS'),
```
bshaffer: nitpicky alert! This should probably be:

```php
file_put_contents(
    getenv('GOOGLE_APPLICATION_CREDENTIALS'),
    base64_decode(getenv('GOOGLE_CREDENTIALS_BASE64'))
);
```

tmatsuo:
LGTM
Nice change :)

bigquery/api/test.php
```diff
@@ -16,8 +16,20 @@
...
+        $path = getenv('GOOGLE_APPLICATION_CREDENTIALS');
+        self::$hasCredentials = $path && file_exists($path) &&
+            filesize($path) > 0;
+    }
+
     public function test()
     {
+        if (!self::$hasCredentials) {
+            $this->markTestSkipped('No application credentials were found.');
+        }
```
bshaffer: As we add more tests, this would be great to put in a base class.

```php
Abstract class BaseTest extends PHPUnit_Framework_TestCase
{
    protected $credentials;

    public function checkCredentials()
    {
        $path = getenv('GOOGLE_APPLICATION_CREDENTIALS');
        $this->credentials = $path && file_exists($path) ? file_get_contents($path) : null;

        if (!$this->credentials) {
            $this->markTestSkipped('No application credentials were found.');
        }
    }
}
```

bigquery/api/test.php
```diff
@@ -16,8 +16,20 @@
  */
 class MainTest extends PHPUnit_Framework_TestCase
 {
+    protected static $hasCredentials;
```
bshaffer: I'm not crazy about static properties and methods. Using a base class would be cleaner IMHO.

bshaffer:
Some suggestions but overall LGTM

dump_credentials.php
SurferJeffAtGoogle: Fixed.

bigquery/api/test.php
SurferJeffAtGoogle: When I add a second test, I'll definitely consider refactoring the code.

bshaffer: SGTM

bshaffer:
Feel free to merge this, although it'd be best to do a `git rebase -i HEAD~3` and then replace the last two `pick` commits with `f` so they are squashed into a single commit.

tmatsuo:
LGTM!

PR #13, opened 2015-10-28
Adds PubSub example for PHP

bshaffer:
TODO: add tests

### DIFF
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -1,0 +1,74 @@
+# Google PubSub PHP Sample Application
+
+## Description
+
+Note: The push endpoints don't work with the App Engine's local
+devserver. The push notifications will go to an HTTP URL on the App
+Engine server even when you run this sample locally. So we recommend
+you deploy and run the app on App Engine.
+TODO(tmatsuo): Better implementation for devserver.
+
+## Register your application
+
+- Go to
+  [Google Developers Console](https://console.developers.google.com/project)
+  and create a new project. This will automatically enable an App
+  Engine application with the same ID as the project.
+
+- Enable the "Google Cloud Pub/Sub" API under "APIs & auth > APIs."
+
+- For local development also follow the instructions below.
+
+  - Go to "Credentials" and create a new Service Account.
+
+  - Select "Generate new JSON key", then download a new JSON file.
+
+  - Set the following environment variable.
+
+    GOOGLE_APPLICATION_CREDENTIALS: the file path to the downloaded JSON file.
+
+## Prerequisites
+
+- Install Python-2.7, pip-6.0.0 or higher and App Engine Python SDK.
+  We recommend you install
+  [Cloud SDK](https://developers.google.com/cloud/sdk/) rather than
+  just installing App Engine SDK.
+
+- Install Google API client library for PHP into 'lib' directory by running:
+
+```
+$ composer install
+```
+
+## Configuration
+
+- Edit config.json
+    - Replace '{A_UNIQUE_TOKEN}' with an arbitrary secret string of
+      your choice to protect the endpoint from abuse.
+
+    - Replace '{YOUR_PROJECT_ID}' with your project ID
+
+## Deploy the application to App Engine
+
+```
+$ gcloud preview app deploy app.yaml --set-default --project YOUR_PROJECT_ID
+```
+
+Then access the following URL:
+  https://{YOUR_PROJECT_ID}.appspot.com/
+
+## Run the application locally
+
+```
+$ dev_appserver.py -A your-project-id .
+```
+
+## Contributing changes
+
+* See [CONTRIBUTING.md](../CONTRIBUTING.md)
+
+## Licensing
+
+* See [LICENSE](../LICENSE)
+
+
--- a/pubsub/README.md
+++ b/pubsub/README.md
@@ -29,7 +29,8 @@ TODO(tmatsuo): Better implementation for devserver.
 
 ## Prerequisites
 
-- Install Python-2.7, pip-6.0.0 or higher and App Engine Python SDK.
+- Install [`composer`](https://getcomposer.org)
+- Install the App Engine Python SDK.
   We recommend you install
   [Cloud SDK](https://developers.google.com/cloud/sdk/) rather than
   just installing App Engine SDK.
--- a/pubsub/app.yaml
+++ b/pubsub/app.yaml
@@ -1,0 +1,11 @@
+runtime: php55
+api_version: 1
+threadsafe: true
+
+handlers:
+- url: /js
+  static_dir: web/js
+- url: /receive_message.*
+  script: web/index.php
+- url: /.*
+  script: web/index.php
--- a/pubsub/app.yaml
+++ b/pubsub/app.yaml
@@ -5,7 +5,5 @@ threadsafe: true
 handlers:
 - url: /js
   static_dir: web/js
-- url: /receive_message.*
-  script: web/index.php
 - url: /.*
   script: web/index.php
--- a/pubsub/src/DatastoreHelper.php
+++ b/pubsub/src/DatastoreHelper.php
@@ -19,10 +19,15 @@
 namespace GoogleCloudPlatform\DocsSamples\Pubsub;
 
 /**
-*
-*/
+ * Utility class for making calls to datastore. Functions have been written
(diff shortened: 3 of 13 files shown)


pubsub/web/index.php
```diff
@@ -0,0 +1,146 @@
...
+
+// Authenticate your API Client
+$client = new Google_Client();
+$client->useApplicationDefaultCredentials();
+$client->addScope(Google_Service_Pubsub::PUBSUB);
+$client->addScope(Google_Service_Datastore::DATASTORE);
+$client->addScope(Google_Service_Datastore::USERINFO_EMAIL);
+
+// this should be done automatically in the underlying Google_Client class
+$client->setHttpClient(new HttpClient(['handler' => new StreamHandler()]));
```
bshaffer: If we merge #715 we can remove this line

bshaffer:
This library relies on https://github.com/google/google-api-php-client/pull/712 for the PubSub API calls

bshaffer:
@SurferJeffAtGoogle @tmatsuo

pubsub/src/DatastoreHelper.php
```diff
@@ -0,0 +1,91 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace GoogleCloudPlatform\DocsSamples\Pubsub;
+
+/**
+*
+*/
+class DatastoreHelper
```
SurferJeffAtGoogle: Add a comment for the class.

pubsub/src/DatastoreHelper.php
```diff
@@ -0,0 +1,91 @@
...
+{
+    public function createQuery($limit = 20)
```
SurferJeffAtGoogle: Add a comment for each function.

For example, I thought this was a very generic class at first.  Then, I was surprised to see line 39.

pubsub/src/PubsubHelper.php
```diff
@@ -0,0 +1,92 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace GoogleCloudPlatform\DocsSamples\Pubsub;
+
+/**
+*
+*/
+class PubsubHelper
```
SurferJeffAtGoogle: class comment please.

pubsub/src/PubsubHelper.php
```diff
@@ -0,0 +1,92 @@
...
+{
+    public function setupTopic($projectId, $topicName, \Google_Service_Pubsub $pubsub)
```
SurferJeffAtGoogle: Comment for each method please.

pubsub/src/PubsubHelper.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    public function setupSubscription($projectId, $topicName, $subscriptionName, $token, \Google_Service_Pubsub $pubsub)
+    {
+        $fullSubscriptionName = sprintf('projects/%s/subscriptions/%s', $projectId, $subscriptionName);
+
+        try {
+            // NOTE: if the version changes, the subscriber will
+            // continue to be the older version, so you'll need to
+            // delete the subscriber in Google Developer Console in
```
SurferJeffAtGoogle: Delete the subscriber or the subscription?

pubsub/test/pubsubTest.php
```diff
@@ -0,0 +1,66 @@
...
+use Silex\WebTestCase;
+
+class pubsubTest extends WebTestCase
+{
+    public function createApplication()
+    {
+        $appCode = file_get_contents(__DIR__.'/../web/index.php');
+
+        // instead of running it, return it
+        $appCode = str_replace('$app->run();', 'return $app;', $appCode);
```
SurferJeffAtGoogle: This looks scary.

pubsub/README.md
```diff
@@ -0,0 +1,74 @@
...
+$ composer install
+```
+
+## Configuration
+
+- Edit config.json
+    - Replace '{A_UNIQUE_TOKEN}' with an arbitrary secret string of
+      your choice to protect the endpoint from abuse.
+
+    - Replace '{YOUR_PROJECT_ID}' with your project ID
```
broady: FYI, @jonparrott and I have been using environment variables

pubsub/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Deploy the application to App Engine
+
+```
+$ gcloud preview app deploy app.yaml --set-default --project YOUR_PROJECT_ID
```
broady: I believe `--set-default` is now `--promote`. you may also want to tell people to use `--version`

pubsub/app.yaml
```diff
@@ -0,0 +1,11 @@
...
+api_version: 1
+threadsafe: true
+
+handlers:
+- url: /js
+  static_dir: web/js
+- url: /receive_message.*
+  script: web/index.php
+- url: /.*
+  script: web/index.php
```
broady: /receive_message handler is redundant

pubsub/src/PubsubHelper.php
broady: `setUp...`

pubsub/test/pubsubTest.php
bshaffer: Yeah, it is... 

I wanted to keep things in one file, and didn't want to make exceptions for the sake of testing. But really it would just be one file (`app.php`) that would live somewhere else (`src` probably ) and then get included by `index.php`.  Although `src` is usually for class files, so that's also annoying.

pubsub/templates/pubsub.html.twig
```diff
@@ -0,0 +1,36 @@
...
+        Start auto update
+    </button>
+    per <input type="text" size="3" ng-model="PubsubController.interval"> seconds.
+    <span ng-show="PubsubController.errorNotice"><b>{{ "{{ PubsubController.errorNotice }}" }}</b></span>
+
+    <h2>Messages:</h2>
+    <ul>
+        <li ng-repeat="m in PubsubController.messages track by $index">{{ "{{ m }}" }}</li>
+    </ul>
+    <script src="//ajax.googleapis.com/ajax/libs/angularjs/1.2.15/angular.min.js"></script>
```
broady: Angular is pretty overkill for a sample like this.

pubsub/web/js/pubsub.js
```diff
@@ -0,0 +1,88 @@
...
+                        status + '.');
+      self.failCount += 1;
+    });
+  if (self.failCount < pubsub.PubsubController.MAX_FAILURE_COUNT) {
+    if (self.isAutoUpdating) {
+      self.promise = self.timeout(
+        function() { self.fetchMessages(); },
+        self.interval * pubsub.PubsubController.TIMEOUT_MULTIPLIER);
+    }
+  } else {
```
broady: Return early.

pubsub/README.md
SurferJeffAtGoogle: YOUR_PROJECT_ID definitely needs to be settable from an environment variable, so that forked repos can run travis tests.

SurferJeffAtGoogle: I've been naming this GOOGLE_PROJECT_ID in our other samples.  We probably want to keep it consistent.

bshaffer: :+1:

bshaffer:
this might be an overkill sample... there are calls made to `datastore` which clutter what we're trying to do.

I can split this into two samples - datastore and pubsub - and try to simplify the pubsub part.

SurferJeffAtGoogle:
It's not clear to me what the app does.  I don't know what data is being
stored where and why.  An overview in the README.md would help.

bshaffer:
Plan for this example: 

1. Split into 2 examples - `pubsub` and `datastore`
2. Add pull to the `pubsub` example
3. Make `pubsub` example store messages in memory only, and add note when on appengine that strange behavior may happen across multiple instances.

SurferJeffAtGoogle:
Could we please not lead users down the wrong path in step 3?  Isn't there
another example we could show?

I got really annoyed in the cloud training class when they said "build it
like this.  But, of course you could never build a production system like
this."

bshaffer:
@SurferJeffAtGoogle That's not the same IMHO - we aren't building production software for the users, we are building simple examples. Our samples shouldn't be opinionated in implementation detail that varies for each application.

PR #14, opened 2015-11-05
Adds a datastore sample

bshaffer:


bshaffer:
bump

datastore/README.md
```diff
@@ -0,0 +1,74 @@
+# Google Datastore PHP Sample Application
+
+## Description
+
+Note: The push endpoints don't work with the App Engine's local
```
SurferJeffAtGoogle: I'm confused because Datastore doesn't have push endpoints.

datastore/README.md
```diff
@@ -0,0 +1,74 @@
...
+  - Select "Generate new JSON key", then download a new JSON file.
+
+  - Set the following environment variables:
+
+    - GOOGLE_APPLICATION_CREDENTIALS: the file path to the downloaded JSON file.
+
+## Prerequisites
+
+- Install [`composer`](https://getcomposer.org)
+- Install the App Engine Python SDK.
```
SurferJeffAtGoogle: Python or PHP?

bshaffer: oh snap!

datastore/README.md
bshaffer: Good call.. I copied the README from pubsub :fearful:

datastore/src/DatastoreHelper.php
```diff
@@ -0,0 +1,111 @@
...
+{
+    const KIND = 'DatastoreComment';
+
+    /**
+     * Creates a query object for pulling the last $limit number of
+     * "PubSubMessage" items
+     */
+    public function createSimpleQuery($limit = 20)
+    {
+        $request = new \Google_Service_Datastore_RunQueryRequest();
```
SurferJeffAtGoogle: It would be helpful to have a psuedo-sql statement in comments.

bshaffer: that's a good idea

datastore/test/datastoreTest.php
```diff
@@ -0,0 +1,64 @@
...
+use Silex\WebTestCase;
+
+class datastoreTest extends WebTestCase
+{
+    public function createApplication()
+    {
+        $appCode = file_get_contents(__DIR__.'/../web/index.php');
+
+        // instead of running it, return it
+        $appCode = str_replace('$app->run();', 'return $app;', $appCode);
```
SurferJeffAtGoogle: This is scary, and I haven't needed to do it when writing my tests.

datastore/src/DatastoreHelper.php
bshaffer: I could add this, but it might end up being more confusing. Thoughts?

```php
    /**
     * Creates a query object for pulling the last $limit number of
     * "PubSubMessage" items, similar to the following statements 
     * in other DBMS:
     *
     * In SQL:
     *
     *     SELECT * from DatastoreComment ORDER BY created DESC LIMIT 20
     *
     * In MongoDB:
     *
     *     db.DatastoreComment.find().limit(20).sort({
     *         "created": -1
     *     });
     */
```

datastore/test/datastoreTest.php
bshaffer: We've discussed this in the pubsub sample... it's there because without it I'd need to create a second file for `app.php` and include it in the front controller, which I really don't care to do.

datastore/src/DatastoreHelper.php
SurferJeffAtGoogle: One line of GQL would be good.  GQL is a real thing that's immediately readable to anyone who knows SQL.  And GQL is the real query language that will be used when querying datastore.

bshaffer: Okay, good call. Looking at the [GQL reference](https://cloud.google.com/appengine/docs/python/datastore/gqlreference?hl=en), I think the first line there is accurate.

bshaffer: done

SurferJeffAtGoogle:
Please update the namespaces to match our new policy.

bshaffer:
@SurferJeffAtGoogle dunzo

SurferJeffAtGoogle:
LGTM.

bshaffer:
updated README, which will hopefully be the last thing. PTAL.

SurferJeffAtGoogle:
LGTM.

PR #23, opened 2016-01-27
[WIP] Adding CakePHP template app

tmatsuo:
The second commit is just invoking `composer create-project --prefer-dist --no-scripts --no-install cakephp/app cakephp`.
The third commit is the changes needed for Managed VMs.

Still, WIP, but I'd like to hear early opinion on this from you @bshaffer @JustinBeckwith


bshaffer:
This seems like a lot of code for what ideally would be a simple app. Can we strive for the absolute minimum configuration here?

tmatsuo:
Are you talking about the bookmarker app?
This is because we want to test some `real app` instead of just a hello world. Otherwise, we can not make sure the framework is really working with our image.

bshaffer:
I agree, but do we really need to include i18n support? Also, email notifications are disabled, so can we get rid of all the email logic? Things like that would make make it easier to find the configuration that matters.

bshaffer:
Also, it seems unnecessary to have two `.htaccess` files. And does this example really need `.gitattributes` and `.editorconfig`?  Also, I believe the `bin/cake.*` files should be ignored, and installed by composer.

bshaffer:
One last thing - do we need `app.php` and `app.default.php`? What's the difference?

bshaffer:
One last thing - do we need `app.php` and `app.default.php`? What's the difference?

tmatsuo:
Those are all good points.

It is because I started with the directory structure generated by `composer create-project --prefer-dist cakephp/app bookmarker`.

The `create-project` command is equivalent to git clone the target repo (in this case [cakephp/app](https://github.com/cakephp/app)), and run `composer install` and that repo contains all of them.

I started to think that we should not re-distribute the `cakephp/app` as a sample. Other options come to mind are:

1. Just have the necessary config files and the instructions in this directory so that you can easily make your CakePHP app "Managed VMs ready".

2. Create our own starter app and upload it to packagist so that anyone can do `composer create-project --prefer-dist google/cakephp-mvms` or similar to get the skeleton app.

I am leaning towards 2 above. What do you think?

JustinBeckwith:
Great conversation.  I have a few goals here:

1. Users should be able to easily start with "I want to run CakePHP on Managed VMs - what do?" and land at a guide that steps them through getting it running.  We have a list of these docs available here for nodejs, just to give you an idea of what I'm striving for:

  https://cloud.google.com/nodejs/resources/frameworks

  Just like we have express/hapi/restify in nodeland, I'd hope to have similar getting started guides for Cake/Laravel/etc.  These should focus on lightweight getting started experiences.

2. To make sure those getting started guides work - we backed them up with a code sample available from the docs-and-samples repository:

  https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/appengine/express

  This allows us to not only show the code sample, but run a test that ensures running them on App Engine still works on a regular basis :) 

I know it's checking in a lot of code - but I would like to check in a basic scaffolded app that has the necessary app.yaml and Dockerfile to clone and just deploy to App Engine.  In the readme, we should just walk through the steps needed to start from the scaffolding tool and build this yourself.  If we want to have more complete e2e samples that's awesome - I just don't think that's what we want to do in php-docs-samples.

tmatsuo:
If we distribute a scaffolded app after all, I think it's better to upload it to packagist because the initial process will be just a one liner: `composer create_project google/cakephp-mvms`, instead of git cloning and cding to the target directory then running composer install.

In order to do that it needs to be in an independent repo, and it comes with some burden to maintain, but I think it's still worth it.

Thoughts?

tmatsuo:
Re: mail

We can configure the mailer to use sendgrid with the port 2525:
https://cloud.google.com/compute/docs/tutorials/sending-mail/

tmatsuo:
Ok, we can start with a subdir in this directory (as this PR is) and let's explorer better options later.

tmatsuo:
@bshaffer 

I updated the PR.

1. Removed .gitattributes and .editorconfig
1. Now the second commit is generated by `composer create-project --prefer-dist --no-scripts --no-install cakephp/app cakephp` thus now we only have `app.config.default`.
1. Added mail tweak for sendgrid
1. Use [`env`](http://api.cakephp.org/3.1/function-env.html) instead of `getenv` if applicable

Can you take another look?

Re: i18n
Can we just leave it since many apps will need i18n after all.

Re: 2 .htaccess files
I assume these are required to run the app locally with apache. Can we leave them?

tmatsuo:
FYI I proposed a small change to cakephp/app at https://github.com/cakephp/app/issues/331 which may allow us to stop hosting cakephp/app here.

appengine/managed_vms/cakephp/config/app.default.php
```diff
@@ -0,0 +1,332 @@
...
+            'driver' => 'Cake\Database\Driver\Mysql',
+            'persistent' => false,
+            'host' => env('MYSQL_HOST', 'localhost'),
+            /**
+             * CakePHP will use the default DB port based on the driver selected
+             * MySQL on MAMP uses port 8889, MAMP users will want to uncomment
+             * the following line and set the port accordingly
+             */
+            //'port' => 'nonstandard_port_number',
+            'username' => env('MYSQL_USER', 'my_app'),
```
ADmad: Instead of multiple env vars you can have a single one with DSN string and only set `'url`' key in the config array as shown [here](http://book.cakephp.org/3.0/en/orm/database-basics.html#configuration). Similar can be done for EmailTransport config too.

tmatsuo: @ADmad Thanks will do.

tmatsuo:
FYI, it turned out that we can use DSN for most of the custom configs, so I changed this to use the DSN.

appengine/managed_vms/cakephp/.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+
+before_script:
+  - sh -c "composer require 'cakephp/cakephp-codesniffer:dev-master'"
+  - phpenv rehash
+
+script:
+  - sh -c "vendor/bin/phpcs -p --extensions=php --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests ./config ./webroot"
+
+notifications:
+  email: false
```
bshaffer: Why do we have a `.travis.yml` file here? It won't be used in this build. Is this for people to copy over into their own project? If so, I think it'd be better to leave it out, since it'll be different for each case.

tmatsuo: Removed

PR #26, opened 2016-03-14
adds symfony example

bshaffer:


tmatsuo:
LGTM

managed_vms/symfony/php.ini
```diff
@@ -0,0 +1,6 @@
+suhosin.post.max_vars = 1000
+suhosin.request.max_vars = 1000
+
+; we should tighten up these restrictions eventually
+suhosin.executor.func.blacklist =
```
JustinBeckwith: :(

JustinBeckwith:
worried about our suhosin abuse, but lgtm

managed_vms/symfony/php.ini
bshaffer: turn that frown upside down!
):
wait

bshaffer:
@JustinBeckwith ok, I'll try to open that up a bit before merging. Still trying to get the `run-tests.sh` script to pass.

bshaffer:
@JustinBeckwith actually, this is such a pain to get to work with suhosin, I really don't think we should try to restrict it until we understand which functions we actually want blacklisted.

bshaffer:
You can [run suhosin in simulation mode](https://suhosin.org/stories/howtos.html#finding-out-which-functions-to-allow), which will log the files you need to un-blacklist.

managed_vms/symfony/php.ini
```diff
@@ -0,0 +1,14 @@
...
+;;;;;;;;;;;;;;;;;;;
+
+; This file can be used to customize the php configuration for your
+; Managed VMs instances. The default PHP runtime includes the suhosin extension.
+; Functions can be whitelisted to keep the blacklist unmodified.
+
+; Below is the list of suhosin blacklisted functions by default. This can be
+; modified by removing functions from the list.
+
+; removed from this list is "libxml_disable_entity_loader" for symfony
```
bshaffer: @tmatsuo is this preferable, or should we just remove `libxml_disable_entity_loader` from the default configuration?

tmatsuo: Yeah, this is ugly, but this is how you remove a function from the list. Probably we should provide a better way for specifying this, but it's not implemented yet.

tmatsuo:
LGTM

JustinBeckwith:
Thank you for placating me :) If we're having this much trouble, I'm pretty sure customers will as well.  Is the answer to enable override of the blacklist and optional use of a whitelist as well?

bshaffer:
@JustinBeckwith I tried using the whitelist, which supersedes the blacklist, and I ran into tons of unexpected errors (functions we hadn't disabled were [suddenly blacklisted](https://github.com/GoogleCloudPlatform/php-docker/issues/72)), so that doesn't work great. Having the blacklist already in the php.ini, so the users can just remove functions from it when needed, may be the easiest option (although it's ugly).

Part of the solution also will be to only blacklist functions which are commonly known to cause problems in the community, and so are avoided in the majority of libraries anyway. I'm arguing that [in this case](https://github.com/GoogleCloudPlatform/php-docker/issues/73), maybe we should just not blacklist the function in the first place.

managed_vms/symfony/php.ini
bshaffer: I just mean, should we do [this](https://github.com/GoogleCloudPlatform/php-docker/issues/73)

managed_vms/symfony/php.ini
```diff
@@ -0,0 +1,14 @@
...
+
+suhosin.executor.func.blacklist = escapeshellarg, escapeshellcmd, exec, highlight_file, lchgrp,lchown, link, symlink, passthru, pclose, popen, proc_close, prog_get_status, proc_nice, proc_open, proc_terminate,shell_exec, show_source, system, gc_collect_cycles, gc_enable, gc_disable, gc_enabled, getmypid, getmyuid, getmygid,getrusage, getmyinode, get_current_user, parse_str, phpinfo, phpversion, php_uname
```
cedricziel: FYI: standard asset linking doesnt work with this configuration

bshaffer: do you know which function(s) need to be removed from the blacklist?

cedricziel:
If I may chime in: A lot of restrictive measures in the php-docker images address issues with shared hosting. Since this is no shared hosting users should *at least* have all the freedom they have on shared hosting where most of the functions above are enabled nowadays since FPM and suexec and friends provide better jailing for the indiviual users. - With docker, the isolation is even higher.

I currently see a strong correlation to the php sandbox env on AppEngine - which I'm not comfortable with.

managed_vms/symfony/php.ini
cedricziel: At least the symlinking capabilities.

But on the other hand: I still need a maintenance gateway to the application. - Currently I use a managed vm, turn it into user mode, attach to the running container and run CLI tasks - which oftentimes break / require workarounds because of the restricted interpreter in combination with for example escapeshellarg and exec.

bshaffer: I think it's more appropriate for `bin/console assets:install` to be ran prior to deploying to managed VMs. What workflow are you using?

cedricziel: As you may know the hook that dumps the assets is executed on composer install and update since the dumping of the assets depends on assets bundles which in turn depend on the environment.

bshaffer: We run the [`composer install`](https://github.com/GoogleCloudPlatform/php-docker/blob/4d55ceea8d993ec30cbc9c6ec9dbaa1a072c9fab/php-nginx/composer.sh#L74) command with the `--no-scripts` flag, so this is never executed.

cedricziel: Ah. Scope. Yes that's true for this example project - I thought you were asking about the wider scope of the php-docker project, where the no scripts flag was added to allow a second run with scripts. Sorry.

PR #27, opened 2016-03-18
Adding wordpress helper

tmatsuo:
It started to work :)

I'll add README and some tests.

### DIFF
--- a/managed_vms/wordpress/src/WordPressSetup.php
+++ b/managed_vms/wordpress/src/WordPressSetup.php
@@ -1,0 +1,258 @@
+<?php
+/**
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\Helper;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+use Symfony\Component\Console\Question\Question;
+
+class WordPressSetup extends Command
+{
+    const DEFAULT_DIR = 'my-wordpress-project';
+    const DEFAULT_ERROR = 1;
+    const LATEST_WP = 'https://wordpress.org/latest.tar.gz';
+    const LATEST_BATCACHE =
+        'https://downloads.wordpress.org/plugin/batcache.1.4.zip';
+    const LATEST_MEMCACHED =
+        'https://downloads.wordpress.org/plugin/memcached.2.0.3.zip';
+
+    const FLEXIBLE_ENV = 'Flexible Environment';
+    const STANDARD_ENV = 'Standard Environment';
+
+    protected function configure()
+    {
+        $this
+            ->setName('setup')
+            ->setDescription('Setup WordPress on GCP')
+            ->addOption(
+                'env',
+                'e',
+                InputOption::VALUE_OPTIONAL,
+                'App Engine environment to use; f: '
+                . self::FLEXIBLE_ENV
+                . ', s: '
+                . self::STANDARD_ENV
+                . '.',
+                null
+            )
+            ->addOption(
+                'dir',
+                'd',
+                InputOption::VALUE_OPTIONAL,
+                'Directory for the new project',
+                self::DEFAULT_DIR
+            )
+            ->addOption(
+                'project_id',
+                'p',
+                InputOption::VALUE_OPTIONAL,
+                'Google Cloud project id',
+                ''
+            )
+            ->addOption(
+                'db_instance',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Cloud SQL instance id',
+                ''
+            )
+            ->addOption(
+                'db_name',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Cloud SQL database name',
+                ''
+            )
+            ->addOption(
+                'db_user',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Cloud SQL database username',
+                ''
+            )
+            ->addOption(
+                'db_password',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Cloud SQL database password',
+                ''
+            );
+    }
+
+    protected function report(OutputInterface $output, ReportInterface $report)
+    {
+        foreach ($report->getInfo() as $value) {
+            $output->writeln("<info>" . $value . "</info>");
+        }
+        if ($report->getErrors() === false) {
+            return true;
+        }
+        foreach ($report->getErrors() as $value) {
+            $output->writeln("<error>" . $value . "</error>");
+        }
+        return false;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $helper = $this->getHelper('question');
+        $configKeys = array(
+            'project_id' => '',
+            'db_instance' => 'wp',
+            'db_name' => 'wp',
+            'db_user' => 'wp',
+            'db_password' => ''
+        );
+        $copyFiles = array(
+            'app.yaml' => '/',
+            'composer.json' => '/',
+            'deploy_wrapper.sh' => '/',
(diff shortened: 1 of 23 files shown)


managed_vms/wordpress-helper/src/files/php.ini
```diff
@@ -0,0 +1,3 @@
+suhosin.executor.func.blacklist=""
```
bshaffer: this is becoming a trend

tmatsuo: I fixed it :)

tmatsuo:
Added the e2e test, which passes locally!
@bshaffer Can you take a look?

bshaffer:
I am not a fan of the subdirectory `wordpress-helper`. I would rather just call it `wordpress` and make the canonical way to install. 

Also (I'm going to do this for symfony), it would be great to [add the option](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/wordpress-helper/managed_vms/wordpress-helper/src/WordPressSetup.php#L41) to specify a different wordpress version, so we could run the tests on multiple versions (wordpress3 and wordpress4). Looking at [the stats](http://wptavern.com/wordpress-version-stats-updated-more-than-13-of-sites-are-running-wordpress-4-1), this is not really very important yet.

tmatsuo:
@bshaffer Alright, changed it to `wordpress`. Re: supporting multiple versions, I'd defer it to another PR. PTAL

managed_vms/wordpress/src/Report.php
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\Helper;
+
+interface Report
```
bshaffer: [By convention](https://github.com/php-fig/fig-standards/blob/master/bylaws/002-psr-naming-conventions.md), all interfaces are suffixed with `Interface`. So the file and class should be renamed to `ReportInterface`.

managed_vms/wordpress/src/WordPressSetup.php
```diff
@@ -0,0 +1,258 @@
...
+        }
+        $project->runComposer();
+        if (!$this->report($output, $project)) {
+            return self::DEFAULT_ERROR;
+        }
+        $output->writeln(
+            '<info>Your WordPress project is ready at '
+            . $project->getDir() . '</info>'
+        );
+        return 0;
```
bshaffer: nit: return should have another return character before it.

managed_vms/wordpress/src/files/gcs-media.php
```diff
@@ -0,0 +1,49 @@
...
+ * @package GCS media
+ * @version 0.1
+ */
+/*
+Plugin Name: GCS media
+Plugin URI: https://tmatsuo-wordpress.appspot.com/
+Description: Use Google Cloud Storage for media upload.
+Author: Takashi Matsuo
+Version: 0.1
+Author URI: https://tmatsuo-wordpress.appspot.com/
```
bshaffer: :+1:

managed_vms/wordpress/src/files/gcs-media.php
```diff
@@ -0,0 +1,49 @@
...
+*/
+
+namespace GCS\Media;
+
+defined('ABSPATH') or die('No direct access!');
+
+add_filter('upload_dir', 'GCS\Media\filter_upload_dir');
+function filter_upload_dir($values)
```
bshaffer: So, I assume this allows you to proxy files to a GCS bucket defined by a global constant `MY_BUCKET`?  Also it looks like this defaults to the storage bucket which is used for uploading all the files for deployment.

A few things:

1. `MY_BUCKET` is a generic constant name for global scope. I would change this to `GOOGLE_CLOUD_STORAGE_BUCKET` to be safe.
2. Do we want to default the bucket like we are? Couldn't this allow anonymous users access to file content?

managed_vms/wordpress/src/files/wp.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+    if (strpos($path, '.php') === false) {
+        return false;
+    } else {
+        chdir(dirname($root . $path));
+        require_once $root . $path;
+    }
+} else {
+    include_once 'index.php';
+}
```
bshaffer: Why do we need to use a custom front controller? Looks like without it you can't run wordpress on PHP's built-in web servier? 

It would be great to not need this, but I see the convenience in keeping it, so LGTM. I'd like to see a comment detailing why we have it, though.

bshaffer:
A bunch of questions and requests for explanations in the comments, otherwise LGTM! Nice work @tmatsuo, can't wait to try it out.

managed_vms/wordpress/src/WordPressSetup.php
tmatsuo: Sorry, but what do you mean?

managed_vms/wordpress/src/files/gcs-media.php
tmatsuo: I think deployment is using a different bucket like `staging...` or something like that.

1. Done (with the next push)
2. Media files are for serving anonymous users. We need to set public-read defacl on the bucket.

managed_vms/wordpress/src/files/wp.php
tmatsuo: > Why do we need to use a custom front controller? Looks like without it you can't run wordpress on PHP's built-in web servier?

Yes. It allows you to maintain your plugins/themes locally on the UI.

Comment added.

managed_vms/wordpress/src/WordPressSetup.php
bshaffer: I thought it was standard to add blank lines before `return`, i.e.:

```php
$output->writeln(
    '<info>Your WordPress project is ready at '
    . $project->getDir() . '</info>'
);

return 0;
```
but after taking a closer look at [the style guide](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#1-overview), I realize I am mistaken, and this is optional for readability only. So this can be ignored.

managed_vms/wordpress/src/files/gcs-media.php
bshaffer: Ahh, ok I think you're right.

Would it be better then to default the bucket to `'wordpress-content.{{project_id}}.appspot.com'` then? Or do you think it's ok to commandeer the main project bucket like this?

tmatsuo: You can change it if you like, but the thing is, the App Engine default GCS bucket is free for the first 5G.

managed_vms/wordpress/src/WordPressSetup.php
tmatsuo: Oh, `return character` == "\n". I see. I'd keep it then.

tmatsuo:
Thanks @bshaffer for the review!
Alright, added some comments. PTAL

managed_vms/wordpress/src/files/gcs-media.php
bshaffer: OK I see. And since it's in `wp-config.php`, the users will have no problem changing it. SGTM.

managed_vms/wordpress/src/files/app.yaml
```diff
@@ -0,0 +1,8 @@
+runtime: custom
```
bshaffer: *technically* this should be `app.yaml.twig` since it's a twig template... I prefer the way it is, however.

managed_vms/wordpress/src/files/gcs-media.php
```diff
@@ -0,0 +1,57 @@
...
+Version: 0.1
+Author URI: https://tmatsuo-wordpress.appspot.com/
+*/
+
+// This plugin changes the destination of your media files to `gs://` URL
+// pointing to your Google Cloud Storage bucket and change the public URL of
+// the media files to the URLs starting with https://storage.googleapis.com/.
+// On environments like App Engine, your instances are ephermeral and the
+// local file system is not shared among multiple instances, so we need the
+// central place for storing and serving your media files.
```
bshaffer: this is perfect! :tada:

managed_vms/wordpress/src/files/wp.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+/**
+ * This file is originally hosted at:
+ * http://ripeworks.com/run-wordpress-locally-using-phps-buily-in-web-server/
+ */
+
+// This is a route PHP file for running your WordPress locally with php
+// builtin web server. The local run allows you to maintain the plugins/themes
+// locally on the UI.
```
bshaffer: I'm being picky, but I think this would be more clear:

```
// This file is a front controller for running your WordPress locally with PHP's
// built-in web server. This allows you to maintain the plugins/themes in the UI
// when running locally. 
```

tmatsuo: Thanks for the suggestion, done!

tmatsuo:
PTAL!

bshaffer:
# LGTM!!!

tmatsuo:
# Thanks, merging!!!

PR #28, opened 2016-03-18
misc fixes for e2e test script, fixes permissions for symfony

bshaffer:


managed_vms/drupal8/tests/run-tests.sh
```diff
@@ -63,11 +63,25 @@ drupal cache:rebuild all
...
+  gcloud preview app deploy \
+    --no-promote --quiet --stop-previous-version --force --docker-build=remote \
+    --project=${GOOGLE_PROJECT_ID} \
+    --version=${GOOGLE_VERSION_ID} \
+      && break
+  attempts=$[$attempts+1]
+  sleep 1
+done
+
 # Deploy to gcloud
```
tmatsuo: This is fine for now, but please consider writing test with php unit and delete the version in tearDownAfterClass.

tmatsuo:
LGTM

managed_vms/drupal8/tests/run-tests.sh
bshaffer: I definitely agree. We could eventually have an e2e testing library for gcloud that we could pull in via composer and use here, in getting started php, and other places too.

PR #32, opened 2016-03-21
allows gcloud to fail

bshaffer:
also:
 - ensures drupal path is correct
 - fails bigquery without project ID


bshaffer:
closing in favor of #33

PR #33, opened 2016-03-21
allows gcloud to fail

bshaffer:
also:
 - ensures drupal path is correct
 - fails bigquery without project ID


bshaffer:
@tmatsuo PTAL

tmatsuo:
LGTM
Still thinking it's better implementing with PHP, but you can do it in another PR

PR #34, opened 2016-03-21
Adding phpmyadmin on App Engine Standard.

tmatsuo:


tmatsuo:
Sorry updated with a minor cs fix

bshaffer:
This seems pretty great. We should add a README though.

tmatsuo:
Thanks, added the README. The preview is available at:
https://github.com/GoogleCloudPlatform/php-docs-samples/tree/phpmyadmin/app_engine/phpmyadmin

PTAL

bshaffer:
LGTM

tmatsuo:
Thanks I just squashed the commits. I will merge this shortly.

bshaffer:
I do want to mention I'm not crazy about using twig to render all these files. It's a little overpowered for this use case, and I think it would be better to just have:

```php
$newContent = strtr([
   '%parameter1%' => $parameter1,
   '%parameter2%' => $parameter2,
   '%parameter3%' => $parameter3,
], file_get_contents($file));
```

PR #35, opened 2016-03-22
Use gcloud instead of appcfg.py.

tmatsuo:
Because there's no way to automate auth with appcfg.py. I need to ommit
the trailing '$' to workaround the following bug:
https://code.google.com/p/google-cloud-sdk/issues/detail?id=573


tmatsuo:
@bshaffer Can you look at this too?

bshaffer:
LGTM

PR #36, opened 2016-03-23
Added a section for adding a slash to requests for /wp-admin.

tmatsuo:


bshaffer:
Seems legit!

Could we add a quick test for ensuring `/wp-admin` is a `200`?

tmatsuo:
@bshaffer Thanks, added a test for that, PTAL

managed_vms/wordpress/tests/DeployTest.php
```diff
@@ -111,4 +111,20 @@ public function testIndex()
             'I am very glad that you are testing WordPress instalation.',
             $resp->getBody()->getContents());
     }
+
+    public function testWpadmin()
+    {
+        // Index serves succesfully with 'Hello World'.
+        // This works because the custom DOCUMENT_ROOT is working.
```
bshaffer: this comment isn't right

managed_vms/wordpress/tests/DeployTest.php
```diff
@@ -111,4 +111,20 @@ public function testIndex()
...
+        $resp = $this->client->get('wp-admin');
+        $this->assertEquals('301', $resp->getStatusCode(),
```
bshaffer: I think it'd be a more useful test to request `/wp-admin/` and make sure it's a `403`.

bshaffer:
minor tweaks

tmatsuo:
@bshaffer Thanks, PTAL

tmatsuo:
Oh wait, guzzle automatically follow the redirects. I think I need to explicitly suppress this.

tmatsuo:
Now I use `https` because nginx thinks the schema is `http` because App Engine terminates SSL. This redirect doesn't work on localhost, but it's not a big deal.

Suppressed the follow_redirects behavior, so that we can just check the 301 response generated by the added configuration.

The test passes locally.

PTAL

managed_vms/wordpress/tests/DeployTest.php
```diff
@@ -111,4 +111,22 @@ public function testIndex()
...
+        $resp = $this->client->request(
+            'GET', 'wp-admin', ['allow_redirects' => false]);
+        $this->assertEquals('301', $resp->getStatusCode(),
+                            'wp-admin status code');
+        $url = sprintf('https://%s-dot-%s.appspot.com/',
+                       getenv(self::VERSION_ENV),
+                       getenv(self::PROJECT_ENV));
+        $this->assertEquals(
+            $url . 'wp-admin/',
+            $resp->getHeader('location')[0]);
```
bshaffer: FYI you can use `$response->getHeaderLine('location')` so it returns a string rather than array, which is pretty handy

bshaffer:
LGTM

managed_vms/wordpress/tests/DeployTest.php
tmatsuo: Ah, thanks. I changed to use that, and also squashed the commit. I will merge when the test passes.

tmatsuo:
Alright, merging

PR #37, opened 2016-03-25
Added a blog_id for gcs media upload.

tmatsuo:
WordPress has a feature called `multisites` that allows you to host multiple blogs from a single WordPress installation. With adding the blog_id as a sub directory, you can have separate directory for each blog.


bshaffer:
will `get_current_blog_id` always return a number? Or is this ever null?

tmatsuo:
Yes it always returns a number

bshaffer:
LGTM

PR #40, opened 2016-04-05
Directory structure change for App Engine standard/flexible.

tmatsuo:
This changes `app_engine` to `app_engine/standard` and `managed_vms` to `app_engine/flexible`.


JustinBeckwith:
+1 on the change, but can we just use appengine instead of app_engine?

tmatsuo:
Something is wrong. I will close this and re open a new one.

PR #43, opened 2016-04-06
Adding support for App Engine standard.

tmatsuo:
Since now the script supports flex and std, so I moved it to `appengine` directory. Added e2e tests for WordPress on std.


tmatsuo:
@bshaffer @JustinBeckwith It's ready for review. Please take a look.

JustinBeckwith:
LGTM

appengine/wordpress/src/WordPressSetup.php
```diff
@@ -216,39 +238,58 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                'gcs-media.php' => '/wordpress/wp-content/plugins/',
+                'nginx-app.conf' => '/',
+                'php.ini' => '/',
+                'wp-config.php' => '/wordpress/',
+            );
+            $templateDir = __DIR__ . '/files/flex';
         }
-        $project->copyFiles($copyFiles, $params);
+        $params = array();
+        self::askParameters($configKeys, $params, $input, $output, $helper);
```
bshaffer: Why are these methods static? I would like to see these be instance methods unless we have a reason.

bshaffer:
in general, we should use `flexible` instead of `flex` in all cases. The abbreviation of `flex` is not a road we want to go down.

appengine/wordpress/src/files/flex/wp-config.php
```diff
@@ -29,52 +29,67 @@
 // Bucket name for the media upload
 define('GOOGLE_CLOUD_STORAGE_BUCKET', '{{project_id}}.appspot.com');
 
-// true on production.
-define('ON_MVM', filter_var(getenv('GAE_VM'), FILTER_VALIDATE_BOOLEAN));
+// ON_GAE is true on production.
+define('ON_GAE', filter_var(getenv('GAE_VM'), FILTER_VALIDATE_BOOLEAN));
```
bshaffer: Because this is only used as a local variable in the two `wp-config.php` files, this should be a local variable `$onGae` instead of a defined constant.

bshaffer:
It's a little confusing that this is outside the `standard` and `flexible` subdirectories. I would like to see the `README` make it clear that this app runs on both, and where the decision is made for which environment it will run in.

appengine/wordpress/src/WordPressSetup.php
tmatsuo: I intentionally made them static because these methods don't use any instance variables, but I can make them instance methods.

tmatsuo:
@bshaffer Re: README, it's a good point! I've modified the README and changed the directory names. PTAL

bshaffer:
awesome! This is really great. LGTM!

tmatsuo:
@bshaffer Thanks! I have just squashed the commits, I'll merge once jenkins build passes.

PR #44, opened 2016-04-06
adds cloudsql example

bshaffer:


appengine/standard/cloudsql/app.php
```diff
@@ -0,0 +1,76 @@
...
+
+$app->post('/', function (Request $request) use ($app) {
+    /** @var PDO $db */
+    $db = $app['database'];
+
+    $name = $request->request->get('name');
+    $content = $request->request->get('content');
+
+    if ($name && $content) {
+        $stmt = $db->prepare('INSERT INTO entries (guestName, content) VALUES (:name, :content)');
```
JustinBeckwith: Not knowing PHP very well - are we sql injection attack safe here?

JustinBeckwith:
LGTM

appengine/standard/cloudsql/README.md
```diff
@@ -0,0 +1,71 @@
...
+composer install
+```
+
+## Setup
+
+Before you can run or deploy the sample, you will need to do the following:
+
+1. Create a [First Generation Cloud SQL](https://cloud.google.com/sql/docs/create-instance) instance. You can do this from the [Cloud Console](https://console.developers.google.com) or via the [Cloud SDK](https://cloud.google.com/sdk). To create it via the SDK use the following command:
+
+        $ gcloud sql instances create YOUR_INSTANCE_NAME
```
tmatsuo: Does it create a first generation instance?

appengine/standard/cloudsql/app.php
tmatsuo: it's safe :)

tmatsuo:
LGTM

appengine/standard/cloudsql/app.php
bshaffer: @JustinBeckwith Good question! PDO [prepared statements](http://php.net/manual/en/pdo.prepared-statements.php) are sufficient to protect against SQL injection: 

    If an application exclusively uses prepared statements, the developer can be sure that no SQL injection will occur

appengine/standard/cloudsql/README.md
bshaffer: Yes

PR #50, opened 2016-04-09
Fixes for WordPress on App Engine standard.

tmatsuo:


tmatsuo:
Jenkins passes too :)

appengine/wordpress/src/WordPressSetup.php
```diff
@@ -249,14 +270,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-            'db_name' => 'wp',
-            'db_user' => 'wp',
-            'db_password' => ''
-        );
         if ($env === self::STANDARD_ENV) {
+            $keys = array(
+                'project_id' => '',
+                'db_instance' => 'wp',
+                'db_name' => 'wp',
+                'cloudsql_root_password' => '',
```
bshaffer: shouldn't this just be `cloudsql_password`? I don't think the user being root should be a requirement.

bshaffer:
Looks good, just one suggestion.

appengine/wordpress/src/WordPressSetup.php
tmatsuo: This is a bit tricky, but from App Engine standard, it always uses root@localhost.

tmatsuo: Ah, sorry you are right. You can use other user too.

tmatsuo: I have changed it to allow other db user. Also always ask you to use the same password locally.

tmatsuo:
@bshaffer PTAL

bshaffer:
LGTM sir

PR #53, opened 2016-04-12
Add code sample for users API.

SurferJeffAtGoogle:
Docs live at https://cloud.google.com/appengine/docs/php/users/

### DIFF
--- a/appengine/standard/users/app.php
+++ b/appengine/standard/users/app.php
@@ -1,0 +1,40 @@
+<?php
+/**
+ * Copyright 2015 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+# [START get_current_user]
+use google\appengine\api\users\User;
+use google\appengine\api\users\UserService;
+use Silex\Application;
+
+// create the Silex application
+$app = new Application();
+
+$app->get('/', function () use ($app) {
+    $user = UserService::getCurrentUser();
+
+    if (isset($user)) {
+        return sprintf('Welcome, %s! (<a href="%s">sign out</a>)',
+            $user->getNickname(),
+            UserService::createLogoutUrl('/'));
+    } else {
+        return sprintf('<a href="%s">Sign in or register</a>',
+            UserService::createLoginUrl('/'));
+    }
+});
+# [END get_current_user]
+
+return $app;
--- a/appengine/standard/users/app.php
+++ b/appengine/standard/users/app.php
@@ -15,15 +15,16 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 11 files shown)


appengine/standard/users/app.php
```diff
@@ -0,0 +1,40 @@
...
+# [START get_current_user]
+use google\appengine\api\users\User;
+use google\appengine\api\users\UserService;
+use Silex\Application;
+
+// create the Silex application
+$app = new Application();
+
+$app->get('/', function () use ($app) {
+    $user = UserService::getCurrentUser();
```
bshaffer: It might be better (for docs purposes) to put the use statements inside the `$app` function:

```php
$app->get('/', function () use ($app) {
    # [START get_current_user]
    use google\appengine\api\users\User;
    use google\appengine\api\users\UserService;
    $user = UserService::getCurrentUser();
    // ....
    # [END get_current_user]
});
```

appengine/standard/users/composer.json
```diff
@@ -0,0 +1,10 @@
+{
+    "require": {
+        "silex/silex": "^1.3",
+        "php-http/guzzle6-adapter": "^1.0",
```
bshaffer: you can remove this dependency - this was specific to `mailgun`

appengine/standard/users/index.php
```diff
@@ -21,23 +21,7 @@
 
 $app = require __DIR__ . '/app.php';
```
bshaffer: This should have been copied and not moved!

appengine/standard/users/app.php
bshaffer: Or follow [this paradigm](https://github.com/GoogleCloudPlatform/php-docs-samples/pull/55/files#diff-f927ee53a61a5d78f3a2d2010a34c58aR18):

```php

# [START import]
use google\appengine\api\users\User;
use google\appengine\api\users\UserService;
# [END import]

$app->get('/', function () use ($app) {
    # [START get_current_user]
    $user = UserService::getCurrentUser();
    // ....
    # [END get_current_user]
});
```

bshaffer:
Tests are failing because we are including an unused namespace:

```php
use google\appengine\api\users\User;
```

appengine/standard/users/app.php
SurferJeffAtGoogle: Intellij complains (which means the interpreter will complain) if I move the use statements into the function:
  Name      app.php    Location      file [php-docs-samples] - .../appengine/standard/users/app.php    Problem synopsis      Use not allowed inside a function at line 25

SurferJeffAtGoogle:
Ready for another look.

appengine/standard/users/composer.json
```diff
@@ -0,0 +1,9 @@
+{
+    "require": {
+        "silex/silex": "^1.3",
+        "google/apiclient": "^1.1"
```
bshaffer: Sorry... I missed this. Looks like we also don't need `google/apiclient`

SurferJeffAtGoogle: Dropped.

SurferJeffAtGoogle:
Ready for another look.

tmatsuo:
Looks good, can you also add deployment test with [E2EDeploymentTrait](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/testing/E2EDeploymentTrait.php)? The modules API sample has [an example](https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/appengine/standard/modules/tests). Maybe you can just test the top page showing a link to the login page.

SurferJeffAtGoogle:
Added tests.

tmatsuo:
LGTM

PR #55, opened 2016-04-12
Adding testutils with a simple testing sample.

tmatsuo:


tmatsuo:
@SurferJeffAtGoogle @bshaffer This is to add the testutils to this repo. The modules app is incomplete, but it shows how to write a test.

appengine/standard/modules/index.php
```diff
@@ -0,0 +1,28 @@
...
+
+// Install composer dependencies with "composer install"
+// @see http://getcomposer.org for more information.
+require __DIR__ . '/vendor/autoload.php';
+
+$app = require __DIR__ . '/app.php';
+
+// Run the app!
+// use "gcloud preview app deploy" or run "php -S localhost:8080"
+// and browse to "index.php"
```
bshaffer: Not exactly true... `index.php` will 404.

appengine/standard/modules/tests/LocalTest.php
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+namespace Google\Cloud\Test;
+
+require_once __DIR__ . '/../../../../testutils/GaeApp.php';
```
bshaffer: This would be better off in `tests/bootstrap.php`

bshaffer:
Rather than have `testutils` and `scripts`, I'd rather combine these both into a single `testing` directory. It's important for this to be named differently, since it doesn't contain any actual tests, only files used in running other tests. But having two top level directories that aren't related to samples is not great, especially when they're closely related.

appengine/standard/modules/index.php
tmatsuo: good catch, done

appengine/standard/modules/tests/LocalTest.php
tmatsuo: done

tmatsuo:
Moved the files to `testing`. @bshaffer PTAL

appengine/standard/modules/tests/LocalTest.php
```diff
@@ -0,0 +1,83 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+namespace Google\Cloud\Test;
+
+use GuzzleHttp\Client;
+use Google\Cloud\TestUtils\GaeApp;
+
+class LocalTest extends \PHPUnit_Framework_TestCase
+{
```
bshaffer: It would be cool if we could have a `LocalDeploymentTrait` also.

testing/E2EDeploymentTrait.php
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Use this trait to deploy the project to GCP for an end-to-end test.
+ */
+trait E2EDeploymentTrait
+{
+    private static $gaeApp;
+    private $client;
+
+    public static $projectEnv = 'GOOGLE_PROJECT_ID';
+    public static $versionEnv = 'GOOGLE_VERSION_ID';
```
bshaffer: These would be better off as constants, unless you think might need to be changed

bshaffer: I imagine you have a reason for doing this, since these same constants are defined in `LocalTest`. Maybe move them to constants of `GaeApp`? It's a small thing, but it'd be cleaner to only have this in one place.

bshaffer:
One minor request for consistency around env var names, but otherwise here's one big fat `LGTM` for you!

testing/E2EDeploymentTrait.php
tmatsuo: Actually LocalTest doesn't need the projectId and versionId, so I will remove them from LocalTestTrait.

tmatsuo: Also Trait can not have constants.

tmatsuo:
Removed the unnecessary constants, added LocalTestTrait (however it is always marking itself as skipped). @bshaffer PTAL

testing/E2EDeploymentTrait.php
bshaffer: Ahh, okay that makes sense. SGTM

bshaffer:
LGTM

tmatsuo:
Thanks! Merging

PR #57, opened 2016-04-12
adds logging example

bshaffer:


bshaffer:
better code for [this guy](https://cloud.google.com/appengine/docs/php/logs/)

appengine/standard/logging/composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": {
+        "google/appengine-php-sdk": "^1.9"
```
tmatsuo: Is this OK to have it here? Isn't it conflicting on production?

cedricziel: autoloaders override include paths-seems fine :)

tmatsuo:
Looks good, can you add a test?

appengine/standard/logging/composer.json
bshaffer: This really isn't very useful, since it isn't needed in dev_appserver or production, and doesn't work locally. But at least you get a sensical error rather than a "class not found" error when you run it locally.

bshaffer:
@tmatsuo added tests - PTAL

appengine/standard/logging/phpunit.xml
```diff
@@ -0,0 +1,26 @@
...
+  limitations under the License.
+-->
+<phpunit bootstrap="test/bootstrap.php">
+    <testsuites>
+        <testsuite name="AppEngine logging test">
+            <directory>test</directory>
+        </testsuite>
+    </testsuites>
+    <logging>
+        <log type="coverage-clover" target="../../../build/logs/clover-cloudsql.xml"/>
```
tmatsuo: Can you make it clover-logging.xml ?

appengine/standard/logging/test/loggingTest.php
```diff
@@ -0,0 +1,73 @@
...
+                ->method($expected['method'])
+                ->will($this->returnValue($expected['return']));
+        }
+        foreach ($expectedAppLog as $expected) {
+            $applog1->expects($this->once())
+                ->method($expected['method'])
+                ->will($this->returnValue($expected['return']));
+        }
+
+        LogService::$logs = [ $log1 ];
```
tmatsuo: Maybe it's a damn question. Why is the code using our mock (`LogService`)? `index.php` has a line `use google\appengine\api\log\LogService;`, how do we override it?

appengine/standard/logging/phpunit.xml
tmatsuo: clover-appengine-logging.xml might be better

appengine/standard/logging/test/loggingTest.php
bshaffer: It's because we include it directly in `bootstrap.php`. Since the class has already been loaded and defined, the `autoload handler` is never triggered.

tmatsuo: Oh I see, thanks!

tmatsuo:
LGTM

tmatsuo:
FYI the test won't run on travis unless you add something in .travis.yml.

bshaffer:
@tmatsuo ahh, good call. Added!

appengine/standard/logging/phpunit.xml
```diff
@@ -0,0 +1,26 @@
...
+<phpunit bootstrap="test/bootstrap.php">
+    <testsuites>
+        <testsuite name="AppEngine logging test">
+            <directory>test</directory>
+        </testsuite>
+    </testsuites>
+    <logging>
+        <log type="coverage-clover" target="../../../build/logs/clover-gae-logging.xml"/>
+    </logging>
+</phpunit>
```
tmatsuo: I think you should add whitelist.

tmatsuo:
LGTM :)

PR #58, opened 2016-04-13
Complete modules sample.

tmatsuo:
Code snippets for https://cloud.google.com/appengine/docs/php/modules/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,11 @@ script:
   - composer install
   - phpunit
   - popd
+  # run modules tests
+  - pushd appengine/standard/modules
+  - if [ ${TRAVIS_PHP_VERSION} == '5.5' ]; then composer install; fi;
+  - if [ ${TRAVIS_PHP_VERSION} == '5.5' ]; then env PHP_CGI_PATH=`which php-cgi` LOCAL_TEST_TARGETS='app.yaml backend.yaml' phpunit; fi;
+  - popd
 
 after_script:
   - composer install
--- a/appengine/standard/modules/README.md
+++ b/appengine/standard/modules/README.md
@@ -1,0 +1,30 @@
+# Google App Engine Modules API
+
+This sample application demonstrates how to use Google App Engine's
+modules API.
+
+## Prerequisites
+
+- Install [`composer`](https://getcomposer.org)
+- Install dependencies by running:
+
+```sh
+composer install
+```
+
+## Deploy to App Engine
+
+**Prerequisites**
+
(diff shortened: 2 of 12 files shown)


tmatsuo:
Finally the LocalTestTrait started to work on Travis. The current failure is about Mailgun rate limiting.
Can you take a look @bshaffer ?

tmatsuo:
@bshaffer Sorry for lots of changes, but it's ready for another look PTAL

.travis.yml
```diff
@@ -73,6 +73,11 @@ script:
   - composer install
   - phpunit
   - popd
+  # run modules tests
+  - pushd appengine/standard/modules
+  - if [ ${TRAVIS_PHP_VERSION} == '5.5' ]; then composer install; fi;
+  - if [ ${TRAVIS_PHP_VERSION} == '5.5' ]; then env PHP_CGI_PATH=`which php-cgi` LOCAL_TEST_TARGETS='app.yaml backend.yaml' phpunit; fi;
```
bshaffer: We could consolidate this to:

```yaml
  - if [ ${TRAVIS_PHP_VERSION} == '5.5' ]; then
    pushd appengine/standard/modules;
    composer install;
    env PHP_CGI_PATH=`which php-cgi` LOCAL_TEST_TARGETS='app.yaml backend.yaml' phpunit;
    popd;
    fi;
    
```

bshaffer:
LGTM, although I'd love to see an explanation (in comment or readme) for `backend.php`

.travis.yml
bshaffer: I assume this is because we only want to run the local tests once per build?

bshaffer: One way we could do this is skip the test if `dev_appserver` isn't running (which would be useful for local testing also), and then in `pre_install` start `dev_appserver.py` for PHP5.5 only. This would be cleaner for future dev appserver testing.

tmatsuo: Re: consolidating

Unfortunately it's not possible. I will make it into a script when we add more tests, so that it will be cleaner in the future.

tmatsuo: Re: running dev_appserver in the travis.yml.

This doesn't make sense because potentially we will need to run multiple apps for tests.

tmatsuo: > I assume this is because we only want to run the local tests once per build?
Yes.

tmatsuo:
@bshaffer Thanks!

Added a comment in `backend.php`, and squashed the commits. I will merge this once the test passes.

.travis.yml
bshaffer: Aha, good point. I forgot how `dev_apperver` worked... :P

PR #59, opened 2016-04-13
adds compute engine logging sample

bshaffer:


tmatsuo:
Is it worthwhile to add a test?

bshaffer:
Yeah, I'll add a mock test

bshaffer:
@tmatsuo PTAL

compute/logging/phpunit.xml
```diff
@@ -0,0 +1,26 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<phpunit bootstrap="test/bootstrap.php">
+    <testsuites>
+        <testsuite name="PHP twilio test">
```
tmatsuo: Name: PHP GCE logging test

compute/logging/test/loggingTest.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    public function testExceptionHandler()
+    {
+        include __DIR__ . '/../index.php';
+        global $logger;
+
+        $lastHandler = set_exception_handler(null);
+        $this->assertEquals('fluentd_exception_handler', $lastHandler);
+
+        $e = new Exception('testing exception handler');
```
tmatsuo: Worth put 'testing exception handler' into a variable?

compute/logging/phpunit.xml
```diff
@@ -0,0 +1,26 @@
...
+-->
+<phpunit bootstrap="test/bootstrap.php">
+    <testsuites>
+        <testsuite name="PHP GCE logging test">
+            <directory>test</directory>
+        </testsuite>
+    </testsuites>
+    <logging>
+        <log type="coverage-clover" target="../../../build/logs/clover-gce-logging.xml"/>
+    </logging>
```
tmatsuo: add whitelist

bshaffer: I don't think I need to whitelist anything, since there's only `index.php`, which gets tested.

tmatsuo: Sorry for not explaining why. We should mark which files to measure the coverage. Otherwise `phpunit` thinks it should measure the coverage for everything including under `vendor` directory and we'll have very low coverage report :)

bshaffer: I noticed there was a 30% decrease in coverage without this

Whitelist has been added!

tmatsuo:
LGTM

compute/logging/phpunit.xml
```diff
@@ -0,0 +1,31 @@
...
+  limitations under the License.
+-->
+<phpunit bootstrap="test/bootstrap.php">
+    <testsuites>
+        <testsuite name="PHP GCE logging test">
+            <directory>test</directory>
+        </testsuite>
+    </testsuites>
+    <logging>
+        <log type="coverage-clover" target="../../../build/logs/clover-gce-logging.xml"/>
```
tmatsuo: I think it should be `../../build/logs/clover-gce-logging.xml`.

tmatsuo:
Oh wait. The compute/index.php is not in the list on coveralls.

Probably because it uses a wrong clover log file path.

compute/logging/phpunit.xml
bshaffer: ahh darn, good catch

tmatsuo:
LGTM
I'm wondering though we should make the coverage repirt mechanism more
robust. The current way is a bit error prone.

PR #64, opened 2016-04-14
Run tests for Users API

### LINKED ISSUE #60: run tests for Users API on travis

tmatsuo:
Fixes #60 
Also adding unit tests for Users API and Modules API


.travis.yml
```diff
@@ -73,10 +73,9 @@ script:
   - composer install
   - phpunit
   - popd
-  # run modules tests
-  - pushd appengine/standard/modules
-  - if [ ${TRAVIS_PHP_VERSION} == '5.5' ]; then composer install; fi;
-  - if [ ${TRAVIS_PHP_VERSION} == '5.5' ]; then env PHP_CGI_PATH=`which php-cgi` LOCAL_TEST_TARGETS='app.yaml backend.yaml' phpunit; fi;
+  # run appengine tests
+  - pushd appengine/standard
+  - if [ ${TRAVIS_PHP_VERSION} == '5.5' ]; then testing/run_ae_e2e_tests.sh; fi;
```
bshaffer: Won't this fail because you're in `appengine/standard` and not the root dir? Seems like the `pushd` and `popd` should be removed here, and the one in `testing_run_ae_e2e` should be the full path.

Also, aren't these dev_appserver tests? so the name of the script should be `testing_run_dev_appserver_tests.sh`?

tmatsuo: @bshaffer You're right! It failed on travis :)

Those tests include tests DeployTests too. Potentially we can skip DeployTests on travis.

tmatsuo:
It really took me a long time to solve, but finally I figured it out. We hit https://github.com/symfony/symfony/issues/5030#issuecomment-20678567 and the second dev_appserver didn't start. Now we have a workaround, adding `exec` before the `dev_appserver.py`, and the tests should be green!

@bshaffer PTAL

testing/GaeApp.php
```diff
@@ -136,7 +136,8 @@ public function run(
         $targets = 'app.yaml',
         $phpCgiPath = '/usr/bin/php-cgi'
     ) {
-        $cmd = 'dev_appserver.py --port ' . $this->port
+        $cmd = 'exec dev_appserver.py --port ' . $this->port
```
cedricziel: Just curios: I've had a bad time with python 3 and dev_appserver.py in the past. - If it still depends on python2, - shouldn't we use `exec python2 dev_appserver.py` instead of relying on the env here?

tmatsuo: @cedricziel Good point, thanks for raising this. That said, I'm not so concerned about this because `dev_appserver.py` should just run regardless of the system python. In other words, when Python3 is becoming norm, gcloud sdk (now `dev_appserver.py` is distributed via gcloud) should take care of it.

tmatsuo:
@bshaffer, rebased to the current master. PTAL

bshaffer:
This is really great! There's a few things I find strange:

1. E2E tests imply (to me) a `gcloud preview app deploy` command, so running `dev_appserver` is an integeration test, but not an E2E test.
2. E2E tests (which actually deploy to App Engine) should not run on travis, and only run in Jenkernetes
3. I don't like the `e2e` dir being top level. I'd rather see it be in `tests/e2e`.

Personally, I think this functionality would make much more sense if:

- The actual `e2e` tests were skipped unless an env var "RUN_E2E_TESTS" set, so it defaults to not running them
- The `dev_appserver` tests were skipped unless dev_appserver was found (this may currently be the case). Also include a `SKIP_DEV_APPSERVER_TESTS` env var that could be added in travis to skip these tests
- All tests were moved to a `test` directory, with subdirectories of `e2e` and `dev_appserver` if you prefer.

tmatsuo:
> E2E tests imply (to me) a gcloud preview app deploy command, so running dev_appserver is an integeration test, but not an E2E test.

I'm OK with calling devserver tests `e2e`, because it's an end to end test for local development.

> E2E tests (which actually deploy to App Engine) should not run on travis, and only run in Jenkins

Yes, I will make a change for this after this PR. I want to submit this with the green state.

> I don't like the e2e dir being top level. I'd rather see it be in tests/e2e

done

> The actual e2e tests were skipped unless an env var RUN_E2E_TESTS set, so it defaults to not running them

> The dev_appserver tests were skipped unless dev_appserver was found (this may currently be the case).

I introduced env vars for conditional run. It's kinda hard for checking dev_appserver.py is there in a portable way (until bash on windows becomes norm), so I would like to have an env var to control it.

> All tests were moved to a test directory inside the sample, with subdirectories of e2e and dev_appserver if you prefer.

Moved to `tests/e2e` subdir.

tmatsuo:
@bshaffer PTAL

.travis.yml
```diff
@@ -28,6 +28,7 @@ env:
     - GOOGLE_APPLICATION_CREDENTIALS=$TRAVIS_BUILD_DIR/credentials.json
     - GOOGLE_VERSION_ID=$TRAVIS_JOB_ID
     - PATH=${HOME}/gcloud/google-cloud-sdk/bin:${PATH}
+    - RUN_DEVSERVER_TESTS=true
```
bshaffer: We could add this as a [matrix include](https://github.com/google/google-api-php-client/blob/master/.travis.yml#L27) only on php5.5 like this:

```yaml
matrix:
  include:
    - php: 5.5
      env: RUN_DEVSERVER_TESTS=true
```

testing/run_ae_e2e_tests.sh
```diff
@@ -0,0 +1,34 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+export PHP_CGI_PATH=`which php-cgi`
```
bshaffer: If we add this env var somewhere else (`.travis.yml` or in the test itself), we could get away with adding all these tests to `.travis.yaml` and removing this script. Then they're just going to be skipped automatically.

.travis.yml
tmatsuo: Done

testing/run_ae_e2e_tests.sh
tmatsuo: Removed!

tmatsuo:
Travis is sleeping, not picking up the change any more

tmatsuo:
Also added a unit tests.

tmatsuo:
Travis didn't pick it up because I had a syntax error in `.travis.yaml`! Doh

tmatsuo:
@bshaffer Ready for another look

.travis.yml
```diff
@@ -17,17 +17,21 @@ branches:
...
-  - hhvm
-  - nightly
+
+matrix:
+  include:
+    - php: 5.5
+      env: RUN_DEVSERVER_TESTS=true
+    - php: 5.6
+    - php: hhvm
+    - php: nightly
```
bshaffer: woohoo!

.travis.yml
```diff
@@ -17,17 +17,21 @@ branches:
...
 
 env:
-  globals:
+  global:
     - GOOGLE_APPLICATION_CREDENTIALS=$TRAVIS_BUILD_DIR/credentials.json
     - GOOGLE_VERSION_ID=$TRAVIS_JOB_ID
     - PATH=${HOME}/gcloud/google-cloud-sdk/bin:${PATH}
+    - PHP_CGI_PATH=/home/travis/.phpenv/shims/php-cgi
```
bshaffer: 👍

.travis.yml
```diff
@@ -78,17 +82,22 @@ script:
...
+  - env LOCAL_TEST_TARGETS='app.yaml backend.yaml' phpunit
+  - popd
+  # run users API tests
+  - pushd appengine/standard/users
+  - composer install
+  - phpunit
+  - popd
 
-after_script:
+after_success:
```
bshaffer: I like this change

appengine/standard/modules/app.php
```diff
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+namespace Google\Cloud\Samples\Appengine\Modules;
```
bshaffer: I haven't been namespacing `app.php`. Typically you only use `namespace` at the top of class files. What was the reasoning behind adding this?

appengine/standard/modules/tests/bootstrap.php
```diff
@@ -14,7 +14,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-require __DIR__ . '/../vendor/autoload.php';
+
+require_once __DIR__ . '/../vendor/autoload.php';
 require_once __DIR__ . '/../../../../testing/E2EDeploymentTrait.php';
 require_once __DIR__ . '/../../../../testing/GaeApp.php';
 require_once __DIR__ . '/../../../../testing/LocalTestTrait.php';
```
bshaffer: An idea for the future (not necessarily this PR) we could include a `testing/autoload.php` which will handle registering all relevant testing files. This would be cleaner when we start adding these types of tests to other samples.

appengine/standard/modules/tests/unit/mocks/Functions.php
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+namespace Google\Cloud\Samples\Appengine\Modules;
+
+/**
+ * Simple mock function for file_get_contents.
+ */
+function file_get_contents($url)
```
bshaffer: ahh, I think this is why you added the namespace in `app.php`?

appengine/standard/modules/app.php
tmatsuo: This is to mock `file_get_contents`. I have `Functions.php` to mock it.

bshaffer:
LGTM

appengine/standard/modules/tests/bootstrap.php
tmatsuo: That's an option. Also I'm thinking to upload the testing tool to packagist, maybe with `google/cloud-testutils` or something which means it will be another independent repo. WDYT?

appengine/standard/modules/tests/unit/mocks/Functions.php
tmatsuo: Yes sir

tmatsuo:
Thanks @bshaffer ! Merging

appengine/standard/modules/tests/bootstrap.php
bshaffer: Yeah, that would be even better :D

Issue #68, opened 2016-04-18
[App Engine] Wordpress Standard Environment - Functionality to use CloudSQL 2nd gen


1MichaelOHara:
When running the WordPress code generation, It does not let you populate the CloudSQL connection name with the region. I have created a fork with a rough implantation of this feature. 

Would there be an interest to add the functionality to choose between using 1st or 2nd generation CloudSQL instance?

Current example behavior:
```
:/cloudsql/{{project_id}}:{{db_instance}}
```

Expected example behavior:
```
:/cloudsql/{{project_id}}:{{db_region}}:{{db_instance}}
```

bshaffer:
@1MichaelOHara Yes! Please submit a PR with your changes and we will definitely look at merging. I think this would be great functionality.

PR #69, opened 2016-04-19
Added selection between 1st and 2nd gen CloudSQL for WordPress on App Engine

1MichaelOHara:
See issue #68 for reference.

### DIFF
--- a/appengine/wordpress/src/WordPressSetup.php
+++ b/appengine/wordpress/src/WordPressSetup.php
@@ -40,6 +40,9 @@ class WordPressSetup extends Command
     const FLEXIBLE_ENV = 'Flexible Environment';
     const STANDARD_ENV = 'Standard Environment';
 
+    const CLOUDSQL_FIRST_GEN = 'First Generation CloudSQL Instance';
+    const CLOUDSQL_SECOND_GEN = 'Second Generation CloudSQL Instance';
+
     protected function configure()
     {
         $this
@@ -53,6 +56,17 @@ class WordPressSetup extends Command
                 . self::FLEXIBLE_ENV
                 . ', s: '
                 . self::STANDARD_ENV
+                . '.',
+                null
+            )
+            ->addOption(
+                'sql_gen',
+                's',
+                InputOption::VALUE_OPTIONAL,
+                'CloudSQL generation to use; 2: '
+                . self::CLOUDSQL_SECOND_GEN
+                . ', 1: '
+                . self::CLOUDSQL_FIRST_GEN
                 . '.',
                 null
             )
@@ -74,7 +88,7 @@ class WordPressSetup extends Command
                 'db_instance',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'Cloud SQL instance id',
+                'Cloud SQL region',
                 ''
             )
             ->addOption(
@@ -152,7 +166,8 @@ class WordPressSetup extends Command
         array &$params,
         InputInterface $input,
         OutputInterface $output,
-        $helper
+        $helper,
+        $sqlGen = false
     ) {
         foreach ($configKeys as $key => $default) {
             $value = $input->getOption($key);
@@ -173,6 +188,18 @@ class WordPressSetup extends Command
                     $q->setHidden(true);
                     $q->setHiddenFallback(false);
                 }
+                if ($key === 'db_region')  {
+                    if ($sqlGen === self::CLOUDSQL_FIRST_GEN) {
+                        $value = ':';
+                        $params[$key] = $value;
+                        continue;
+                    } elseif ($sqlGen === self::CLOUDSQL_SECOND_GEN) {
+                        $value = $helper->ask($input, $output, $q);
+                        $value = ':' . $value . ':';
+                        $params[$key] = $value;
+                        continue;
+                    }
+                }
                 $value = $helper->ask($input, $output, $q);
                 if (empty($value)) {
                     $output->writeln(
@@ -181,6 +208,7 @@ class WordPressSetup extends Command
             }
             $params[$key] = $value;
         }
+        return 0;
     }
 
     protected function execute(InputInterface $input, OutputInterface $output)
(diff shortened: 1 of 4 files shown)


appengine/wordpress/src/WordPressSetup.php
```diff
@@ -57,6 +60,17 @@ protected function configure()
                 null
             )
             ->addOption(
+                'sql_gen',
+                's',
+                InputOption::VALUE_OPTIONAL,
+                'CloudSQL generation to use; 2: '
+                . self::CLOUDSQL_SECOND_GEN
+                . ', 1: '
```
tmatsuo: Why using `2` and `1` here and `o` and `t` there? I would also use `f` for First and `s` for Second.

appengine/wordpress/src/WordPressSetup.php
```diff
@@ -74,7 +88,7 @@ protected function configure()
                 'db_instance',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'Cloud SQL instance id',
+                'Cloud SQL region',
```
tmatsuo: Why did you change it to `Cloud SQL region`? Can you add another option instead of changing this?

appengine/wordpress/src/WordPressSetup.php
```diff
@@ -152,7 +166,8 @@ protected function askParameters(
         array &$params,
         InputInterface $input,
         OutputInterface $output,
-        $helper
+        $helper,
+        $sqlGen = false
```
tmatsuo: Do you really need to set `db_region` in `askParameters`?

appengine/wordpress/src/WordPressSetup.php
1MichaelOHara: That is not clean code, I got excited about signing the commit with a GPG key. I was leaning towards using the arguments ```1``` (1st gen) and ```2``` (2nd gen) for the CloudSQL instance generation option. This is because the short version is ```-s``` and the arguments for the environment option already use ```f```  for flexible and ```s``` for standard. @tmatsuo  What are your thoughts?

appengine/wordpress/src/WordPressSetup.php
1MichaelOHara: Yet again, this is not clean code; got excited about signing the commit with a GPG key. I will revert the string change and add another option for ```db_region```.

appengine/wordpress/src/WordPressSetup.php
1MichaelOHara: ```$sqlGen``` was need for making a check on the key ```db_region``` to properly concatenate the string with colons for the connection name. In addition, it does not prompt the user for  a region if they are using a 1st gen CloudSQL instance.

appengine/wordpress/src/WordPressSetup.php
tmatsuo: `1` and `2` work for me

appengine/wordpress/src/WordPressSetup.php
tmatsuo: I prefer the following flow:

1. If you choose App Engine Flex, just assume to use 2nd gen. If you specify `1`, maybe it can just show error
2. If you choose App Engine Standard, defaults to use 2nd gen. You can override it with specifying an option.
3. If you use 2nd gen, defaults to use `us-central1`, but you can override it with specifying an option
4. Check the value for `db_region` and only allow correct values

Sorry for asking a lot, but I want the script to **always just work**

Thanks for the contribution!

tmatsuo:
@1MichaelOHara Another idea came up to my mind.

There is [an API for administrating Cloud SQL instances](https://cloud.google.com/sql/docs/admin-api/), and [the client library](https://github.com/google/google-api-php-client-services/tree/master/SQLAdmin). With this API, you can list your instances, so that you can create a selection list for users, also you'll know which generation is it, which region it's in, etc in order to construct the connection string. This means the script only need to know the instance name.

Here is the example user experiences:

```
What is your project id: tmatsuo-wp
Detecting available SQL instances...
You have following instances:
1: wp (1st Gen)
2: drupal (2nd Gen)
3: test (1st Gen)
...
...
n: Next page
Which instance to use (1-9, or n for next page): 1
Using Cloud SQL instance "wp".
Downloading WordPress...
...
...
Ready!
```

What do you think? Of course, we can have the intermediate solution with the current approach.

appengine/wordpress/src/WordPressSetup.php
1MichaelOHara: Sounds good, changed locally.

appengine/wordpress/src/WordPressSetup.php
1MichaelOHara: All sounds good, just have a few questions. 

1 & 2: My understanding it will prompt users with something like the following. 
```
Please select the App Engine Environment (defaults to Flexible Environment)
  [0] Flexible Environment (Next selection, confirm that it is only CloudSQL 2nd gen instance)
  [1] Standard Environment (Next selection, default to 2nd and offer 1st gen CloudSQL instance) 
```
3: Are you thinking of something like below during the selection

```
Please select the SQL Instance's Region (defaults to us-centeral1)
  [0] us-central1
  [1] us-east1
  [2] ect...
```

4: Just validate the ```db_region``` option's argument against the list of possible regions?

1MichaelOHara:
@tmatsuo That would be definitely be the optimal solution, but I think that would be out of scope for this pull request. How about making it another issue on GitHub, and I will get this one wrapped up and then we can look into that flow more using CloudSQL admin API?

appengine/wordpress/src/WordPressSetup.php
tmatsuo: Yes, you are correct.

tmatsuo:
@1MichaelOHara Yes, it sounds good.

tmatsuo:
@1MichaelOHara Do you still want to work on this?

tmatsuo:
@1MichaelOHara ping

1MichaelOHara:
@tmatsuo My apologies, the project that I was working on moved in another direction. I will not be able to continue working on this pull request.

tmatsuo:
Closing, I opened a new one at #121

PR #70, opened 2016-04-20
adds appengine storage example

bshaffer:


appengine/standard/storage/index.php
```diff
@@ -0,0 +1,31 @@
...
+// @see http://getcomposer.org for more information.
+require __DIR__ . '/vendor/autoload.php';
+
+$app = require __DIR__ . '/app.php';
+
+// change this to your bucket name!
+$app['bucket_name'] = 'cloud-samples-tests-php';
+
+// Run the app!
+// use "gcloud preview app deploy" or run "php -S localhost:8080"
```
tmatsuo: Does the stream wrapper work when you use the PHP builtin web server?

tmatsuo:
LGTM

appengine/standard/storage/index.php
bshaffer: no, it exits with "you must run this on appengine or dev appserver"

tmatsuo: Then remove that part :)

bshaffer:
@tmatsuo @tpryan PTAL

appengine/standard/storage/index.php
bshaffer: good call!

appengine/standard/storage/test/storageTest.php
```diff
@@ -0,0 +1,241 @@
...
+        $app['debug'] = true;
+
+        // prevent HTML error exceptions
+        unset($app['exception_handler']);
+
+        // the bucket name doesn't matter because we mock the stream wrapper
+        $app['bucket_name'] = 'test-bucket';
+
+        // register the mock stream handler
+        if (!in_array('gs', stream_get_wrappers())) {
```
tmatsuo: Not a big deal at all, but can it be in the `test/mocks/CloudStorageStreamWrapper.php` which seems more natural to me.

bshaffer: As in, you want me to call `CloudStorageStreamWrapper::register()`?

tmatsuo: Yes

tmatsuo: But I'm fine with the current code

tmatsuo:
LGTM

appengine/standard/storage/test/storageTest.php
bshaffer: I like this better because it's explicit, and I don't like having actions take place when including a class file. For instance, if you called `class_exists('CloudStorageStreamWrapper')`, the autoloader will include the file, and register the stream wrapper, which may not be the desired behavior.

It's not quite the issue in this case, since it's a mock, but I still like to avoid it.

PR #73, opened 2016-04-20
Feature mail sample

bshaffer:


appengine/standard/mail/index.php
```diff
@@ -0,0 +1,38 @@
...
+// Notice that $image_content_id is the optional Content-ID header value of the
+// attachment. Must be enclosed by angle brackets (<>)
+$image_content_id = '<image-content-id>';
+
+// Pull in the raw file data of the image file to attach it to the message.
+$image_data = file_get_contents('image.jpg');
+
+try {
+    $message = new Message();
+    $message->setSender('from@google.com');
```
tmatsuo: Nit: better use `example.com` if possible

tmatsuo:
LGTM only with a small nit, you can merge

appengine/standard/mail/index.php
bshaffer: I agree, I'll change it

PR #80, opened 2016-04-22
FEATURE mail sample

tpryan:
Forgot to add region tags to code.


tmatsuo:
LGTM

tmatsuo:
I changed it to use `//` and directly pushed to master

PR #85, opened 2016-04-25
Moved createTables.php into the app itself.

tmatsuo:
I want to see if it really skip tests that need credentials for pull requests from external repository.


SurferJeffAtGoogle:
LGTM.

PR #86, opened 2016-04-26
Test refactorings

### LINKED ISSUE #65: Run tests for appengine/standard/memcache

tmatsuo:
- Added more tests for the Users API sample
- Fixed failing test for cloudsql on hhvm
- Retrying when installing test dependencies
- Run tests for appengine/standard/memcache

Fixes #65 and failing build on hhvm

### DIFF
--- a/appengine/standard/users/tests/unit/mocks/User.php
+++ b/appengine/standard/users/tests/unit/mocks/User.php
@@ -22,6 +22,29 @@ namespace google\appengine\api\users;
  */
 class User
 {
+
+    private $email;
+    private $url;
+
+    /**
+     * A simple constructor for the mock User object.
+     */
+    public function __construct(
+            $email = null,
+            $federated_identity = null,
+            $federated_provider = null,
+            $user_id = null)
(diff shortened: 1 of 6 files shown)


appengine/standard/users/tests/unit/mocks/User.php
```diff
@@ -22,13 +22,40 @@
...
      * Returns the user's nickname.
      *
      * @return string
      */
     public function getNickname()
     {
-        return 'Nick';
+        if ($this->email !== null) {
+            return explode('@', $this->email)[0];
+        } else {
```
bshaffer: NIT: remove the `else`, since the `if` has already returned.

bshaffer:
LGTM!

appengine/standard/users/tests/unit/mocks/User.php
tmatsuo: Done

tmatsuo:
Thanks, I'll merge once the test passes

PR #87, opened 2016-04-26
adds http sample

bshaffer:


tmatsuo:
Can you add `appengine/standard/http` to `testing/run_test_suite.sh` so that travis and jenkins will pick up the tests?

bshaffer:
@tmatsuo good catch! It would be awesome to detect directories with `phpunit.xml` that aren't in the list in `run_test_suite.sh`, and throw an error if they aren't run.

tmatsuo:
@bshaffer 

We still need to have special test cases in run_test_suite.sh, for example `appengine/standard/modules` needs to have special environment variable for now. This makes automatic detection a bit difficult.

Isn't it easy enough to just add a directory to `testing/run_test_suite.sh` for now?

appengine/standard/http/http.html.twig
```diff
@@ -0,0 +1,18 @@
...
+
+  <body>
+    <h1>Http Example</h1>
+
+    <form action="/http_request" method="post">
+        <input type="submit" value="Make an HTTP Request" />
+    </form>
+
+    {% if result is defined %}
+        <pre>{{ result }}</pre>
```
tmatsuo: You can use `default` filter:
http://twig.sensiolabs.org/doc/filters/default.html

bshaffer:
@tmatsuo  I wrote a quick test and found two samples not being tested in travis. So I will add it ;)

appengine/standard/http/http.html.twig
tmatsuo: For example, how about:
```php
<pre>{{ result|default('no result')}}</pre>
```

appengine/standard/http/app.php
```diff
@@ -0,0 +1,55 @@
...
+$app['twig.path'] = [ __DIR__ ];
+
+$app->get('/', function () use ($app) {
+    /** @var Twig_Environment $twig */
+    $twig = $app['twig'];
+
+    return $twig->render('http.html.twig');
+});
+
+$app->post('/http_request', function () use ($app) {
```
tmatsuo: Can you make this just a `GET` because it doesn't change anything.

appengine/standard/http/app.php
```diff
@@ -0,0 +1,55 @@
...
+        'http' => [
+            'method' => 'POST',
+            'header' => "content-type: application/x-www-form-urlencoded\r\n" .
+                        "custom-header: custom-value\r\n" .
+                        "custom-header-two: custom-value-2\r\n",
+            'content' => $postData
+        ]
+    ];
+    $context = stream_context_create($context);
+    $result = file_get_contents('http://httpbin.org/post?query=update', false, $context);
```
tmatsuo: This application and the test depends on `http://httpbin.org`, which is not ideal

tmatsuo: How about silently mock file_get_contents with a namespace?

bshaffer: @tmatsuo A dependency on `httpbin.org` in the test *is* ideal (IMHO), because the sample itself depends on httpbin.org. If the site changes or shuts down, we will know about it.

Additionally, adding a namespace to `index.php` and mocking the method is a little hackish, and I only want to do this if it's absolutely necessary

tmatsuo: Ok, I'm fine with this

appengine/standard/http/app.php
bshaffer: since the request happening behind the scenes is a POST, I feel like it makes more sense to make this a POST

appengine/standard/http/http.html.twig
bshaffer: doesn't it make more sense to just leave it out if no request was made (i.e. we are on `/`)? Otherwise, when someone browses to the example, they'll just see "`no result`" and not know what that means

appengine/standard/http/app.php
bshaffer: I can change it to `GET` though, I think it makes sense that way also...

tmatsuo:
LGTM
I don't know why coveralls is not happy, but go ahead!

bshaffer:
@tmatsuo I think it's unhappy because I canceled the build in travis accidentally, and then restarted it :-/

PR #88, opened 2016-04-26
automatically phpunit in any directory containing phpunit.xml

bshaffer:
@tmatsuo PTAL

we were missing `appengine/standard/twilio` and `appengine/standard/mail` tests

### DIFF
--- a/testing/run_test_suite.sh
+++ b/testing/run_test_suite.sh
@@ -20,43 +20,22 @@ if [ "${RUN_CS_FIXER}" = "true" ]; then
     ./php-cs-fixer fix --dry-run --diff --config-file=.php_cs .
 fi
 
-DIRS=(
-    appengine/standard/cloudsql
-    appengine/standard/http
-    appengine/standard/logging
-    appengine/standard/mailgun
-    appengine/standard/mailjet
-    appengine/standard/memcache
-    appengine/standard/phpmyadmin
-    appengine/standard/storage
-    appengine/standard/taskqueue
-    appengine/standard/users
-    appengine/wordpress
-    bigquery/api
-    compute/logging
-    datastore
-    pubsub
-    storage/api
-)
-
-for DIR in "${DIRS[@]}"; do
-  pushd ${DIR}
-  composer install
-  phpunit
-  if [ -f build/logs/clover.xml ]; then
-      cp build/logs/clover.xml \
-          ${TEST_BUILD_DIR}/build/logs/clover-${DIR//\//_}.xml
-  fi
-  popd
+# loop through all directories containing "phpunit.xml" and run them
+find * -name phpunit.xml -not -path '*/vendor/*' -exec dirname {} \; | while read DIR
+do
+    pushd ${DIR}
+    if [ -f "composer.json" ]; then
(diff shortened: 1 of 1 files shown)


bshaffer:
If you want to explicitly include the `DIRS`, we could add something like this at the bottom to check all tests are run:

```bash
# make sure we didn't miss any tests
array_contains () {
    local seeking=$1; shift
    local in=1
    for element; do
        if [[ $element == $seeking ]]; then
            in=0
            break
        fi
    done
    return $in
}

ALL_TESTS_RUN="true"
# change to root dir (for find command)
cd $( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/..
find * -name phpunit.xml -not -path '*/vendor/*' -exec dirname {} \; | while read line
do
    if ! array_contains "$line" "${DIRS[@]}"; then
      echo "ERROR - no test for ${line}"
      ALL_TESTS_RUN="false"
    fi
done

if [ "${ALL_TESTS_RUN}" = "false" ]; then
    exit 1
fi
```

testing/run_test_suite.sh
```diff
@@ -20,42 +20,22 @@ if [ "${RUN_CS_FIXER}" = "true" ]; then
...
-  fi
-  popd
+# loop through all directories containing "phpunit.xml" and run them
+find * -name phpunit.xml -not -path '*/vendor/*' -exec dirname {} \; | while read DIR
+do
+    pushd ${DIR}
+    if [ -f "composer.json" ]; then
+        composer install
+    fi
+    if [ "$DIR" -eq "appengine/standard/modules" ]; then
```
tmatsuo: Little bit concern on this, but maybe I will change the test in `appengine/standard/modules` to not needing this special care.

tmatsuo:
LGTM

Thanks!
Before merging, please check the coveralls site and see whether you have expected files are listed there.

testing/run_test_suite.sh
tmatsuo: maybe you can just use
```bash
if [ "$DIR" = "appengine/standard/modules" ]; then
```
?

bshaffer: @tmatsuo looks like we can add default env-vars to `phpunit.xml` [like this](http://elnur.pro/using-environment-variables-to-add-flexibility-to-phpunit-tests/). I like this approach

bshaffer: @tmatsuo I see you purposefully defaulted it to just `app.yaml`, so maybe using `phpunit.xml` is not desirable?

tmatsuo: Yeah, please try adding the envvar to the `phpunit.xml`!

tmatsuo: No problem at all with using `phpunit.xml`.

tmatsuo: or maybe I can do that in the next cl

bshaffer:
@tmatsuo it looks [good to me](https://coveralls.io/builds/5945336), but I'm not sure how to see the individual files that were captured.

tmatsuo:
LGTM, let's merge it

PR #90, opened 2016-04-27
Update install_test_deps.sh

tmatsuo:
- Disable prompts
- Stopped pinning to 106.0.0
- Print out gcloud info for filing an issue


bshaffer:
noice, LGTM

PR #93, opened 2016-04-28
Moves Drupal E2E tests to PHPUnit

bshaffer:
... and fixes them


tmatsuo:
LGTM

bshaffer:
wow, I thought for sure you were going to tear this apart :-D

So, I should mention I intend to refactor the logic so we can use the test utils for this. But in order to get it done first, I just refactored the existing script.

tmatsuo:
The current Traits don't provide a way to do any preparations before deployment in a reliable way. I think it's fair to merge this now. We can improve our test traits then migrate this test using them in the future.

PR #94, opened 2016-04-28
adds ability to specify a module in drupal8 test

bshaffer:


tmatsuo:
`module` in yaml files is deprecated. Use `service` instead.

bshaffer:
@tmatsuo good catch, PTAL

tmatsuo:
LGTM

PR #95, opened 2016-04-28
endpoints PHP example and tests

bshaffer:


appengine/flexible/endpoints/Dockerfile
```diff
@@ -0,0 +1,6 @@
+# This file will go away once gcloud implements fingerprinting.
```
tmatsuo: With gcloud sdk version 106.00 or later, we can use 
``` 
runtime: php
```
in `app.yaml`, and then you don't need to have the `Dockerfile`.

bshaffer: works for me! :=D

appengine/flexible/endpoints/app.yaml
```diff
@@ -0,0 +1,8 @@
+runtime: custom
```
tmatsuo: Can you try using `rutime: custom`?

You also need runtime_config, so full `app.yaml` would be

```yaml
runtime: php
vm: true

beta_settings:
  # Enable Google Cloud Endpoints API management.
  use_endpoints_api_management: true
  # Specify the Swagger API specification.
  endpoints_swagger_spec_file: swagger.yaml

runtime_config:
  document_root: web 
```

appengine/flexible/endpoints/index.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+// Install composer dependencies with "composer install"
+// @see http://getcomposer.org for more information.
+require __DIR__ . '/vendor/autoload.php';
+
+$app = require __DIR__ . '/app.php';
+
+// Run the app!
+// use "gcloud preview app deploy" or run "php -S localhost:8080"
```
tmatsuo: having `gcloud preview app deploy` here will be an issue when it becomes `gcloud app deploy`.

Can you just remove the comment?

appengine/flexible/endpoints/app.yaml
tmatsuo: Wait, with this sample, you don't need to specify the document_root at all, because it defaults to `/app`, so:

```yaml
runtime: php
vm: true

beta_settings:
  # Enable Google Cloud Endpoints API management.
  use_endpoints_api_management: true
  # Specify the Swagger API specification.
  endpoints_swagger_spec_file: swagger.yaml
```

tmatsuo: Only downside of doing this (use runtime:php) is that we can not pin gcloud sdk to prior version, I feel it's OK though.

appengine/flexible/endpoints/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+In your web browser, go to the following address: http://localhost:8080.
+
+### Using the echo client
+
+With the app running locally, you can execute the simple echo client using:
+
+    $ php endpoints.php make-request http://localhost:8080 APIKEY
+
+The `APIKEY` doesn't matter as the endpoint proxy is not running to do authentication.
```
tmatsuo: How about:
```
The `APIKEY` can be any string as the local endpoint proxy doesn't need authentication.
```

appengine/flexible/endpoints/swagger.yaml
```diff
@@ -0,0 +1,106 @@
+swagger: "2.0"
```
tmatsuo: It might be a problem with the product itself, rather than a problem with this sample itself, but I feel this yaml file is too dificult for me. Can we simplify the auth part for most of the users?

appengine/flexible/endpoints/app.yaml
bshaffer: I agree

appengine/flexible/endpoints/README.md
bshaffer: agreed, I like that better

appengine/flexible/endpoints/swagger.yaml
bshaffer: I agree the `swagger.yaml` is very complicated, and the auth section is hard to parse. @broady and I are looking into simplifying it (or naming some things to make it clearer). 

At the very least, we could add links to the endpoints documentation

broady:
Submit this soon, please?

appengine/flexible/endpoints/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Deploying to Google App Engine
+
+Open the `swagger.yaml` file and in the `host` property, replace
+`YOUR-PROJECT-ID` with your project's ID.
+
+Then, deploy the sample using `gcloud`:
+
+    gcloud preview app deploy app.yaml
```
JustinBeckwith: no app.yaml needed here

appengine/flexible/endpoints/swagger.yaml
tmatsuo: It's ok for now

appengine/flexible/endpoints/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+Once deployed, you can access the application at https://YOUR-PROJECT-ID.appspot.com/.
```
JustinBeckwith: could use `gcloud preview app browse` here if you want to be super nice

JustinBeckwith:
LGTM

appengine/flexible/endpoints/index.php
bshaffer: We will have to change all of them at once, so I think it's ok to keep it

tmatsuo:
LGTM

bshaffer:
thanks guys!

PR #99, opened 2016-05-06
Update index.php

carieshmarie:
Add region tags for displaying in the documentation.

### DIFF
--- a/appengine/standard/mailjet/index.php
+++ b/appengine/standard/mailjet/index.php
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+# [START config]
 // Install composer dependencies with "composer install"
 // @see http://getcomposer.org for more information.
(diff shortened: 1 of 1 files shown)


bshaffer:
This code is rather specific to our implementation (i.e. using `silex`, including `app.php`), and does not illustrate anything useful to someone running mailgun in their own implementation. I am worried that showing this in our documentation will be more confusing than useful.

The idea was to make the code in [functions.php](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/standard/mailgun/functions.php) to stand on its own, similar to [mailjet](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/standard/mailjet/app.php), and then *also* have a sample app if users wanted to see the sample actually running in some context.

Long story short, I don't think we should merge this.

PR #102, opened 2016-05-12
better code for mailgun sample

bshaffer:


appengine/standard/mailgun/functions.php
```diff
@@ -1,43 +0,0 @@
-<?php
```
tmatsuo: Remove functions.php from `phpunit.xml`

tmatsuo:
LGTM

appengine/standard/mailgun/functions.php
bshaffer: good catch! removed.

PR #103, opened 2016-05-13
Use files from php-tools repo.

tmatsuo:


.travis.yml
```diff
@@ -30,13 +30,16 @@ env:
...
-    - PATH=${HOME}/gcloud/google-cloud-sdk/bin:${PATH}
+    - PATH=${HOME}/google-cloud-sdk/bin:${PATH}
     - PHP_CGI_PATH=/home/travis/.phpenv/shims/php-cgi
     - TEST_BUILD_DIR=$TRAVIS_BUILD_DIR
 
 before_install:
-  - php dump_credentials.php
-  - travis_retry testing/install_test_deps.sh
+  - pushd ${HOME}
+  - git clone https://github.com/GoogleCloudPlatform/php-tools.git
```
bshaffer: Why are we cloning this instead of just adding it to `require-dev` in `composer.json`?

tmatsuo: I can do that, but currently we don't have top level `composer.json` and I just thought it's not warrant to add the top level `composer.json`. Each test has it in their `require-dev`.

bshaffer: Ahh, I see. Yeah let's not worry about it then.

bshaffer:
LGTM

tmatsuo:
Thanks!

PR #106, opened 2016-05-24
Add an app engine flex memcache sample.

SurferJeffAtGoogle:


appengine/flexible/memcache/app.php
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+# [START memcached]
+use Silex\Application;
```
bshaffer: I've typically tried to avoid silex-specific logic in the doc blocks.

bshaffer:
We might be able to set [PHP_CGI_PATH](https://github.com/GoogleCloudPlatform/php-tools/blob/master/src/DevAppserverTestTrait.php#L63) to `${HOME}/google-cloud-sdk/platform/php55/php-cgi` so we don't get an error on the unsupported extension, as that is the version bundled with the `gcloud-sdk`.

bshaffer:
As for the `HHVM` failure, I think we should just remove support for it. Nobody needs to run these samples on HHVM, as they're meant to run in our cloud or locally.

.travis.yml
```diff
@@ -42,6 +47,8 @@ before_install:
   - popd
   - mkdir -p build/logs
 
+before_script: phpenv config-add testing/travis-php.ini
```
bshaffer: After some googling, something like this would also work:

```
before_script:
  - if [[ $TRAVIS_PHP_VERSION = hhvm ]]; then 
      cat testing/travis-php.ini >> /etc/hhvm/php.ini;
    else 
       phpenv config-add testing/travis-php.ini;
    fi

File: appengine/flexible/memcache/app.yaml [hunk 3]
```diff
@@ -0,0 +1,5 @@
+runtime: php
+vm: true
+
+runtime_config:
```
JustinBeckwith: is this required?  I thought we defaulted to app or web?

appengine/flexible/memcache/nginx-app.conf
```diff
@@ -0,0 +1,4 @@
+location / {
+  # try to serve file directly, fallback to front controller
+  try_files $uri /index.php$is_args$args;
```
JustinBeckwith: @tmatsuo do we really need to have an nginx-app.conf here?  Wouldn't we have a default rule that looks kinda like this?  I'm just hoping that we won't require an nginx config for every sample, unless it's really required.

appengine/flexible/memcache/app.yaml
bshaffer: I believe the default is `/app`

bshaffer: So yes, this is necessary because `index.php` will be in `/app/web` (see [here](https://github.com/GoogleCloudPlatform/php-docker/blob/2be08b74359c6d5bfd54c058e86ada19b4bfb94b/php-nginx/entrypoint.sh#L62))

appengine/flexible/memcache/nginx-app.conf
tmatsuo: The rules are slightly different among frameworks. I would not make something specific as a default.

bshaffer: This config file basically says **"route all traffic to `index.php`, unless the exact file exists"**. For frameworks, this works great. For flat-file PHP apps, this works great except when the app 404s, and instead of a proper `404` message, the homepage is shown. 

I think defaulting to using `index.php` is a good idea, as flat-file PHP apps are pretty much from the 90's anyway.

bshaffer: What happens if a user doesn't supply an `nginx-app.conf`?

tmatsuo: If a user doesn't supply an `nginx-app.conf`, the requests like `/goodbye` will just 404, instead of being handled by the `index.php`.

bshaffer: @tmatsuo would a request to `/goodbye.php` load the file located in `/app/goodbye.php`? and would a request to `/` load `/app/index.php`? I think that (as a minimum) would be the expected behavior.

PR #112, opened 2016-06-03
A sample that shows how to serve static via nginx on app engine flex.

SurferJeffAtGoogle:


bshaffer:
LGTM!

appengine/flexible/staticcontent/README.md
```diff
@@ -0,0 +1,26 @@
+# Static content and Google App Engine Flexible Environment
+
+This sample application demonstrates how to serve static content with Google App
```
JustinBeckwith: This is a total nit - but can we be very specific here that this is for PHP only?  PHP is the only runtime with nginx baked into the app container.

JustinBeckwith:
LGTM

appengine/flexible/staticcontent/README.md
SurferJeffAtGoogle: I agree this is an important fact for users to consider.  Do you think Python programmers may stumble into the php-docs-samples repo and come to the wrong conclusion?

PR #113, opened 2016-06-07
fix logging routes

bshaffer:


tmatsuo:
LGTM
Ideally there was a test for catching this though

bshaffer:
true... I'll have to add some E2E tests

PR #115, opened 2016-06-08
Add an analytics sample.

SurferJeffAtGoogle:
For https://cloud.google.com/appengine/docs/flexible/php/integrating-with-analytics

### DIFF
--- a/appengine/flexible/analytics/README.md
+++ b/appengine/flexible/analytics/README.md
@@ -1,0 +1,42 @@
+# Google Analytics and Google App Engine Flexible Environment
+
+This sample application demonstrates how track events with Google Analytics
+when running in Google App Engine Flexible Environment.
+
+## Prerequisites
+
+- Install [`composer`](https://getcomposer.org)
+- Install dependencies by running:
+
+```sh
+$ composer install
+```
+
+## Deploy to App Engine
+
+**Prerequisites**
+
+- Install the [Google Cloud SDK](https://developers.google.com/cloud/sdk/).
+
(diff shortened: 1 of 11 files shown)


appengine/flexible/analytics/README.md
```diff
@@ -0,0 +1,40 @@
...
+    For example:
+
+    ```
+    env_variables:
+        GA_TRACKING_ID: your-tracking-id
+    ```
+
+    Before running the sample app locally, set the environment variables required by the app:
+
+    ```export GA_TRACKING_ID=your-tracking-id```
```
bshaffer: we should put these on their own lines for consistency (like the code block above it)

appengine/flexible/analytics/tests/LocalTest.php
```diff
@@ -0,0 +1,45 @@
...
+    public function setUp()
+    {
+        parent::setUp();
+        $this->client = $this->createClient();
+    }
+
+    public function createApplication()
+    {
+        $app = require __DIR__ . '/../app.php';
+        $app['GA_TRACKING_ID'] = 'UA-78971600-1';
```
bshaffer: we could set this in `phpunit.xml` using environment variables like we do [here](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/9c09d5e6e366601eb24be30b71db9bb0590a0332/appengine/standard/modules/phpunit.xml#L19), which would function as a default. This way it could be overridden (or added to our travis config)

bshaffer:
Some nits but otherwise LGTM

PR #116, opened 2016-06-10
Add a datastore sample.

SurferJeffAtGoogle:
For https://cloud.google.com/appengine/docs/flexible/php/using-cloud-datastore

### DIFF
--- a/appengine/flexible/datastore/app.php
+++ b/appengine/flexible/datastore/app.php
@@ -1,0 +1,109 @@
+<?php
+/**
+ * Copyright 2015 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use Silex\Application;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+// create the Silex application
+$app = new Application();
+
+$app['datastore'] = function () {
+    // Datastore API has intermittent failures, so we set the
+    // Google Client to retry in the event of a 503 Backend Error
+    $retryConfig = ['retries' => 2 ];
+    $client = new \Google_Client(['retry' => $retryConfig ]);
+    $client->setScopes([
+        Google_Service_Datastore::CLOUD_PLATFORM,
+        Google_Service_Datastore::DATASTORE,
+    ]);
+    $client->useApplicationDefaultCredentials();
+    return new \Google_Service_Datastore($client);
+};
+
+$app->get('/', function (Application $app, Request $request) {
+    /** @var \Google_Service_Datastore $datastore */
+    $datastore = $app['datastore'];
+    $ip = $request->GetClientIp();
+    // Keep only the first two octets of the IP address.
+    $octets = explode($separator = ':', $ip);
+    if (count($octets) < 2) {  // Must be ip4 address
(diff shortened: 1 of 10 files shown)


appengine/flexible/datastore/app.php
```diff
@@ -0,0 +1,108 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use Silex\Application;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+// Use UTC time.
+date_default_timezone_set('UTC');
```
bshaffer: Rather than set the system-wide default, we should set the timezone on a per-date basis like this:

```php
$date = new DateTime();
$date->setTimezone(new DateTimeZone('UTC'));
//...
    ['timestampValue' => $date->format('Y-m-d\TH:i:s\Z')]
```

appengine/flexible/datastore/app.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+    if (count($octets) < 2) {
+        $octets = ['bad', 'ip'];
+    }
+    // Replace empty chunks with zeros.
+    $octets = array_map(function ($x) { return $x == '' ? '0' : $x; }, $octets);
+    $user_ip = $octets[0] . $separator . $octets[1];
+    // Create an entity to insert into datastore.
+    $key = new \Google_Service_Datastore_Key(['path' => ['kind' => 'visit']]);
+    $properties = ['user_ip' => ['stringValue' => $user_ip],
```
bshaffer: this is a nit, but I'd rather see this as 
```php
$properties = [
    'user_ip' => ['stringValue' => $user_ip],
    'timestamp' => ['timestampValue' => $date->format("Y-m-d\TH:i:s\Z")]
];
```

appengine/flexible/datastore/app.php
```diff
@@ -0,0 +1,108 @@
...
+    ]);
+    $request = new \Google_Service_Datastore_RunQueryRequest();
+    $request->setQuery($query);
+    $response = $datastore->projects->runQuery($dataset_id, $request);
+    /** @var \Google_Service_Datastore_QueryResultBatch $batch */
+    $batch = $response->getBatch();
+    $visits = ["Last 10 visits:"];
+    foreach ($batch->getEntityResults() as $entityResult) {
+        $properties = $entityResult->getEntity()->getProperties();
+        array_push($visits, 'Time: '
```
bshaffer: I'd like to see this as a sprintf:

```php
array_push($visits, sprintf('Time: %s Addr: %s'
    $properties['timestamp']['timestampValue'],
    $properties['user_ip']['stringValue']));
```

bshaffer:
Did we look into using [tomwalder/php-gds](https://github.com/tomwalder/php-gds) at all for this sample?

appengine/flexible/datastore/app.php
SurferJeffAtGoogle: Done.

appengine/flexible/datastore/app.php
SurferJeffAtGoogle: Done.

appengine/flexible/datastore/app.php
SurferJeffAtGoogle: Done.

SurferJeffAtGoogle:
I didn't look at using using tomwalder/php-gds.  I haven't seen us use third-party code like this in samples before.  But that library looks good.  Should I try using it in a follow-up PR?

appengine/flexible/datastore/composer.json
```diff
@@ -0,0 +1,10 @@
+{
+    "require": {
+        "silex/silex": "^1.3",
+        "google/apiclient": "~2.0@dev"
```
tmatsuo: @bshaffer Do we still need @dev?

tmatsuo:
LGTM

appengine/flexible/datastore/composer.json
SurferJeffAtGoogle: Dropped.

bshaffer:
@SurferJeffAtGoogle I think it's worth looking into for a follow up PR. If we did use `php-gds`, we'd want to factor our other usages of datastore in `pubsub` and `datastore` as well. 

This is definitely LGTM!

PR #118, opened 2016-06-14
Automatically create the topic and subscription.

SurferJeffAtGoogle:
If the first attempt at posting a message fails.

This means one fewer error that the user has to read, figure out, and fix.

### DIFF
--- a/pubsub/src/app.php
+++ b/pubsub/src/app.php
@@ -106,4 +106,44 @@ $app->post('/send_message', function () use ($app) {
     return new Response('', 400);
 });
 
+$app->post('/create_topic_and_subscription', function () use ($app) {
+    /** @var Google_Client $client */
+    $client = $app['google_client'];
+    $projectName = sprintf('projects/%s', $app['project_id']);
+    $pubsub = new Google_Service_Pubsub($client);
+    $topic = new Google_Service_Pubsub_Topic();
+    $topic->setName(sprintf('%s/topics/%s', $projectName, $app['topic']));
+    try {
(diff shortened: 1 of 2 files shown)


pubsub/src/app.php
```diff
@@ -106,4 +106,40 @@
     return new Response('', 400);
 });
 
+$app->post('/create_topic_and_subscription', function () use ($app) {
+    /** @var Google_Client $client */
+    $client = $app['google_client'];
+    $projectName = sprintf('projects/%s', $app['project_id']);
+    $pubsub = new Google_Service_Pubsub($client);
+    $topic = new \Google_Service_Pubsub_Topic();
```
bshaffer: The starting backslash is not required here

tmatsuo:
LGTM

pubsub/web/js/pubsub.js
```diff
@@ -54,6 +55,21 @@ pubsub.PubsubController.prototype.sendMessage = function(message) {
     self.message = null;
   }).error(function(data, status) {
     self.logger.error('Failed to send the message. Status: ' + status + '.');
+    if (!self.createdTopicAndSubscription) {
```
bshaffer: Shouldn't this be `hasAttemptedToCreateTopicAndSubscription`?

pubsub/src/app.php
SurferJeffAtGoogle: Done.

pubsub/web/js/pubsub.js
SurferJeffAtGoogle: Done.

PR #119, opened 2016-06-14
Add environment variable place holders to app.yaml.

SurferJeffAtGoogle:
So each app.yaml file matches its README.

### DIFF
--- a/appengine/flexible/cloudsql/app.yaml
+++ b/appengine/flexible/cloudsql/app.yaml
@@ -3,3 +3,10 @@ vm: true
 
 runtime_config:
   document_root: web
+
+#[START env]
+env_variables:
+  MYSQL_DSN: "mysql:host=127.0.0.1;port=3306;dbname=guestbook"
+  MYSQL_USERNAME: root
+  MYSQL_PASSWORD: 
+#[END env]
--- a/appengine/flexible/analytics/app.yaml
+++ b/appengine/flexible/analytics/app.yaml
@@ -3,3 +3,8 @@ vm: true
 
 runtime_config:
   document_root: web
+
+#[START env]
+env_variables:
+  GA_TRACKING_ID: your-tracking-id
+#[END env]
--- a/appengine/flexible/datastore/app.yaml
+++ b/appengine/flexible/datastore/app.yaml
@@ -3,3 +3,8 @@ vm: true
 
 runtime_config:
   document_root: web
+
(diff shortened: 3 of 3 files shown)


appengine/flexible/cloudsql/app.yaml
```diff
@@ -2,4 +2,11 @@ runtime: php
...
 runtime_config:
-  document_root: web
\ No newline at end of file
+  document_root: web
+
+#[START env]
+env_variables:
+  MYSQL_DSN: "mysql:host=127.0.0.1;port=3306;dbname=guestbook"
+  MYSQL_USERNAME: root
+  MYSQL_PASSWORD:
```
JustinBeckwith: I've been kind of careful in node.js not to encourage adding secrets to app.yaml.  Typically, we tell users to create a .env file (or something similar) for secrets, and make sure it's in the .gitignore.  Maybe we should use this?

https://github.com/vlucas/phpdotenv

SurferJeffAtGoogle: There are nodejs samples that put secrets into app.yaml.  For example:
https://cloud.google.com/appengine/docs/flexible/nodejs/sending-emails-with-mailjet

tmatsuo: I'm fine with `app.yaml`, but if we use this file for secrets, I think we should put `app.yaml` in .gitignore to prevent users from unintentionally submit their secrets to public place.

tmatsuo: If we want to use dotenv, maybe we should support it in the docker image.

bshaffer: I don't think we want to recommend using `dotenv` in production for secrets management. The repo even says to not use it in production, but instead *actually set the environment variables*. 

We have been using `settings.yml.dist`  and ignoring `settings.yml` for the bookshelf, so we could do something along these lines. It's the same amount of overhead (which isn't great) and is more familiar.

JustinBeckwith:
Oh yeah - I'm fine with the yaml approach.  I just don't want secrets in app.yaml.

tmatsuo:
Alright, I like settings.yml too. Being consistent with our getting started guide is nice.

bshaffer:
👍 agreed 👍

bshaffer:
Looks like we are closing this one in favor of `settings.yaml`

PR #120, opened 2016-06-14
Add README and config files for flexible/pubsub.

SurferJeffAtGoogle:
Also, remove the text describing how to create the topic and
subscription, because that no happens automatically.


bshaffer:
This looks great! We should definitely add an e2e test for this, though.

SurferJeffAtGoogle:
Moved to https://github.com/GoogleCloudPlatform/php-docs-samples/pull/124

PR #121, opened 2016-06-14
Added some options for choosing Cloud SQL generation and the db region.

### LINKED ISSUE #68: [App Engine] Wordpress Standard Environment - Functionality to use CloudSQL 2nd gen

tmatsuo:
fixes #68

### DIFF
--- a/appengine/wordpress/src/WordPressSetup.php
+++ b/appengine/wordpress/src/WordPressSetup.php
@@ -263,7 +263,7 @@ class WordPressSetup extends Command
                                      . self::FLEXIBLE_ENV . '.</error>');
                     return self::DEFAULT_ERROR;
                 }
-                $db_region_str = ':';
+                $db_connection_pattern = '%s:%s';
                 break;
             case '2':
                 $db_region = $input->getOption('db_region');
@@ -279,7 +279,7 @@ class WordPressSetup extends Command
                     $output->writeln('Using a db_region: <info>' . $db_region
                                      . '</info>');
                 }
-                $db_region_str = sprintf(":%s:", $db_region);
+                $db_connection_pattern = "%s:$db_region:%s";
                 break;
             default:
                 $output->writeln(
@@ -364,7 +364,11 @@ class WordPressSetup extends Command
         }
         $params = array();
         $this->askParameters($keys, $params, $input, $output, $helper);
-        $params['db_region'] = $db_region_str;
+        $params['db_connection'] = sprintf(
+            $db_connection_pattern,
+            $params['project_id'],
+            $params['db_instance']
+        );
         $q = new ConfirmationQuestion(
             'Do you want to use the same db user and password for '
             . 'local run? (Y/n)',
--- a/appengine/wordpress/src/WordPressSetup.php
+++ b/appengine/wordpress/src/WordPressSetup.php
@@ -253,12 +253,7 @@ class WordPressSetup extends Command
         $output->writeln('Creating a new project for: <info>' . $env
                          . '</info>');
 
-        $output->writeln('Downloading the WordPress archive...');
-        $wpUrl = $input->getOption('wordpress_url');
-        $project->downloadArchive('the WordPress archive', $wpUrl);
-        if (!$this->report($output, $project)) {
-            return self::DEFAULT_ERROR;
-        }
+        // Determine the Cloud SQL Generation to use.
         $sql_gen = $input->getOption('sql_gen');
         switch ($sql_gen) {
             case '1':
@@ -294,6 +289,13 @@ class WordPressSetup extends Command
                     )
                 );
                 return self::DEFAULT_ERROR;
+        }
+
+        $output->writeln('Downloading the WordPress archive...');
+        $wpUrl = $input->getOption('wordpress_url');
+        $project->downloadArchive('the WordPress archive', $wpUrl);
+        if (!$this->report($output, $project)) {
+            return self::DEFAULT_ERROR;
         }
         $output->writeln('Downloading the Batcache plugin...');
         $project->downloadArchive(
--- a/appengine/wordpress/src/WordPressSetup.php
+++ b/appengine/wordpress/src/WordPressSetup.php
@@ -76,9 +76,9 @@ class WordPressSetup extends Command
                 'sql_gen',
                 '',
                 InputOption::VALUE_OPTIONAL,
-                sprintf ('Cloud SQL generation to use; 2: %s, 1: %s',
-                         'Second Generation',
-                         'First Generation'),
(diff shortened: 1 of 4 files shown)


tmatsuo:
The WordPress tests are passing on our jenkins :)

appengine/wordpress/src/WordPressSetup.php
```diff
@@ -64,13 +76,29 @@ protected function configure()
                 self::DEFAULT_DIR
             )
             ->addOption(
+                'sql_gen',
+                '',
+                InputOption::VALUE_OPTIONAL,
+                'Cloud SQL generation to use; 2: '
+                . self::CLOUD_SQL_SECOND_GEN . ', 1: '
+                . self::CLOUD_SQL_FIRST_GEN . '.',
+                null
```
bshaffer: we should actually set the default to `2`, so this is surfaced in the documentation.

appengine/wordpress/src/WordPressSetup.php
```diff
@@ -225,15 +253,43 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        if ($sql_gen === '1') {
+            if ($env === self::FLEXIBLE_ENV) {
+                $output->writeln('<error>You can not use '
+                                 . self::CLOUD_SQL_FIRST_GEN . ' with '
+                                 . self::FLEXIBLE_ENV . '.</error>');
+                return self::DEFAULT_ERROR;
+            }
+            $sql_gen = self::CLOUD_SQL_FIRST_GEN;
+            $db_region_str = ':';
+        } else {
```
bshaffer: How about a `switch` statement with an exception for invalid `sql_gen`?

appengine/wordpress/src/WordPressSetup.php
```diff
@@ -64,13 +76,29 @@ protected function configure()
                 self::DEFAULT_DIR
             )
             ->addOption(
+                'sql_gen',
+                '',
+                InputOption::VALUE_OPTIONAL,
+                'Cloud SQL generation to use; 2: '
```
bshaffer: would definitely rather see this as 

```php
sprintf('Cloud SQL generation to use; 2: %s, 1:%s',
```

appengine/wordpress/src/WordPressSetup.php
```diff
@@ -225,15 +253,43 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-
+        $sql_gen = $input->getOption('sql_gen');
+        if ($sql_gen === '1') {
+            if ($env === self::FLEXIBLE_ENV) {
+                $output->writeln('<error>You can not use '
+                                 . self::CLOUD_SQL_FIRST_GEN . ' with '
+                                 . self::FLEXIBLE_ENV . '.</error>');
+                return self::DEFAULT_ERROR;
+            }
+            $sql_gen = self::CLOUD_SQL_FIRST_GEN;
```
bshaffer: Why do we set `$sql_gen` to this? We do not use it again.

appengine/wordpress/src/WordPressSetup.php
```diff
@@ -40,6 +40,18 @@ class WordPressSetup extends Command
     const FLEXIBLE_ENV = 'Flexible Environment';
     const STANDARD_ENV = 'Standard Environment';
 
+    const CLOUD_SQL_FIRST_GEN = 'Cloud SQL First Generation';
+    const CLOUD_SQL_SECOND_GEN = 'Cloud SQL Second Generation';
```
bshaffer: Not a huge fan of using constants for interface strings. I do not see an advantage to doing this.

appengine/wordpress/src/WordPressSetup.php
tmatsuo: Done

appengine/wordpress/src/WordPressSetup.php
tmatsuo: Done

appengine/wordpress/src/WordPressSetup.php
tmatsuo: Done

appengine/wordpress/src/WordPressSetup.php
tmatsuo: Good catch, done

appengine/wordpress/src/WordPressSetup.php
tmatsuo: Good catch, done

tmatsuo:
@bshaffer Ready for another look, PTAL

tmatsuo:
FYI, I just moved sql_gen part upper than downloading wordpress

tmatsuo:
@bshaffer The tests are passing on our jenkins too :)

appengine/wordpress/src/files/flexible/app.yaml
```diff
@@ -2,7 +2,7 @@ runtime: php
 vm: true
 
 beta_settings:
-  cloud_sql_instances: {{project_id}}:us-central1:{{db_instance}}
+  cloud_sql_instances: {{project_id}}{{db_region}}{{db_instance}}
```
bshaffer: I would rather be explicit with the Instance Connection Name between 1st and 2nd gen. Something like:
```php
switch ($sql_gen) {
    case '1':
        // ...
        $connection_name = sprintf('%s:%s', $project_id, $db_instance);
    case '2':
        // ...
        $connection_name = sprintf('%s:%s:%s', $project_id, $db_region, $db_instance);
```
and then in `app.yaml` / `DB_HOST`:

```yaml
beta_settings:
    cloud_sql_instances: {{connection_name}}
```

tmatsuo: Yeah, I agree, but it's kind of complicated.

I want to fail fast (before downloading WordPress) so I want to check $sql_gen before that. On the other hand, when I check the environment (flexible/standard) we only download the app engine standard, but at that time, wordpress should be there, etc.

Anyways, point taken, I changed it to create a pattern for sprintf.

bshaffer: Your changes work for me! It's much more clear what's going on. Thank you!

bshaffer:
image

PR #122, opened 2016-06-14
Update app.yaml

thesandlord:
Changed sample to use connection name to be inline with the rest of the docs


bshaffer:
LGTM! Will merge once tests pass.

PR #123, opened 2016-06-15
Suppress test when env vars are not available.

tmatsuo:


tmatsuo:
Otherwise this test will fail for pull requests from other repo.

bshaffer:
lgtm

PR #124, opened 2016-06-15
Add README and deploy test for flexible/pubsub.

SurferJeffAtGoogle:


testing/FileUtil.php
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\TestUtils;
+
+class FileUtil
```
bshaffer: It may be better to use the [Filesystem Component](http://symfony.com/doc/current/components/filesystem/introduction.html) for these functions.

bshaffer:
I like this! Although I wish we could have the test run in `appengine/flexible/pubsub`. I am assuming the reason we aren't doing it this way is because of the required `composer install` in a separate directory. I can't really think of a way to make this happen w/o a lot of hassle, so LGTM.

tmatsuo:
LGTM

PR #128, opened 2016-06-17
Twilio sample for app engine flex.

SurferJeffAtGoogle:
Twilio sample for app engine flex.

### DIFF
--- a/appengine/flexible/twilio/README.md
+++ b/appengine/flexible/twilio/README.md
@@ -1,0 +1,57 @@
+# Twilio and Google App Engine Flexible Environment
+
+This sample application demonstrates how to use [Twilio with Google App Engine](https://cloud.google.com/appengine/docs/flexible/php/using-sms-and-voice-services-via-twilio).
+
+## Setup
+
+Before running this sample:
+
+1. You will need a [Twilio account](https://www.twilio.com/user/account).
+1. Update `TWILIO_ACCOUNT_SID` and `TWILIO_AUTH_TOKEN` in `app.yaml` to match your
+   Twilio credentials. These can be found in your [account settings]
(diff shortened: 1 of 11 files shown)


appengine/flexible/twilio/README.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+## Run locally
+
+you can run locally using PHP's built-in web server:
+
+```sh
+export TWILIO_ACCOUNT_SID=your-account-sid
+export TWILIO_AUTH_TOKEN=your-auth-token
+export TWILIO_NUMBER=your-twilio-number
```
tmatsuo: These secrets, now you need to specify them twice, once in the `app.yaml` and once in the terminal. I think we can converge them by having a configuration file template for them like our getting started app.

How about to have `settings.yml.dist` file in the repo which has the placeholders for those secrets and have the app always read from it?

tmatsuo:
LGTM with one minor suggestion!

appengine/flexible/twilio/test/DeployAppEngineFlexTest.php
```diff
@@ -0,0 +1,76 @@
...
+use Google\Cloud\TestUtils\FileUtil;
+use Symfony\Component\Yaml\Yaml;
+
+class DeployAppEngineFlexTest extends \PHPUnit_Framework_TestCase
+{
+    use AppEngineDeploymentTrait;
+
+    public function beforeDeploy()
+    {
+        $tmpDir = FileUtil::cloneDirectoryIntoTmp(__DIR__ . '/..');
```
bshaffer: We should move this class into `google/cloud-tools` at some point

bshaffer:
LGTM!

PR #130, opened 2016-06-23
more actions for storage

bshaffer:


storage/api/downloadObject.php
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+// [START all]
+// composer autoloading
+require_once __DIR__ . '/vendor/autoload.php';
+
+// grab the first argument
+if (empty($argv[3])) {
+    die("usage: php downloadObject [project_id] [bucket_name] [object_name] [optional_download_name]\n");
```
SurferJeffAtGoogle: All the code inside doc tags will be displayed in an iframe that's only 80 characters wide.  When lines are longer than 80 characters, it's annoying to use the horizontal scroll bar to see them.  Can we make all the lines 80 characters or less?

bshaffer: done!

SurferJeffAtGoogle:
LGTM.

PR #135, opened 2016-07-27
Allow phpmyadmin to use 2nd gen instances

tmatsuo:


tmatsuo:
The test succeeded on our jenkins too.

bshaffer:
LGTM

PR #137, opened 2016-07-28
Composer cleanup and update

tmatsuo:


bshaffer:
Unless there's a reason for using tilde instead, we should use `google/apiclient:^2.0`, as this is more strict to semantic versioning.

appengine/flexible/analytics/composer.json
```diff
@@ -6,7 +6,7 @@
         "symfony/css-selector": "^3.1"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4",
-        "google/cloud-tools": "^0.1.0"
+        "google/cloud-tools": "<=1.0",
```
bshaffer: This goes against semantic versioning, are you sure we don't want to just tag this to a specific version?

tmatsuo: I actually want to auto update cloud-tools to the latest until it hits 1.0

tmatsuo:
The e2e tests are also passing on our jenkins!

bshaffer:
LGTM!

PR #144, opened 2016-08-12
[BigQuery] Add Export Command and clean up Doc Blocks

bshaffer:


bigquery/api/README.md
```diff
@@ -137,6 +173,17 @@ Are you sure you want to delete the BigQuery table "my_table"? [y/n]: y
...
+
+List tables for a BigQuery dataset.
+
+```sh
+$ php bigquery.php tables test_dataset1
+test_table1
+test_table2
+test_table3
+```
+
```
tmatsuo: Can you provide help text for each subcommand? If so, can we stop having example output in the README since you can put such output in the help text which is little bit better from maintenance standpoint? My only worry is the document will easily become stale.

bigquery/api/composer.json
```diff
@@ -9,7 +9,18 @@
             "Google\\Cloud\\Samples\\BigQuery\\": "src/"
         },
         "files": [
-            "src/functions.php"
+            "src/functions/browse_table.php",
```
tmatsuo: Why do we need to list all the files here?

bigquery/api/src/BrowseTableCommand.php
```diff
@@ -0,0 +1,110 @@
...
+    <info>php %command.full_name%</info>
+
+EOF
+            )
+            ->addArgument(
+                'dataset.table',
+                InputArgument::REQUIRED,
+                'The dataset to list tables'
+            )
+            ->addOption(
```
tmatsuo: Do you really prefer having the same code again and again everywhere?

tmatsuo: Can gcloud-php guess the project id from gcloud sdk configuration?

bigquery/api/src/BrowseTableCommand.php
```diff
@@ -0,0 +1,110 @@
...
+        };
+
+        $totalRows = paginate_table($projectId, $datasetId, $tableId, $maxResults, $shouldPaginate);
+
+        printf('Found %s row(s)' . PHP_EOL, $totalRows);
+    }
+
+    private function getProjectIdFromGcloud()
+    {
+        exec("gcloud config list --format 'value(core.project)' 2>/dev/null", $output, $return_var);
```
tmatsuo: Doesn't it make more sense to put this code into `gcloud-php`?

bigquery/api/src/DatasetsCommand.php
```diff
@@ -52,22 +54,14 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        $GLOBALS['output'] = $output;
```
tmatsuo: Why do you need $GLOBALS?

bigquery/api/src/functions/functions.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+namespace Google\Cloud\Samples\BigQuery;
+
+/**
+ * these functions are used to send the command's output though the
+ * Console component, but still appear as standard PHP functions in
+ * the code samples.
+ */
+function printf()
```
tmatsuo: Why do you need such a hack in the first place?

bigquery/api/src/BrowseTableCommand.php
bshaffer: We could use `Google\Cloud\ClientTrait::detectProjectId`

bigquery/api/src/DatasetsCommand.php
bshaffer: see `src/functions/functions.php` - it's to use the output class in execution while allowing the samples to use `printr`

bigquery/api/src/functions/functions.php
bshaffer: so samples can be copied w/o having to worry about the output classes

bigquery/api/src/BrowseTableCommand.php
bshaffer: You mean having `--project` defined on each one? Yes, I prefer to define that explicitly for each command.

bigquery/api/src/BrowseTableCommand.php
bshaffer: The problem is, these are commands and not Clients, and also it pulls in a lot of other code we don't need.

bigquery/api/src/functions/functions.php
bshaffer: Basically without it we'd have to pick between not using the `output` classes, which means output from the commands would appear during tests (There's a phpunit command to prevent this, but I always output for debugging purposes, so I'd prefer to not use it), and having the sample code copy-pasteable. I'm not super stoked on this either, but I think it's a good compromise.

bigquery/api/composer.json
bshaffer: Because it's the only way to autoload those functions

bigquery/api/README.md
bshaffer: Good idea!

bigquery/api/src/functions/functions.php
tmatsuo: Do you really see the output from the code when running phpunit without `--verbose` flag? Which version of phpunit are you using?

How about to test the output with [`expectOutputString()`](https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.output)?

bshaffer: I don't want to use expectOutputString when I can mock the output classes, this is a much cleaner way to write tests IMHO

bigquery/api/src/BrowseTableCommand.php
tmatsuo: I prefer having a Trait for maintenance reasons.

Having one more level of indirection is not a big problem in my opinion, compared to the downside the need of changing every file.

bigquery/api/src/functions/functions.php
tmatsuo: I tried `expectOutputString()` and it worked perfectly, and also it suppress the output in question. Why do you want to avoid it?

bigquery/api/src/BrowseTableCommand.php
bshaffer: I respectfully disagree. Yes, it's important to follow DRY principles, but it can be taken too far. If this was the `gcloud` command line tool, I might agree, but this is a simple CLI wrapper. There is no need to abstract away the `--project` option. If changes needed to be made, the amount of time to changing in 8 files is trivial.

So the desired change is to have something like this:

```php
Trait ProjectOptionTrait
{
    private function addProjectOption()
    {
        return $this->addOption(
            'project',
             null,
             InputOption::VALUE_REQUIRED,
             'The Google Cloud Platform project name to use for this invocation. ' .
             'If omitted then the current gcloud project is assumed. '
         );
    }
}
```

And then in each command we'd have the following
```php
class SomeCommand
{
    use ProjectOptionTrait;
    protected function configure()
    {
        //....
        $this->addProjectOption();
    }
}
```

I do not think this is better. It's more confusing.

bigquery/api/src/functions/functions.php
bshaffer: Because output buffers are a hack. It suppresses expected output from functions such as `echo` and `var_dump`. It's much cleaner to use a mocked output class.

bshaffer: I decided this is fine. Added in ec7eba0

bigquery/api/src/BrowseTableCommand.php
bshaffer: Added in ec7eba0

bigquery/api/src/BrowseTableCommand.php
bshaffer: I removed `getProjectFromGcloud` in favor of using the google-cloud `ClientTrait`, but I am leaving the project option explicit. PTAL.

bigquery/api/src/BrowseTableCommand.php
```diff
@@ -0,0 +1,103 @@
...
+                10
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        if (!$projectId = $input->getOption('project')) {
+            if (!$projectId = $this->detectProjectId()) {
+                throw new Exception('Could not derive a project ID from gloud. ' .
```
tmatsuo: TYPO:

gloud -> gcloud

bigquery/api/src/DatasetsCommand.php
```diff
@@ -53,29 +58,11 @@ protected function configure()
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         if (!$projectId = $input->getOption('project')) {
-            if (!$projectId = $this->getProjectIdFromGcloud()) {
-                throw new \Exception('Could not derive a project ID from gloud. ' .
+            if (!$projectId = $this->detectProjectId()) {
+                throw new Exception('Could not derive a project ID from gloud. ' .
```
tmatsuo: TYPO:

gloud -> gcloud

bigquery/api/src/ExportCommand.php
```diff
@@ -0,0 +1,132 @@
...
+                'csv'
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        if (!$projectId = $input->getOption('project')) {
+            if ($projectId = $this->detectProjectId()) {
+                throw new Exception('Could not derive a project ID from gloud. ' .
```
tmatsuo: TYPO:

gloud -> gcloud

bigquery/api/src/ImportCommand.php
```diff
@@ -74,30 +81,23 @@ protected function configure()
...
+            if ($projectId = $this->detectProjectId()) {
                 if ($input->isInteractive()) {
                     $message = sprintf('Import data for project %s? [y/n]: ', $projectId);
                     if (!$question->ask($input, $output, new ConfirmationQuestion($message))) {
                         return $output->writeln('<error>Task cancelled by user.</error>');
                     }
                 }
             } else {
-                throw new \Exception('Could not derive a project ID from gloud. ' .
+                throw new Exception('Could not derive a project ID from gloud. ' .
```
tmatsuo: TYPO:

gloud -> gcloud

bigquery/api/src/QueryCommand.php
```diff
@@ -63,12 +71,10 @@ protected function configure()
...
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        // set the output to globals for functions defined in functions.php
-        $GLOBALS['output'] = $output;
         $question = $this->getHelper('question');
         if (!$projectId = $input->getOption('project')) {
-            if (!$projectId = $this->getProjectIdFromGcloud()) {
-                throw new \Exception('Could not derive a project ID from gloud. ' .
+            if (!$projectId = $this->detectProjectId()) {
+                throw new Exception('Could not derive a project ID from gloud. ' .
```
tmatsuo: TYPO:

gloud -> gcloud

bigquery/api/src/SchemaCommand.php
```diff
@@ -84,8 +98,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $question = $this->getHelper('question');
         if (!$projectId = $input->getOption('project')) {
-            if (!$projectId = $this->getProjectIdFromGcloud()) {
-                throw new \Exception('Could not derive a project ID from gloud. ' .
+            if (!$projectId = $this->detectProjectId()) {
+                throw new Exception('Could not derive a project ID from gloud. ' .
```
tmatsuo: TYPO:

gloud -> gcloud

bigquery/api/src/TablesCommand.php
```diff
@@ -0,0 +1,76 @@
...
+                'If omitted then the current gcloud project is assumed. '
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        if (!$projectId = $input->getOption('project')) {
+            if (!$projectId = $this->detectProjectId()) {
+                throw new Exception('Could not derive a project ID from gloud. ' .
```
tmatsuo: TYPO:

gloud -> gcloud

bigquery/api/test/ProjectsCommandTest.php
```diff
@@ -0,0 +1,58 @@
...
+class ProjectsCommandTest extends \PHPUnit_Framework_TestCase
+{
+    public function testProjects()
+    {
+        if (!$projectId = getenv('GOOGLE_PROJECT_ID')) {
+            $this->markTestSkipped('No project ID');
+        }
+        if (!CredentialsLoader::fromWellKnownFile()) {
+            if ($keyFile = getenv('GOOGLE_KEY_FILE')) {
+                $home = getenv('HOME');
```
tmatsuo: What happens if HOME is not set?

bigquery/api/test/ImportCommandTest.php
```diff
@@ -41,30 +42,56 @@ public static function setUpBeforeClass()
...
-            ->getMock();
-        $import->importFromFile($table, '/this/file/doesnotexist.json');
+        if (!$projectId = getenv('GOOGLE_PROJECT_ID')) {
+            $this->markTestSkipped('No project ID');
+        }
+
+        // run the import
+        $application = new Application();
+        $application->add(new ImportCommand());
+        $commandTester = new CommandTester($application->get('import'));
```
tmatsuo: Do you still need the commandTester?

tmatsuo: nm, it executes the command

bigquery/api/test/ProjectsCommandTest.php
bshaffer: Then it would fail to write to the path. Do you think it should throw an exception instead?

tmatsuo: Is `${HOME}` usually available on windows?

tmatsuo: If we can guarantee that the following mkdir never harm your system, that's fine, but I'd stop executing if HOME is not set.

bshaffer: fine by me

tmatsuo:
LGTM

appengine/standard/http/app.php
```diff
@@ -101,7 +101,7 @@
     $twig = $app['twig'];
 
     # [START guzzle_request]
-    $url = 'http://httpbin.org/post?query=update';
+    $url = 'https://httpbin.org/post?query=update';
```
tmatsuo: As I said before, this test depends on external resource which may change anytime. I think we should use something else that we can fully control.

tmatsuo: It can be done in another PR.

(also it's not related to bigquery, so shouldn't it be another PR?)

bshaffer: I respectfully disagree, and I do not see how this is relevant. I updated this to test HTTPS for a separate issue, it has nothing to do with a problem using an external resource.

tmatsuo: It's not relevant, but we should not rely on external site in general

PR #146, opened 2016-08-18
Update wp-cli to the latest version

tmatsuo:
Seems like we need to update `wp-cli` for newer WordPress.


tmatsuo:
Seems like Travis test for php-nightly is failing. Can we ignore tests for php-nightly?

.travis.yml
```diff
@@ -25,6 +25,8 @@ matrix:
     - php: 5.6
     - php: hhvm
     - php: nightly
+  allow_failures:
+    - php: nightly
```
bshaffer: I do not think we should do this...

bshaffer: Oh sorry, I thought this was for **our** nightly builds. We should add PHP 7 and remove nightly.

tmatsuo: Ah good point. Done

tmatsuo:
@bshaffer PTAL

bshaffer:
LGTM!

PR #152, opened 2016-08-27
Add option to use standard SQL syntax to BigQuery sample.

tswast:
This is so that we can include PHP in the list of languages when we add veneer samples to https://cloud.google.com/bigquery/sql-reference/enabling-standard-sql.

I fully expect that I have a style issue on every single line. The last time I used PHP, the latest version was 4.3.


tswast:
(I guess PHP 5.0 was out, but no hosting company had support for it.)

tmatsuo:
Seems like the linter likes your code :)

tmatsuo:
LGTM, thanks

PR #153, opened 2016-09-01
Adding the first snippet for the Logging API

tmatsuo:


logging/src/GoogleProjectTrait.php
```diff
@@ -0,0 +1,70 @@
...
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\Samples\Logging;
+
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+
+/**
+ * Trait for determining the current project ID using "gcloud"
```
SurferJeffAtGoogle: Add a note that the class using the trait must inherit from Symfony\Component\Console\Command\Command.

logging/src/GoogleProjectTrait.php
```diff
@@ -0,0 +1,70 @@
...
+            'The Google Cloud Platform project name to use for this command. ' .
+            'If omitted then the current gcloud project is assumed. ',
+            $this->getProjectIdFromGcloud()
+        );
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function interact(InputInterface $input)
```
SurferJeffAtGoogle: What if multiple traits implement interact?

logging/src/GoogleProjectTrait.php
tmatsuo: I started to think I should make this a subclass of `Symfony\Component\Console\Command\Command`. I'll change it so in the next push.

logging/src/GoogleProjectTrait.php
```diff
@@ -0,0 +1,70 @@
...
+            null,
+            InputOption::VALUE_REQUIRED,
+            'The Google Cloud Platform project name to use for this command. ' .
+            'If omitted then the current gcloud project is assumed. ',
+            $this->getProjectIdFromGcloud()
+        );
+    }
+
+    /**
+     * @inheritdoc
```
tmatsuo: Good point, I'll subclass so it'll make sense

logging/src/GoogleProjectTrait.php
tmatsuo: http://php.net/manual/en/language.oop5.traits.php
You have to resolve conflict manually, or it'll throw an error

tmatsuo:
@SurferJeffAtGoogle Thanks! PTAL

SurferJeffAtGoogle:
LGTM.

tmatsuo:
Thanks!

logging/logging-sample-cli.php
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env php
```
bshaffer: We've been using underscores for PHP filesnames.

logging/src/CommandWithProject.php
```diff
@@ -0,0 +1,68 @@
...
+namespace Google\Cloud\Samples\Logging;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+
+/**
+ * A base class for commands which needs project id.
+ */
+class CommandWithProject extends Command
```
bshaffer: This should be abstract

logging/src/Write.php
```diff
@@ -0,0 +1,66 @@
...
+            );
+        $this->addProjectOption();
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $message = $input->getArgument("message");
+        $project = $input->getOption('project');
+        // [START write_log]
+        $logging = new LoggingClient(['projectId' => $project]);
```
bshaffer: This should be `$projectId` to be consistent with the rest of the samples.

logging/src/Write.php
```diff
@@ -0,0 +1,66 @@
...
+                InputArgument::OPTIONAL,
+                "The log message to write",
+                "Hello"
+            );
+        $this->addProjectOption();
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $message = $input->getArgument("message");
```
bshaffer: This should be single quotes.

logging/src/CommandWithProject.php
```diff
@@ -0,0 +1,68 @@
...
+        exec(
+            "gcloud config list --format 'value(core.project)' 2>/dev/null",
+            $output,
+            $return_var
+        );
+        if (0 === $return_var) {
+            return array_pop($output);
+        }
+        return null;
+    }
```
bshaffer: did we actually abstract an entire class for a single command?

This adds indirection and abstraction without even the advantage of DRY. Although even w/ DRY it's not worth it. For samples especially, gaining clarity is worth repeating code.

logging/src/Write.php
tmatsuo: will fix in the next PR

logging/src/Write.php
tmatsuo: will fix in the next PR

logging/src/CommandWithProject.php
tmatsuo: I will add few more commands

logging/src/CommandWithProject.php
tmatsuo: will do

logging/logging-sample-cli.php
tmatsuo: will do

PR #154, opened 2016-09-07
Add region tags to BigQuery run_query_as_job.

tswast:
This will be used to show how to enable standard SQL.
https://cloud.google.com/bigquery/sql-reference/enabling-standard-sql
I chose the name based on the similar run_query region tags in
https://github.com/GoogleCloudPlatform/php-docs-samples/blob/60bf9879ad85475c20b2f23c9ad989b85296ed1a/bigquery/api/src/functions/run_query.php#L51


tmatsuo:
LGTM

bshaffer:
I think the desired tag was `query_as_job`

PR #156, opened 2016-09-09
Adds expiration to Memcached#set()

tmatsuo:
This should fix the flaky test with this app.

See also:
https://github.com/GeckoPackages/GeckoMemcacheMock/issues/5


tmatsuo:
@bshaffer Please take a look

@SurferJeffAtGoogle FYI

tmatsuo:
The test is flaky and you can reproduce with running many times like:

```bash
for x in `seq 1 10000`; do phpunit > /dev/null || echo 'failed'; done
```

bshaffer:
LGTM!

PR #158, opened 2016-09-09
removes unneeded google/apiclient dependency

bshaffer:


tmatsuo:
Maybe you can rebase to the head? Otherwise, travis won't run the tests.

tmatsuo:
LGTM

PR #159, opened 2016-09-09
Adds Speech Translate Command

bshaffer:


speech/api/phpunit.xml.dist
```diff
@@ -0,0 +1,31 @@
...
+        <testsuite name="PHP speech test">
+            <directory>test</directory>
+        </testsuite>
+    </testsuites>
+    <logging>
+        <log type="coverage-clover" target="build/logs/clover.xml"/>
+    </logging>
+    <filter>
+        <whitelist>
+            <file>speech.php</file>
```
tmatsuo: Add `src` directory too

speech/api/composer.json
```diff
@@ -0,0 +1,14 @@
+{
+    "require": {
+        "google/cloud": "0.8",
+        "symfony/console": "^3.0"
```
tmatsuo: Is `ext-grpc` mandatory? If not, is it worthwhile to `suggest` it?

speech/api/phpunit.xml.dist
bshaffer: We shouldn't whitelist the `src` directory because we need to make sure we have coverage.

speech/api/composer.json
bshaffer: It isn't mandatory, but it's required for streaming so I can suggest it. I am going to add the streaming part in another PR.

speech/api/phpunit.xml.dist
tmatsuo: Unless we add `src`, the coverage report won't measure the coverage for files there.

bshaffer: ahh ok, then `speech.php` shouldn't be there. Fixed.

speech/api/src/functions/base64_encode_audio.php
```diff
@@ -0,0 +1,17 @@
...
+ * Usage:
+ * ```
+ * $audioFile = '/path/to/YourAudio.raw';
+ * $base64Audio = require '/path/to/base64_encode_audio.php';
+ * ```
+ */
+# [START base64_audio]
+$base64Audio = base64_encode(stream_get_contents($audioFile));
+# [end base64_audio]
+return $base64Audio;
```
tmatsuo: small nit: newline?

tmatsuo:
LGTM

speech/api/src/functions/base64_encode_audio.php
bshaffer: done

bshaffer:
flushed out `README.md`, @tmatsuo PTAL

speech/api/README.md
```diff
@@ -0,0 +1,46 @@
...
+## Installation
+
+Install the dependencies for this library via [composer](https://getcomposer.org)
+
+    $ cd /path/to/php-docs-samples/speech/api
+    $ composer install 
+
+Authenticate your project using [gcloud](https://cloud.google.com/sdk/gcloud/)
+
+    $ gcloud config set project YOUR_PROJECT_ID
```
tmatsuo: "Authenticate your project" sounds weird. Maybe just "Configure your project"?

bshaffer: good call

PR #160, opened 2016-09-12
Adding commands for log export.

tmatsuo:


logging/src/DeleteSinkCommand.php
```diff
@@ -0,0 +1,60 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\Samples\Logging;
+
+// [START delete_sink_use]
+use Google\Cloud\Logging\LoggingClient;
```
bshaffer: In the docs, are you planning on including these inline with the code below (so they can be copied together) or are you planning on having this be above the header block for the sample below?

I like this method, but I'm worried about docs formatting and the inconsistency with our other samples.

logging/src/DeleteSinkCommand.php
```diff
@@ -0,0 +1,60 @@
...
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $sinkName = $input->getOption('sink');
+        $projectId = $input->getOption('project');
+        // [START delete_sink]
+        $logging = new LoggingClient(['projectId' => $projectId]);
+        $logging->sink($sinkName)->delete();
+        // [END delete_sink]
+        printf("Deleted a sink '%s'.\n", $sinkName);
```
bshaffer: optional nit: I've been using `PHP_EOL` instead of "\n".

logging/src/DeleteSinkCommand.php
tmatsuo: My plan is to add 2 gitincludes, one for the `use` above with the github header, and the other without the github header which I hope rendered nicely.

logging/src/DeleteSinkCommand.php
tmatsuo: Done

tmatsuo:
@bshaffer PTAL

logging/src/DeleteSinkCommand.php
bshaffer: I've tried this and there's a ~5px white gap between them which I find very annoying. This is why I switched to function includes. But we could possibly fix the rendering, so it's up to you.

logging/README.md
```diff
@@ -0,0 +1,27 @@
+# Stackdriver Logging v2 API Samples
+
+`logging_sample_cli.php` is a simple command-line program to demonstrate writing
```
bshaffer: I've called them `bigquery.php` and `speech.php`.  Do you know what other languages are doing? I like this convention since it makes it clear this is a sample, but we should be consistent.

tmatsuo: Well, it's already not consistent among languages. For example, Python has `logging/cloud-client/snippets.py` and nodejs has `logging/logs.js`.

bshaffer: personally I'm in favor of calling it `logging.php` since it's implicitly a sample (because of the repo name) and a CLI (from the README).

bshaffer:
One minor rename suggestion, otherwise LGTM

tmatsuo:
Thanks, merging

PR #162, opened 2016-09-14
the Speech Veneer does not care about your project ID

bshaffer:


tmatsuo:
The code LGTM

Shouldn't google cloud library pass the projectId to the API?

bshaffer:
The issue is the [Speech API](https://cloud.google.com/speech/reference/rest/) does not accept a Project ID parameter

PR #163, opened 2016-09-15
adds pubsub for veneer, updates pubsub app and README

bshaffer:


pubsub/app/app.yaml
```diff
@@ -0,0 +1,13 @@
+runtime: php55
+api_version: 1
+threadsafe: true
+module: pubsub
```
tmatsuo: Now it should be `service` instead of `module`

pubsub/cli/src/functions/create_push_subscription.php
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/pubsub/README.md
+ */
+
+namespace Google\Cloud\Samples\PubSub;
+
+# [START create_push_subscription]
+use Google\Cloud\ServiceBuilder;
```
tmatsuo: How about to use `Google\Cloud\PubSub\PubSubClient` like:
```php
use Google\Cloud\PubSub\PubSubClient;
$pubSub = new PubSubClient([
    'projectId' => 'my_project'
]);

// Get an instance of a previously created topic.
$topic = $pubSub->topic('my_topic');

// Publish a message to the topic.
$topic->publish([
    'data' => 'My new message.',
    'attributes' => [
        'location' => 'Detroit'
    ]
]);
```
instead of using `Google\Cloud\ServiceBuilder`?

PR #164, opened 2016-09-16
Add a Google Cloud Natural Language API sample.

SurferJeffAtGoogle:


language/api/src/AnalyzeEntitiesCommand.php
```diff
@@ -0,0 +1,59 @@
...
+                'Text to analyze'
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $text = implode(" ", $input->getArgument('text'));
+        $result = analyze_entities($text);
+        print_r($result->info());
```
bshaffer: In the other samples, we have the output take place inside the included functions. This is so that the copy/paste sample outputs something meaningful by default. 

`print_r` is pretty generic, could we output something more structured? Something like:

```php
printf('Name: %s' . PHP_EOL, $annotation['name']);
printf('Tag: %s' . PHP_EOL, $annotation['tag']);
printf('Sentiment: polarity of %s with magnitude of %s' . PHP_EOL,
    $annotation['documentSentiment']['polarity'],
    $annotation['documentSentiment']['magnitude']);
```

bshaffer:
See the [output from `php-cs-fixer`](https://travis-ci.org/GoogleCloudPlatform/php-docs-samples/jobs/160581441#L455) to fix the CS. You can [install php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) and run it like this:

```
cd php-docs-samples/language/api
php-cs-fixer fix --config-file=../../.php_cs
```

language/api/src/AnalyzeEntitiesCommand.php
bshaffer: What if we did something like this for syntax:

```php
print('TOKENS' . PHP_EOL);
$info = $result->info();
foreach ($info['tokens'] as $token) {
    printf('%s - %s' . PHP_EOL, $token['text']['content'], $token['text']['partOfSpeech']);
}
```

And something similar for entities?

SurferJeffAtGoogle: I feel a tug of war between good coding discipline and a good demonstration of the code.  Coding discipline dictates that a function should return values, not print values.  Command handlers should print values.  In other samples, I think the functions were quite illustrative.  But in this sample, the functions are only 3 lines with no branches.  What do you think of just moving the code into the commands and deleting the functions? The unit tests that test the commands already 100% cover the functions.

Regarding the print_r, will PHP programmers be familiar with the print_r output?  As a reader, I definitely want to see every annotation that language API can give me, and print_r shows me every detail, and exactly where to find it.  Also, writing custom print statements means more code for us to maintain, and that print code would be unlikely to be copied and pasted into user code.

language/api/src/functions/print_annotation.php
```diff
@@ -0,0 +1,56 @@
...
+use Google\Cloud\NaturalLanguage\Annotation;
+
+function print_annotation(Annotation $annotation) {
+    $info = $annotation->info();
+    if (isset($info['language'])) {
+        print "language: " . $info['language'] . "\n";
+    }
+
+    if (isset($info['documentSentiment']))
+    {
```
bshaffer: This will cause the linter to fail. They need to be cuddly!

language/api/src/functions/print_annotation.php
```diff
@@ -0,0 +1,56 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\Samples\Language;
+
+use Google\Cloud\NaturalLanguage\Annotation;
+
+function print_annotation(Annotation $annotation) {
```
bshaffer: Curly brace here needs to be on its own line.

language/api/src/functions/print_annotation.php
```diff
@@ -0,0 +1,56 @@
...
+                $sentence['text']['content'] . "\n";
+        }
+    }
+
+    if (isset($info['entities']))
+    {
+        print "entities:\n";
+        foreach ($info['entities'] as $entity) {
+            print "  " . $entity['name'];
+            if (isset($entity['metadata']['wikipedia_url']))
```
bshaffer: Always use curly braces even for one-liners.

language/api/src/functions/print_annotation.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+namespace Google\Cloud\Samples\Language;
+
+use Google\Cloud\NaturalLanguage\Annotation;
+
+function print_annotation(Annotation $annotation) {
+    $info = $annotation->info();
+    if (isset($info['language'])) {
+        print "language: " . $info['language'] . "\n";
```
bshaffer: We've been using single quotes for all strings, and `PHP_EOL` instead of `\n`

PR #170, opened 2016-09-27
fixes speech sample README links

bshaffer:


tmatsuo:
Do you prefer having the full URL in the PHP doc? Can it be something like "refer to the README.md in this directory" or something? Whenever you change the directory structure, it will break again.

bshaffer:
We could have something like "see the README in the root of this project." but I think that is less convenient than providing the link directly.

tmatsuo:
LGTM go ahead

PR #172, opened 2016-09-28
Added basic code snippets for Cloud Datastore

tmatsuo:
@bshaffer JFYI, the first commit of the Datastore sample


tmatsuo:
The tests are failing because of https://github.com/GoogleCloudPlatform/google-cloud-php/issues/173

jdpedrie:
Please see GoogleCloudPlatform/google-cloud-php#174.

tmatsuo:
@jdpedrie Thanks! The tests are passing now :)

tmatsuo:
BTW, @bshaffer 
The coverage of pubsub is very low on travis. I think it's because there are no environment variables like GOOGLE_PUBSUB_TOPIC. Of course, we can set those envvars, but how about just use `uniqid()` or something for topic and subscription names?

tmatsuo:
This PR includes snippets on this page:
https://cloud.google.com/datastore/docs/concepts/entities

@bshaffer Please take a look!

datastore/src/functions/concepts.php
```diff
@@ -0,0 +1,197 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\Samples\Datastore;
+
+require_once __DIR__ . '/../../vendor/autoload.php';
```
bshaffer: we don't need an autoload include here, since these functions are handled by the autoloader.

datastore/src/functions/concepts.php
```diff
@@ -0,0 +1,197 @@
...
+
+// [START datastore_use ]
+use Google\Cloud\Datastore\DatastoreClient;
+
```
bshaffer: Not a fan of the whitespace here.

datastore/src/functions/concepts.php
```diff
@@ -0,0 +1,197 @@
...
+    // [START init_client]
+    $datastore = new DatastoreClient();
+    // [END init_client]
+    return $datastore;
+}
+
+function upsert_entity(DatastoreClient $datastore)
+{
+    // [START create_entity]
+    $task = $datastore->entity('Task', [
```
bshaffer: This should be in its own method, or combined with the `upsert_entity` docblock below. Also isn't this more `create_task` than create entity? And what is the difference between this and `insert_entity`?

datastore/src/functions/concepts.php
```diff
@@ -0,0 +1,197 @@
...
+    // [END delete_entity]
+}
+
+function upsert_multi(DatastoreClient $datastore, array $tasks)
+{
+    // [START upsert_multi]
+    $datastore->upsertBatch($tasks);
+    // [END upsert_multi]
+}
+
```
bshaffer: Would love to see an example of `$keys`, i.e. 
```php
/**
 * Example:
 * ```
 *  $keys = [1,2,3];
 * lookup_multi($datastore, $keys);
 * ```
 */
function lookup_multi(...)
```

datastore/src/functions/concepts.php
```diff
@@ -0,0 +1,197 @@
...
+{
+    // [START delete_multi]
+    $datastore->deleteBatch($keys);
+    // [END delete_multi]
+}
+
+function create_complete_key(DatastoreClient $datastore)
+{
+    // [START complete_key]
+    $taskKey = $datastore->key('Task', 'sampleTask');
```
bshaffer: what is a `complete_key`? I believe this should be `named_key`

tmatsuo: `complete_key` is a datastore term, a key with fully populated key whereas incomplete keys only have Kind.

datastore/src/functions/concepts.php
tmatsuo: The document has section for creating keys in detail. I think it's easy enough to guess it's an array of `Key`s.

datastore/src/functions/concepts.php
tmatsuo: Ok, I separated them. The name `create_entity` is OK.

datastore/src/functions/concepts.php
tmatsuo: Me neither, php-cs-fixer complains.

datastore/src/functions/concepts.php
tmatsuo: Good point, removed.

tmatsuo:
@bshaffer PTAL

tmatsuo:
Rebased and squashed

datastore/src/functions/concepts.php
bshaffer: Apparently not, as I was wrong. The least you can do is add a param doc:

```php
/**
 * @param array<Google\Cloud\Datastore\Key> $keys
 */
function lookup_multi(...)
```

datastore/src/functions/concepts.php
bshaffer: [Python](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/datastore/api/snippets.py#L31)
[NodeJS](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/datastore/concepts.js)
[Go](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/datastore/snippets/snippet_test.go#L39)
[Java](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/datastore/src/main/java/com/google/datastore/snippets/Concepts.java#L152)
[C#](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/datastore/api/TaskListTest/DatastoreTest.cs#L78)

Let's be consistent with the other language samples.

datastore/README.md
```diff
@@ -1,82 +0,0 @@
-# Google Datastore PHP Sample Application
```
bshaffer: We need to add a README explaining that this is and how to run the tests

bshaffer:
To separate from the `quickstart` directory, and to be consistent with the other samples (bigquery, speech, and storage), this whole sample should be moved into an `/api/` directory.

bshaffer:
Handful of straightforward changes, otherwise LGTM!

datastore/src/functions/concepts.php
tmatsuo: Done

datastore/src/functions/concepts.php
tmatsuo: Added docs

tmatsuo:
@bshaffer Alright, moved to the api, and added README. Once it looks good to you, can you approve with the new github review feature?

datastore/api/src/functions/concepts.php
```diff
@@ -0,0 +1,297 @@
...
+    $datastore->upsertBatch($tasks);
+    // [END upsert_multi]
+}
+
+/**
+ * Lookup multiple entities.
+ *
+ * @param DatastoreClient $datastore
+ * @param array <Key> $keys
+ * @return array <Google\Cloud\Datastore\Entity>
```
bshaffer: This may need a leading `\`

bshaffer:
I have one comment regarding the docblock, otherwise 👍

PR #173, opened 2016-09-28
Adds Storage Veneer samples and CLI

bshaffer:


bshaffer:
NOTE: These samples will need to be updated for https://github.com/GoogleCloudPlatform/google-cloud-php/pull/169 once a new tag is available.

storage/api/src/AclCommand.php
```diff
@@ -0,0 +1,134 @@
...
+            )
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $bucketName = $input->getArgument('bucket');
+        $entity = $input->getOption('entity');
+        $role = $input->getOption('role');
+        if ($objectName = $input->getArgument('object')) {
```
tmatsuo: The if-else clauses here makes me nervous.
Can it be simpler? How about to split this command into 3 or 4 commands like `object-acl`, `default-acl`, `show-acl`?

tmatsuo: sounds good

bshaffer: PTAL

storage/api/src/EncryptionCommand.php
```diff
@@ -0,0 +1,103 @@
...
+                'Path to store the dowloaded file'
+            )
+            ->addOption(
+                'key',
+                null,
+                InputOption::VALUE_REQUIRED,
+                'Supply your encryption key'
+            )
+            ->addOption(
+                'generate',
```
tmatsuo: Nit: So it's `encryption --generate`. I would say `--generate-key`

storage/api/src/functions/add_object_acl.php
```diff
@@ -0,0 +1,50 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/storage/api/README.md
```
tmatsuo: Does the README really have the instruction? How about to point them to our docs?

storage/api/src/functions/rotate_encryption_key.php
```diff
@@ -0,0 +1,41 @@
...
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/storage/api/README.md
+ */
+
+namespace Google\Cloud\Samples\Storage;
+
+use Exception;
+
+# [START rotate_encryption_key]
```
tmatsuo: Do you want to put this snippet on our docs?

storage/api/test/BucketAclCommandTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+        $this->commandTester->execute(
+            [
+                'bucket' => $bucketName,
+                '--entity' => 'allAuthenticatedUsers',
+                '--delete' => true,
+            ],
+            ['interactive' => false]
+        );
+        $this->expectOutputRegex("/Deleted allAuthenticatedUsers from \S+ ACL/");
```
tmatsuo: This test will pass when the API call silently fails. Ideally can we test the actual behavior? Maybe you can file a bug for it and tackle it later.

storage/api/test/ObjectsCommandTest.php
```diff
@@ -0,0 +1,161 @@
...
+
+        $this->commandTester->execute(
+            [
+                'bucket' => $bucketName,
+                'object' => $objectName . '-copy',
+                '--delete' => true,
+            ],
+            ['interactive' => false]
+        );
+        $this->expectOutputRegex("/Deleted \S+$basename-copy/");
```
tmatsuo: Be careful about calling `expectOutputRegex` multiple times. Only the last call will take effect.

storage/api/src/functions/rotate_encryption_key.php
bshaffer: We might as well leave it in - it'll be supported by the veneers soon.

storage/api/src/functions/add_object_acl.php
bshaffer: The docs don't show how to run the CLI

storage/api/test/BucketAclCommandTest.php
bshaffer: This is now fixed!

storage/api/test/ObjectsCommandTest.php
bshaffer: Good catch, thanks. These are now all tackled in a single assert.

storage/api/src/functions/add_object_acl.php
tmatsuo: So how about the README? It's now almost empty.

storage/api/test/BucketAclCommandTest.php
tmatsuo: Hmm, it's not. I mean, for example, if the following call:

```php
$acl->add($entity, $role, $options);
```
doesn't raise any exception, but it fails to actually add the acl for some reason, our test will pass, but it's not working.

I think the ideal test should check if the acl is actually created as expected.

I'm fine with the current test for the initial version though.

tmatsuo:
Personally I don't like the date driven development because it sacrifices quality.
LGTM if you want to meet the deadline with following TODOs:

- More appropriate test
- README that actually instruct how to run the cli
- nit, but `generate` -> `generate-key`

storage/api/test/BucketAclCommandTest.php
bshaffer: If the API call to `$acl->add` fails, it *will* throw an exception. These tests are not to test the underlying `google-cloud` library, but rather to test the samples. I believe the tests are sufficient.

tmatsuo: If it fails with an exception, yes you're right. But potentially it can fail without throwing any exceptions. For example, the code may put the argument in a wrong order, normally it will throw an exception, but potentially they are legal, etc.

Anyway, you can merge and discuss this outside of this PR

PR #174, opened 2016-09-29
Add new "quickstart" samples.

jmdobry:
These are intended to be simple examples that explicitly show:

* Importing the client library
* Instantiating a client
* Making a single API call
* Printing the result

**Product covered:**

* BigQuery
  * [x] - sample
  * [x] - test
* Datastore
  * [x] - sample
  * [x] - test
* Language
  * [x] - sample
  * [x] - test
* Logging
  * [x] - sample
  * [x] - test
* Pub/Sub
  * [x] - sample
  * [x] - test
* Speech
  * [x] - sample
  * [x] - test
* Storage
  * [x] - sample
  * [x] - test
* Translate
  * [x] - sample
  * [x] - test
* Vision
  * [x] - sample
  * [x] - test

For reference, see the corresponding samples in other languages:

* [GoogleCloudPlatform/nodejs-docs-samples](https://github.com/GoogleCloudPlatform/nodejs-docs-samples) (samples are on `master` in various `quickstart.js` files)
* GoogleCloudPlatform/python-docs-samples#547
* GoogleCloudPlatform/ruby-docs-samples#77
* GoogleCloudPlatform/golang-samples#131
* GoogleCloudPlatform/java-docs-samples (pending)
* GoogleCloudPlatform/dotnet-docs-samples (pending)

Thanks!


jmdobry:
How do you want these `quickstart.php` files organized? I've just been putting them into their own `quickstart` folder within each product folder, e.g.:

```
bigquery/
  quickstart/
    composer.json
    composer.lock
    quickstart.php
```

bigquery/quickstart/quickstart.php
```diff
@@ -0,0 +1,22 @@
...
+
+# [START bigquery_quickstart]
+# Imports the Google Cloud client library
+use Google\Cloud\BigQuery\BigQueryClient;
+
+# Your Google Cloud Platform project ID
+$projectId = 'YOUR_PROJECT_ID';
+
+# Instantiates a client
+$bigqueryClient = new BigQueryClient([
```
tmatsuo: The google-cloud-php can guess the projectId most of the cases.
I would omit the projectId, which makes the code more testable.

Same for the rest

datastore/quickstart/quickstart.php
```diff
@@ -0,0 +1,26 @@
...
+
+# The kind of the entity to retrieve
+$kind = 'Task';
+# The id of the entity to retrieve
+$id = 1234567890;
+# The Datastore key for the entity
+$taskKey = $datastoreClient->key($kind, $id);
+
+# Retrieves the entity
+$dataset = $datastoreClient->lookup($taskKey);
```
tmatsuo: `$dataset` -> `$entity`

Considering it's the first example for users, I don't think `lookup` is appropriate, because it will return `null` for the first time users.

bigquery/quickstart/quickstart.php
bshaffer: The quickstarts should follow the same rules as the samples, and rely on the implicitly set project ID, rather than setting it explicitly, as this should be handled by the Client Libraries page.

datastore/quickstart/quickstart.php
```diff
@@ -0,0 +1,26 @@
...
+
+# [START datastore_quickstart]
+# Imports the Google Cloud client library
+use Google\Cloud\Datastore\DatastoreClient;
+
+# Your Google Cloud Platform project ID
+$projectId = 'YOUR_PROJECT_ID';
+
+# Instantiates a client
+$datastoreClient = new DatastoreClient([
```
bshaffer: `$datastoreClient` => `$datastore` for consistency with our other samples.

bshaffer:
@tmatsuo PTAL

bigquery/quickstart/test/quickstartTest.php
```diff
@@ -0,0 +1,39 @@
...
+        );
+        file_put_contents($file, $contents);
+
+        // Invoke quickstart.php
+        $dataset = include $file;
+
+        // Make sure it looks correct
+        $this->assertInstanceOf('Google\Cloud\BigQuery\Dataset', $dataset);
+        $this->assertEquals($datasetId, $dataset->id());
+        $dataset->delete();
```
tmatsuo: Is it more appropriate to call `delete()` in `tearDown` or something?

datastore/quickstart/quickstart.php
```diff
@@ -6,21 +6,17 @@
...
 $kind = 'Task';
 # The id of the entity to retrieve
 $id = 1234567890;
 # The Datastore key for the entity
-$taskKey = $datastoreClient->key($kind, $id);
+$taskKey = $datastore->key($kind, $id);
 
 # Retrieves the entity
-$dataset = $datastoreClient->lookup($taskKey);
+$entity = $datastore->entity($taskKey);
```
tmatsuo: This doesn't retrieve the entity, it just creates a local `Entity` object. No API calls are made.

datastore/quickstart/quickstart.php
```diff
@@ -6,21 +6,17 @@
...
 # Instantiates a client
-$datastoreClient = new DatastoreClient([
-    'projectId' => $projectId
-]);
+$datastore = new DatastoreClient();
 
 # The kind of the entity to retrieve
 $kind = 'Task';
 # The id of the entity to retrieve
 $id = 1234567890;
```
tmatsuo: It's a bit controversial to manually assign the numeric id (since you need to `allocateId` to avoid the collision with auto assigned id).

datastore/quickstart/test/quickstartTest.php
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+class quickstartTest extends PHPUnit_Framework_TestCase
+{
+    public function testQuickstart()
+    {
+        $file = sys_get_temp_dir() . '/datastore_quickstart.php';
+        $contents = file_get_contents(__DIR__ . '/../quickstart.php');
+        $contents = str_replace(
```
tmatsuo: I don't think you need to replace anything

logging/quickstart/quickstart.php
```diff
@@ -0,0 +1,20 @@
...
+# Your Google Cloud Platform project ID
+$projectId = 'YOUR_PROJECT_ID';
+
+# Instantiates a client
+$logging = new LoggingClient([
+    'projectId' => $projectId
+]);
+
+# List logging entries
+$entries = $logging->entries();
```
tmatsuo: Nit: this will give you nothing, right? Is it appropriate for quickstart?

logging/quickstart/quickstart.php
```diff
@@ -0,0 +1,20 @@
...
+# [START logging_quickstart]
+# Imports the Google Cloud client library
+use Google\Cloud\Logging\LoggingClient;
+
+# Your Google Cloud Platform project ID
+$projectId = 'YOUR_PROJECT_ID';
+
+# Instantiates a client
+$logging = new LoggingClient([
+    'projectId' => $projectId
```
tmatsuo: I think we can remove $projectId here

logging/quickstart/test/quickstartTest.php
```diff
@@ -0,0 +1,42 @@
...
+    public function testQuickstart()
+    {
+        if (!$projectId = getenv('GOOGLE_PROJECT_ID')) {
+            $this->markTestSkipped('GOOGLE_PROJECT_ID must be set.');
+        }
+
+        $file = sys_get_temp_dir() . '/logging_quickstart.php';
+        $contents = file_get_contents(__DIR__ . '/../quickstart.php');
+        $contents = str_replace(
+            ['YOUR_PROJECT_ID', '__DIR__'],
```
tmatsuo: ditto

pubsub/quickstart/test/quickstartTest.php
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+class quickstartTest extends PHPUnit_Framework_TestCase
+{
+    public function testQuickstart()
+    {
+        $topicName = 'my-new-topic-' . time();
+        $file = sys_get_temp_dir() . '/pubsub_quickstart.php';
+        $contents = file_get_contents(__DIR__ . '/../quickstart.php');
+        $contents = str_replace(
```
tmatsuo: ditto

jmdobry:
> The quickstarts should follow the same rules as the samples, and rely on the implicitly set project ID, rather than setting it explicitly, as this should be handled by the Client Libraries page.

These sample _are_ for the Client Libraries page, therefore, can you bring back the explicit project IDs? These quickstart samples (which will only be shown on the Client Libraries pages) should be the only ones that use an explicit project ID. All other samples (as in, everything _but_ the quickstart samples) should infer project ID.

p.s. Thanks for writing the tests!

Also, the various quickstart samples (Node.js, Ruby, Python) are nearly identical in terms of variable names, number of executable statement, comments in the code, etc., so if possible, it would be best to keep these PHP versions of those quickstart samples as close to the Node.js, Python, Ruby quickstart samples as possible.

We have specific requests from UX/PM for these samples to look a certain way, use certain comments, etc. I think it's okay for the quickstart samples to be "quirky" (they're only for the Client Libraries pages), meaning they don't follow the exact same conventions as the rest of your samples.

For example:

Node.js

```js
// Your Google Cloud Platform project ID
const projectId = 'YOUR_PROJECT_ID';

// Instantiates a client
const bigqueryClient = BigQuery({ projectId: projectId });
```

Python

```python
# Your Google Cloud Platform project ID
project_id = 'YOUR_PROJECT_ID'

# Instantiates a client
bigquery_client = BigQuery.Client(project=project_id)
```

Ruby

```ruby
# Your Google Cloud Platform project ID
project_id = "YOUR_PROJECT_ID"

# Instantiates a client
gcloud = Google::Cloud new project_id
bigquery_client = gcloud.bigquery
```

PHP

```php
# Your Google Cloud Platform project ID
$projectId = 'YOUR_PROJECT_ID';

# Instantiates a client
$bigqueryClient = new BigQueryClient([
    'projectId' => $projectId
]);
```

Yes, they're harder to test with an explicit project ID, but there shouldn't be _too_ many of these types of sample.

jmdobry:
Maybe "quickstart" was the wrong label for these samples?

tmatsuo:
Yeah technically it's still testable with the replace hack...

What's the rationale of having projectIds?

datastore/quickstart/quickstart.php
bshaffer: you're right. What do you think we should do for this test?

logging/quickstart/quickstart.php
bshaffer: My understanding was Logging was the one API we wanted to pass Project ID in for.

logging/quickstart/quickstart.php
bshaffer: I am assuming there's at least something in the logs for their project, since this requests the logs without any filter. What do you suggest instead?

datastore/quickstart/test/quickstartTest.php
bshaffer: We need to at least replace `__DIR__` for all of these tests

jmdobry:
> What's the rationale of having projectIds?

@omaray?

bshaffer:
My understanding is we would show both with and without the project ID being set, e.g. in the docs:

```
To instantiate the [PRODUCT] client:

<code>

If you need to explicitly set the project ID, you can pass it in like this:

<code>
```

If we do it this way (which I prefer), we can create another quickstart sample for each one `create_client_with_project_id.php` or something.

datastore/quickstart/test/quickstartTest.php
bshaffer: `__DIR__` needs replaced for the autoloading to work

jmdobry:
We, will just not on the same page. The quickstart samples, shown on Client Libraries pages, will use the explicit project ID. Everywhere else in the docs (the vast majority of samples) will use an implicit project ID.

logging/quickstart/quickstart.php
jmdobry: Can we make this sample match the other logging quickstart samples (here's [the python one](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/quickstarts/logging/cloud-client/quickstart.py)), i.e. log an entry as opposed to listing entries.

bshaffer: SGTM

bshaffer:
Hmm, seems strange to recommend doing it two different ways. The docs can cover explicitly setting the project ID, but they should explain when this would be done (when application default credentials span multiple projects) and why.

bshaffer:
As for using `$bigqueryClient` vs `$bigquery`, this is something we also want to be consistent with our other samples. I can't speak to the other languages, but I think it's more important to be consistent within the language docs than across languages.

datastore/quickstart/test/quickstartTest.php
tmatsuo: If the `__DIR__` is the only thing you need to replace, you don't need to copy the file.

logging/quickstart/quickstart.php
tmatsuo: For the CLI, we still need an explicit projectId, but for creating the LoggingClient and do some simple stuff, it can be omitted.

datastore/quickstart/quickstart.php
tmatsuo: I'd create the key with named id and `upsert` the entity.
But I think we should match with the other language samples.

logging/quickstart/quickstart.php
bshaffer: Ok, added

bshaffer: This is now fixed - the logging quickstart writes a logging sample

datastore/quickstart/test/quickstartTest.php
bshaffer: I disagree, I do not want to modify the file in place when we run the tests

bshaffer: ahh, good point

logging/quickstart/quickstart.php
bshaffer: Why do we need to pass in a Project ID for the CLI?

datastore/quickstart/quickstart.php
bshaffer: The other language samples do what this sample does. But I agree this is not a good sample. @jmdobry thoughts on changing this to an `upsert`?

datastore/quickstart/quickstart.php
bshaffer: If we call "lookup" instead of "upsert" here, and maybe use a name instead of ID, would that be more appropriate?

```php
$kind = 'Person';
$name = 'Bob';
$key = $datastore->key($kind, $name);
$entity = $datastore->lookup($key);
```

pubsub/quickstart/test/quickstartTest.php
bshaffer: This one needs to copy the file so we can verify the new topic is created!

speech/quickstart/quickstart.php
```diff
@@ -0,0 +1,32 @@
...
+# Your Google Cloud Platform project ID
+$projectId = 'YOUR_PROJECT_ID';
+
+# Instantiates a client
+$speech = new SpeechClient([
+    'projectId' => $projectId
+]);
+
+# The name of the audio file to transcribe
+$fileName = '/resources/audio.raw';
```
bshaffer: We should add the `__DIR__` here and not below.

speech/quickstart/quickstart.php
```diff
@@ -0,0 +1,32 @@
...
+# The audio file's encoding and sample rate
+$options = [
+    'encoding' => 'LINEAR16',
+    'sampleRate' => 16000,
+];
+
+# Detects speech in the audio file
+$results = $speech->recognize(fopen(__DIR__ . $fileName, 'r'), $options);
+
+echo 'Transcription: '.$results[0]['transcript'];
```
bshaffer: need spaces between concatinating `.`

speech/quickstart/quickstart.php
jmdobry: Done

vision/quickstart/quickstart.php
```diff
@@ -0,0 +1,34 @@
...
+# Your Google Cloud Platform project ID
+$projectId = 'YOUR_PROJECT_ID';
+
+# Instantiates a client
+$vision = new VisionClient([
+    'projectId' => $projectId
+]);
+
+# The name of the image file to annotate
+$fileName = '/resources/wakeupcat.jpg';
```
jmdobry: Done

speech/quickstart/quickstart.php
jmdobry: Done

datastore/quickstart/quickstart.php
```diff
@@ -10,18 +10,21 @@
...
-# Retrieves the task
-$task = $datastore->lookup($taskKey);
+# Prepares the new entity
+$task = $datastore->entity($taskKey, ['description' => 'Buy milk']);
 
-echo 'Fetched task: ' . $task->key();
+# Saves the entity
+$datastore->upsert($task);
+
+echo 'Saved ' . $task->key() . ': ' . $task['description'];
```
bshaffer: This output will look like this:

```
Saved [ [Task: simpletask1 ] ]: Buy Milk
```

Seems ok to me :D

jmdobry: I couldn't for the life of me get it to just print `simpletask1` (that's what I did in other languages).

datastore/quickstart/quickstart.php
```diff
@@ -10,18 +10,21 @@
 # Instantiates a client
 $datastore = new DatastoreClient();
 
-# The kind of the entity to retrieve
+# The kind for the new entity
 $kind = 'Task';
 
-# The name/ID of the entity to retrieve
+# The name/ID for the new entity
 $name = 'sampletask1';
```
jmdobry: Nah, what would be the point of the description field in that case?

bshaffer:
I believe this guy is ready for a final look!

jmdobry:
LGTM except for one thing. The PHP Cloud Client library does not yet support auto-detecting the project ID, meaning that the user must set the `GCLOUD_PROJECT` environment variable for these samples to work. The docs will expect that all the user needs to do is `gcloud beta auth application-default login`. Is much work to make them all use an explicit project ID until the library supports auto-detecting the project ID?

jmdobry:
Are `composer.lock` supposed to be committed to version control?

Also, I'm now seeing a number of files in this PR with changes (or being renamed or something) that are unrelated to the quickstart samples, was that intentional?

bshaffer:
@jmdobry 

`composer.lock` is committed to projects but not to dependencies. So this library should contain `composer.lock` (so the dependencies don't change w/o an explicit update) whereas a dependent library like `google-cloud-php` should never contain `composer.lock`

The renames had to happen for the organization of having `quickstart` not clutter up another sample application. I.e. `logging/*` became `logging/api/*` so that the logging sample code is clearly separated.

bshaffer:
@tmatsuo PTAL

jmdobry:
Can we please keep the branch around until I can update the docs? We don't want them to 404.

bshaffer:
@jmdobry should just take a second to update these. I have the change in flight.

PR #180, opened 2016-10-06
adds Jenkins Build Status badge

bshaffer:


tmatsuo:
Do we need a test utility for eventual consistent stuff?
Something like this:
https://github.com/GoogleCloudPlatform/python-repo-tools/blob/master/gcp/testing/eventually_consistent.py

bshaffer:
@tmatsuo I think that's a good idea

tmatsuo:
@bshaffer 
Here is the prototype:
https://github.com/GoogleCloudPlatform/php-docs-samples/pull/181/files#diff-a05c516ac20f40279c1f18e7c3ea5166R457

bshaffer:
I added https://github.com/GoogleCloudPlatform/php-tools/pull/15 for this, WDYT?

PR #181, opened 2016-10-06
Adding snippets for Datastore queries

tmatsuo:


tmatsuo:
The datastore index!

I created the datastore index manually.
```bash
$ gcloud preview datastore --project cloud-samples-tests-php create-indexes index.yaml
```

datastore/api/src/functions/concepts.php
```diff
@@ -295,3 +296,226 @@ function create_entity_with_array_property(DatastoreClient $datastore, Key $key)
...
+ * Run a given query.
+ *
+ * @param DatastoreClient $datastore
+ * @return \Generator <\Google\Cloud\Datastore\Entity>
+ */
+function run_query(DatastoreClient $datastore, Query $query)
+{
+    // [START run_query]
+    $result = $datastore->runQuery($query);
+    // Result is a generator of \Google\Cloud\Datastore\Entity
```
bshaffer: I don't really feel this line is necessary, but if you feel strongly about it, it should be above the line of code and not below

bshaffer: Not necessary because the `@return` states it's a generator

datastore/api/test/ConceptsTest.php
```diff
@@ -297,7 +298,182 @@ public function testCreateEntityWithArrayProperty()
...
+        // TODO: Generalize and move it to php-tools.
+        $i = 0;
+        $retry = 3;
+        $originalException = null;
+        while ($i <= $retry) {
+            $i += 1;
+            try {
+                $func();
+                return;
+            } catch (\Exception $e) {
```
bshaffer: We should make this `PHPUnit_Framework_ExpectationFailedException`

datastore/api/src/functions/concepts.php
tmatsuo: Hm, when you look at the source code on github, yes, it's obvious. But if you look at our docs, it's not.

But I will actually remove the comment. It's easy to guess the result is a list (or generator) and you can loop it with `foreach`.

datastore/api/test/ConceptsTest.php
tmatsuo: I can do that in this repo, but when we generalize it, I don't think it will work with phpunit 5.x

datastore/api/test/ConceptsTest.php
```diff
@@ -297,7 +298,182 @@ public function testCreateEntityWithArrayProperty()
...
+            $result = run_projection_query(self::$datastore, $query);
+            $this->assertEquals(2, count($result));
+            $this->assertEquals([4], $result[0]);
+            $this->assertEquals([50], $result[1]);
+        });
+    }
+
+    public function runEventualConsistentTest(Callable $func)
+    {
+        // TODO: Generalize and move it to php-tools.
```
bshaffer: See https://github.com/GoogleCloudPlatform/php-tools/pull/15

tmatsuo:
The tests should fail now, but they should pass once the following PRs are submitted.
https://github.com/GoogleCloudPlatform/google-cloud-php/pull/194
https://github.com/GoogleCloudPlatform/google-cloud-php/pull/200

tmatsuo:
Now the tests are passing with `dev-master` branch of the client library :)
I have squashed, rebased, and pushed. Let's see the tests are now stable with our eventual consistent test runner.

@bshaffer PTAL

tmatsuo:
Oh logging test is failing
```
1) Google\Cloud\Samples\Logging\Tests\ListEntriesCommandTest::testListEntries
```

I will use the eventual consistent test runner for them too in a different PR.

datastore/api/composer.json
```diff
@@ -1,9 +1,10 @@
 {
     "require": {
-        "google/cloud": "dev-master"
+        "google/cloud": "<2.0"
```
bshaffer: Would prefer semantic versioning here, i.e. `^1.0`

tmatsuo: Unfortunately `~1.0` and `^1.0` don't work because the latest version of `google/cloud` is still `0.10.2`.

tmatsuo: Is it more intuitive if I do `"~1.0|~0.10"` or something like that?

tmatsuo:
@bshaffer PTAL
I changed it to forms like `"~1.0|~0.10"`

datastore/api/src/functions/concepts.php
```diff
@@ -60,9 +61,9 @@ function create_entity(DatastoreClient $datastore)
  * @param DatastoreClient $datastore
  * @return \Google\Cloud\Datastore\Entity
  */
-function upsert_entity(DatastoreClient $datastore)
+function upsert(DatastoreClient $datastore)
```
bshaffer: I prefer `upsert_entity` to upsert, and same for below functions. What is the reason for changing it?

datastore/api/src/functions/concepts.php
```diff
@@ -256,7 +257,7 @@ function create_key_with_multi_level_parent(DatastoreClient $datastore)
  * @param Key $key
  * @return \Google\Cloud\Datastore\Entity
  */
-function create_entity_with_option(DatastoreClient $datastore, Key $key)
+function properties(DatastoreClient $datastore, Key $key)
```
bshaffer: Again, I think these functions (especially this one and `array_value` below) are being named too generically. I prefer the older names.

datastore/api/src/functions/concepts.php
```diff
@@ -281,16 +283,443 @@ function create_entity_with_option(DatastoreClient $datastore, Key $key)
...
+}
+
+/**
+ * Create a kindless query.
+ *
+ * @param DatastoreClient $datastore
+ * @param Key $lastSeenKey
+ * @return Query
+ */
+function kindless_query(DatastoreClient $datastore, Key $lastSeenKey)
```
bshaffer: kindless query? should this be keyless query?

bshaffer: As in `Kind`. Got it 👍

datastore/api/src/functions/concepts.php
tmatsuo: Specifically it matches to the region tags. This pattern matches with nodejs samples.

datastore/api/src/functions/concepts.php
tmatsuo: It also matches to other languages, specifically nodejs

datastore/api/src/functions/concepts.php
```diff
@@ -281,16 +283,443 @@ function create_entity_with_option(DatastoreClient $datastore, Key $key)
...
+        ->order('priority', Query::ORDER_DESCENDING);
+    // [END basic_query]
+    return $query;
+}
+
+/**
+ * Run a given query.
+ *
+ * @param DatastoreClient $datastore
+ * @return \Generator <\Google\Cloud\Datastore\Entity>
```
bshaffer: We could add `use Generator` above so that the forward slash isn't necessary. Same with `use Google`.

datastore/api/src/functions/concepts.php
```diff
@@ -21,6 +21,7 @@
 use Google\Cloud\Datastore\DatastoreClient;
 // [END datastore_use ]
```
bshaffer: nit extra whitespace

datastore/api/src/functions/concepts.php
tmatsuo: Done

PR #186, opened 2016-10-11
Use EventuallyConsistentTestTrait for a test

tmatsuo:


logging/api/composer.json
```diff
@@ -4,7 +4,8 @@
         "symfony/console": "^3.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.8"
+        "phpunit/phpunit": "~4.8",
+        "google/cloud-tools": "<2.0"
```
bshaffer: This is an odd way to declare this. We should start using semantic versioning and use `^1.0`

logging/api/test/ListEntriesCommandTest.php
```diff
@@ -59,18 +62,22 @@ public function setUp()
...
-        $this->expectOutputRegex('/: Test Message/');
+        $this->runEventuallyConsistentTest(function() use ($commandTester) {
+            $commandTester->execute(
+                [
+                    '--project' => $this->projectId,
+                    '--logger' => 'my_test_logger'
+                ],
+                ['interactive' => false]
+            );
+            $this->expectOutputRegex('/: Test Message/');
```
bshaffer: This won't work as designed, as `expectOutputRegex` is only assessed at the end of the test's execution.  Instead you'll have to capture the output:

```php
ob_start();
$commandTester->execute(
    [
        '--project' => $this->projectId,
        '--logger' => 'my_test_logger'
    ],
    ['interactive' => false]
);
$output = ob_get_clean();
$this->assertContains(':Test Message', $output);
```

tmatsuo: Good catch! I'll fix this.

logging/api/composer.json
bshaffer: I realize this won't work, since our highest version is `0.5`. I think I see what you're going for, you want to basically say we'll maintain BC up to `2.0`. I wonder if there's a better way to do this.

tmatsuo: It might seem odd, but it's sane to me. Here is the reason.

Now `google/cloud-tools` is around `0.5`. We can not use `^1.0` because `0.5` doesn't match it. Sometimes in the future, I may release the official version `1.0` of that tool. After that, it should follow semantic versioning. I don't want this test to break when I release `google/cloud-tools` `2.0`. So I'm using `<2.0`.

logging/api/test/ListEntriesCommandTest.php
tmatsuo: Done

tmatsuo:
Use `ob_start` and `ob_get_clean`. Also rebased to the master. @bshaffer PTAL

PR #188, opened 2016-10-11
Adding files for moved Logging functions

bshaffer:
Will "fix" 404's for some [Logging Docs Samples](https://cloud.google.com/logging/docs/api/tasks/authorization)


tmatsuo:
What does `sylink` stand for?

tmatsuo:
First I thought it was a typo for symlink but looking at the change, they are not symlinks, so I'm asking.

bshaffer:
Oh, that's the commit name apparently. That was what I tried first but didn't work. Renaming!

PR #189, opened 2016-10-11
Echo the current directory before phpunit

tmatsuo:
to get a better glance on the travis console UI


bshaffer:
Hmm, you can just scroll above the `composer install` to see the `pushd` directory. But LGTM

PR #190, opened 2016-10-12
Uses "runtime:php" instead of "runtime:custom" for Symfony sample

bshaffer:


appengine/flexible/symfony/tests/DeployTest.php
```diff
@@ -131,7 +132,7 @@ private static function createSymfonyProject($targetDir)
         file_put_contents($installFile, Yaml::dump($config));
 
         // move the code for the sample to the new drupal installation
-        $files = ['app.yaml', 'Dockerfile', 'nginx-app.conf'];
+        $files = ['app.yaml', 'nginx-app.conf'];
```
tmatsuo: re: `nginx-app.conf`

This file just specifies the root controller file name. I'm ok for now, but is it possible to not have this file for symfony apps?

Is it a good idea to introduce another envvar like `FRONT_CONTROLLER_FILE` or something, so that you can do:

app.yaml
```yaml
runtime: php
vm: true

runtime_config:
  document_root: web

env_variables:
  WHITELIST_FUNCTIONS: libxml_disable_entity_loader
  FRONT_CONTROLLER_FILE: app.php
```

It will introduce another small complexity into the docker image, maybe I can use `envsubst` for this purpose.

bshaffer: I think that'd be a great idea. I would love to eliminate `nginx-app.conf` for basic configurations!

appengine/flexible/symfony/add_composer_scripts.php
```diff
@@ -0,0 +1,58 @@
...
+        $newScripts[$name] = [];
+    }
+    $newScripts[$name] = array_unique(array_merge($newScripts[$name], $script));
+}
+if ($json['scripts'] == $newScripts) {
+    exit("composer.json is already configured\n");
+}
+
+$json['scripts'] = $newScripts;
+file_put_contents($composerFile, json_encode($json, JSON_PRETTY_PRINT));
```
bshaffer: Composer does this same thing when you run `composer require`, so I'm not too worried about it.

appengine/flexible/symfony/add_composer_scripts.php
```diff
@@ -0,0 +1,58 @@
...
+
+if (!$json = json_decode($contents, true)) {
+    exit("Could not parse composer.json\n");
+}
+
+if (!isset($json['scripts'])) {
+    $json['scripts'] = [];
+}
+
+$scriptsContents = file_get_contents(__DIR__ . '/composer_scripts.json');
```
bshaffer: I copied this over from the `drupal8` example, which has 6+ scripts that may change. I wanted to be consistent.

bshaffer: I agree this is overkill for symfony. I removed them in favor of instructions in the `README`.

appengine/flexible/symfony/tests/DeployTest.php
bshaffer: tracked in https://github.com/GoogleCloudPlatform/php-docker/issues/151

PR #192, opened 2016-10-12
Merges endpoints.html.twig with the README for the Endpoints sample.

wmwong:
Removes endpoints.html.twig and the paths to / and /api-docs.
Adds sections for the Endpoints graphs and the Swagger UI in the README.
Fixes deploy command in the README.
Updates the dependencies.


tmatsuo:
What's the rationale of removing the top page, and the `/api-docs`?
I think it's nice if you see something useful when you visit the root of your app.

wmwong:
Endpoints will only allow requests to paths that are specified in swagger.yaml. swagger.yaml is used to define the API surface. / and /api-docs are not conceptually part of the API surface and are not included in swagger.yaml so Endpoints will reject requests to these paths. Therefore, after deploying to AppEngine, these paths are no longer reachable. This is why the information in / has been moved to the README. /api-docs is basically swagger.yaml which the user already has access to.

tmatsuo:
I see, thanks for the PR!

tmatsuo:
We should have an e2e test for making sure it's really running endoint in cloud, filed https://github.com/GoogleCloudPlatform/php-docs-samples/issues/193

PR #194, opened 2016-10-13
Adding snippets for Indexes and Transactions.

tmatsuo:


datastore/api/src/functions/concepts.php
```diff
@@ -725,3 +726,148 @@ function inequality_sort_invalid_not_first(DatastoreClient $datastore)
...
+            'tags' => ['fun', 'programming', 'learn'],
+            'collaborators' => ['alice', 'bob', 'charlie'],
+            'created' => new \DateTime(),
+        ]
+    );
+    // [END exploding_properties]
+    return $task;
+}
+
+// [START transactional_update]
```
bshaffer: is this intentional to have the whole function for this snippet?

datastore/api/src/functions/concepts.php
```diff
@@ -725,3 +726,148 @@ function inequality_sort_invalid_not_first(DatastoreClient $datastore)
...
+            // if $i >= $retries, the failure is final
+            continue;
+        }
+        // Succeeded!
+        break;
+    }
+    // [END transactional_retry]
+}
+
+// [START transactional_get_or_create]
```
bshaffer: same here, I feel we should be consistent with having either functions or snippets.

datastore/api/src/functions/concepts.php
tmatsuo: This makes sense in the context of the documentation. In the later section, we will explain how to use this function `transfer_funds`, specifically in the region `transactional_retry`.

datastore/api/src/functions/concepts.php
tmatsuo: Yeah for this snippet, probably we can just show the code, instead of the whole function.

tmatsuo:
@bshaffer Thanks! PTAL

bshaffer:
Awesome, works for me! Well done.

PR #198, opened 2016-10-18
Adding snippets for consistency and metadata queries.

tmatsuo:


datastore/api/src/functions/concepts.php
```diff
@@ -860,3 +860,169 @@ function get_task_list_entities(DatastoreClient $datastore)
...
+ * Create and run a kind query.
+ *
+ * @param DatastoreClient $datastore
+ * @return array<string> kinds returned from the query
+ */
+function kind_run_query(DatastoreClient $datastore)
+{
+    // [START kind_run_query]
+    $query = $datastore->query()
+        ->kind('__kind__')
```
bshaffer: I know we're being consistent here, but `kind_run_query` tells me nothing about this. Can we change the comment above to "Creates a query to list all kinds in datastore"

datastore/api/src/functions/concepts.php
```diff
@@ -860,3 +860,169 @@ function get_task_list_entities(DatastoreClient $datastore)
...
+    $query = $datastore->query()
+        ->kind('__property__')
+        ->projection(['__key__']);
+    $result = $datastore->runQuery($query);
+    /* @var array<string => array<string>> $properties */
+    $properties = [];
+    /* @var Entity $entity */
+    foreach ($result as $entity) {
+        $kind = $entity->key()->path()[0]['name'];
+        $propertyName = $entity->key()->path()[1]['name'];
```
bshaffer: Is there not a way to get this information without array indexes? The reason we have veneers is to handle this type of thing in a cleaner way. Why not `$entity->getKind` and `$entity->getProperties` ?

datastore/api/src/functions/concepts.php
```diff
@@ -860,3 +860,169 @@ function get_task_list_entities(DatastoreClient $datastore)
...
+
+/**
+ * Create and run a property query with a kind.
+ *
+ * @param DatastoreClient $datastore
+ * @return array<string => string>
+ */
+function property_by_kind_run_query(DatastoreClient $datastore)
+{
+    // [START property_run_query]
```
bshaffer: this should be `property_by_kind_run_query`

datastore/api/src/functions/concepts.php
tmatsuo: Done

tmatsuo:
@bshaffer Thanks, PTAL

datastore/api/src/functions/concepts.php
bshaffer: Yeah, `getProperties` doesn't really make sense. I was more thinking of a way to get the names of all the fields returned with an entity. But I haven't really thought it through so please disregard

datastore/api/src/functions/concepts.php
```diff
@@ -860,3 +860,172 @@ function get_task_list_entities(DatastoreClient $datastore)
...
+    $kinds = [];
+    foreach ($result as $kind) {
+        $kinds[] = $kind->key()->pathEnd()['name'];
+    }
+    // [END kind_run_query]
+    return $kinds;
+}
+
+/**
+ * Create and run a property query.
```
bshaffer: Would love to see a better description here also.  Like "Create and run a query to list the properties of all kinds in Datastore."

datastore/api/src/functions/concepts.php
```diff
@@ -860,3 +860,172 @@ function get_task_list_entities(DatastoreClient $datastore)
...
+ * Create and run a property query with property filtering.
+ *
+ * @param DatastoreClient $datastore
+ * @return array
+ */
+function property_filtering_run_query(DatastoreClient $datastore)
+{
+    // [START property_filtering_run_query]
+    $ancestorKey = $datastore->key('__kind__', 'Task');
+    $startKey = $datastore->key('__property__', 'priority')
```
bshaffer: With these it seems like we should have helper methods to avoid the ugly `__xx__` syntax. Such as `$datastore->kindKey()`, `$datastore->propertyKey()`, and `$query->filterKey()`. Also, can't `$query->kind('__property__')` be inferred by the startKey?

Changes to the veneer like this should obviously require a lot of consideration, so this is just brainstorming.

cc @jdpedrie

datastore/api/src/functions/concepts.php
tmatsuo: Ok, a bug filed (I want to merge this earlier than one more iteration of useless test run):
https://github.com/GoogleCloudPlatform/php-docs-samples/issues/200

tmatsuo:
Thanks!

PR #199, opened 2016-10-18
Adding a Cloud Datastore tutorial application.

tmatsuo:


tmatsuo:
Sorry, I always run the php-cs-fixer before creating a PR, but this time I misread the result.

datastore/tutorial/src/functions.php
```diff
@@ -0,0 +1,108 @@
...
+
+// [START add_entity]
+/**
+ * Create a new task with a given description.
+ *
+ * @param DatastoreClient $datastore
+ * @param $description
+ * @return \Google\Cloud\Datastore\Entity
+ */
+function add_task(DatastoreClient $datastore, $description)
```
bshaffer: shouldn't the function names match the region tags?

bshaffer: I actually like how these are explicitly mentioning the `tasks` are the kinds getting included. It makes it more explicit that these are example functions.

tmatsuo: I don't know why, but nodejs has the different names

datastore/tutorial/src/CreateTaskCommand.php
```diff
@@ -0,0 +1,54 @@
...
+                'The description of the new task'
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $datastore = build_datastore_service();
+        $description = $input->getArgument('description');
+        $task = add_task($datastore, $description);
+        printf(
```
bshaffer: We should use the `OutputInterface` object here. This will improve tests (we can perform more explicit tests with the `CommandTester` object)

datastore/tutorial/src/functions.php
```diff
@@ -0,0 +1,108 @@
...
+// [END add_entity]
+
+// [START update_entity]
+/**
+ * Mark a task with a given id as done.
+ *
+ * @param DatastoreClient $datastore
+ * @param int $taskId
+ */
+function mark_done(DatastoreClient $datastore, $taskId)
```
bshaffer: Can't we just manually create `DataStoreClient` every time? I feel like this would make for a better sample.

tmatsuo: I know what you're talking about, but I have a slightly different opinion, especially with the Datastore client. Receiving the DatastoreClient is a great pattern because it makes very easy to test it properly.

In our test, it creates the special datastoreClient with the namespace and call these functions, so that you can isolate multiple test runs.

tmatsuo:
@bshaffer Updated for the code review. I also added system tests for the commands. PTAL

PR #201, opened 2016-10-19
Updated region tags for docs

WalterHub:
To enable doc build


tmatsuo:
I'm just curious. Why the previous tags didn't work?

WalterHub:
For each snippets section in the docs, the topics use only a single tag, regardless of programming language. If a specific region tag isn't found in a new set of snippets, the doc won't build.

tmatsuo:
The test is failing, but I'm pretty sure it's because the test can not handle the PR from other repos. Merging and will fix it later.

bshaffer:
@WalterHub is this a new requirement? I have not heard of this before.

WalterHub:
It's been used by Datastore since late 2015.

PR #202, opened 2016-10-19
Added a comment for a memory efficient loop.

tmatsuo:


tmatsuo:
Unfortunately, now some tests in this repo is broken, but the tests and the cs-check are passing locally for `datastore/api` directory. I'm going to merge without running the test on travis (job cancelled).

WalterHub:
LGTM

PR #203, opened 2016-10-19
Fixed broken tests for the language API

tmatsuo:


tmatsuo:
@bshaffer The tests should pass

tmatsuo:
The test and cs-fixer runs locally, I'm going to merge this

PR #204, opened 2016-10-19
Removed limit from keys_only_query

WalterHub:
limit() not needed for this snippet


tmatsuo:
Thanks, I have also some other cleanups, and this change seems like test change also. I will incorporate this change into my PR. SG?

WalterHub:
Yes, sg.

PR #208, opened 2016-10-21
Fixed a broken test for Language API quickstart

tmatsuo:


tmatsuo:
Also PHP5.5 seems disappeared on travis.

PR #212, opened 2016-10-25
updates datastore sample to env:flex

bshaffer:


appengine/flexible/datastore/tests/LocalTest.php
```diff
@@ -29,7 +29,7 @@ public function setUp()
     public function createApplication()
     {
         $app = require __DIR__ . '/../app.php';
-        $app['google.dataset_id'] = getenv('GOOGLE_PROJECT_ID');
+        $app['google.dataset_id'] = getenv('GCLOUD_PROJECT');
```
tmatsuo: Did you set the envvar `GCLOUD_PROJECT`? The test seems failing. Also why the change?

bshaffer: This environment variable is added to every container in the `env:flex` environment, so we won't need to add it anywhere.

PR #214, opened 2016-10-25
Updates Mailjet sample to env:flex

bshaffer:


bshaffer:
closing so we can submit to upgrade-to-env-flex branch instead

PR #217, opened 2016-10-25
updates drupal sample for env:flex

bshaffer:


appengine/flexible/drupal8/README.md
```diff
@@ -1,13 +1,13 @@
-Drupal 8 on Managed VMs
-=======================
+Drupal 8 on App Engine Flexible
```
tmatsuo: I think it's App Engine Flexible Environment

appengine/flexible/drupal8/composer.json
```diff
@@ -4,8 +4,10 @@
     },
     "require-dev": {
         "guzzlehttp/guzzle": "^6.2",
+        "google/cloud-tools": "<1.0",
```
tmatsuo: I think we discuss it recently and concluded that we should use "~1.0|~0.5" or something.

appengine/flexible/drupal8/nginx-app.conf
```diff
@@ -1,4 +0,0 @@
-location / {
```
tmatsuo: Yay!

PR #219, opened 2016-10-25
upgrades Mailjet sample for env:flex

bshaffer:


appengine/flexible/mailjet/composer.json
```diff
@@ -0,0 +1,12 @@
+{
+    "require": {
+        "silex/silex": "^1.3",
+        "mailjet/mailjet-apiv3-php": "^1.1",
+        "guzzlehttp/guzzle": "~6.1.0"
+    },
+    "require-dev": {
+        "symfony/browser-kit": "^3.0",
+        "paragonie/random_compat": "^2.0",
+        "google/cloud-tools": "<=1.0"
```
tmatsuo: '~1.0|~0.5' ?

bshaffer: Same as the other, I'll update to `^0.5` unless you have a problem with that.

PR #222, opened 2016-10-25
update analytics sample for env:flex

bshaffer:


bshaffer:
Merging - This received approval in #210

PR #227, opened 2016-10-27
upgrades symfony sample to env:flex

bshaffer:


appengine/flexible/symfony/composer.json
```diff
@@ -4,6 +4,8 @@
         "monolog/monolog": "^1.19",
         "symfony/console": " ~2.6.0",
         "symfony/process": "^3.0",
-        "symfony/yaml": "^3.0"
+        "symfony/yaml": "^3.0",
+        "google/cloud-tools": "<1.0",
```
tmatsuo: '~1.0|~0.5'

bshaffer: Not a fan of this, since `~1.0` might break BC. We should stick with `^0.5`

PR #229, opened 2016-10-27
fixes appengine standard cloudsql test skip statment

bshaffer:


bshaffer:
Merging so we can run jenkins tests

appengine/standard/cloudsql/test/cloudsqlTest.php
```diff
@@ -30,7 +30,7 @@ public function createApplication()
         $dsn = getenv('MYSQL_DSN');
         $username = getenv('MYSQL_USERNAME');
         $password = getenv('MYSQL_PASSWORD');
-        if (!isset($dsn, $username) || false === $password) {
+        if (empty($dsn) || empty($username) || false === $password) {
```
tmatsuo: It's trivial, but why it only compares `$password` to false? Can it also be `empty($password)`?

PR #232, opened 2016-11-01
adds create_table, create_dataset, and copy_table to bigquery samples

bshaffer:


bigquery/api/src/CopyTableCommand.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        if (!$projectId = $input->getOption('project')) {
+            $projectId = $this->getProjectIdFromGcloud();
+        }
+        $datasetId = $input->getArgument('dataset');
+        $sourceTableId = $input->getArgument('source-table');
+        $builder = new ServiceBuilder([
```
tmatsuo: Do you want to keep using the `ServiceBuilder`?

bshaffer: no, I'll update all the samples :D

bshaffer: On second thought, this should be done in another PR, because if we change it for this sample we should change it for all of them, to be consistent.

bshaffer: see https://github.com/GoogleCloudPlatform/php-docs-samples/pull/237

PR #233, opened 2016-11-01
Translate samples.

SurferJeffAtGoogle:


translate/api/src/DetectLanguageCommand.php
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Command line utility to detect which language some text is written in.
+ */
+class DetectLanguageCommand extends Command
+{
+    public function __construct($apiKey)
+    {
+        parent::__construct();
+        $this->apiKey = $apiKey;
```
bshaffer: We have been passing this in to the command itself, i.e. with an `--api_key` option (see [here](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/efd78b3d00e7ff1f1935634763f909146b2ea563/appengine/flexible/endpoints/EndpointsCommand.php#L43)). I prefer this, because it does not require the user to edit a file in order to run the command.

translate/api/src/DetectLanguageCommand.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+    protected function configure()
+    {
+        $this
+            ->setName('detect')
+            ->setDescription('Detect which language text was written in using '
+                . 'Google Cloud Translate API')
+            ->setHelp(<<<'EOF'
```
bshaffer: I have never seen single quotes around `EOF`. It's a nit, but it can be removed.

bshaffer: I have never seen single quotes around `EOF`. It's a nit, but they can be removed.

translate/api/src/DetectLanguageCommand.php
```diff
@@ -0,0 +1,74 @@
...
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->detectLanguage($input->getArgument('text'), $output);
+    }
+
+    // [START translate_detect_language]
+    protected function detectLanguage($text, OutputInterface $output)
```
bshaffer: I am not a huge fan of this code snippet, as it will not run when copied/pasted. It might be better to move the whole snippet into it's own include file, like this:

```php
<?php

// [START translate_detect_language]
use Google\Cloud\Translate\TranslateClient;

// $apiKey = 'YOUR API KEY';
// $text = 'your text to translate';
$translate = new TranslateClient([
    'key' => $apiKey,
]);
$result = $translate->detectLanguage($text);
print("Language code: $result[languageCode]\n");
print("Confidence: $result[confidence]\n");
// [END translate_detect_language]
```

and then in the command: 

```php
protected function detectLanguage($text, $apiKey)
{
    require_once(__DIR__ . '/snippets/detect_language.php');
}
```

translate/api/src/DetectLanguageCommand.php
SurferJeffAtGoogle: Done.

translate/api/src/DetectLanguageCommand.php
SurferJeffAtGoogle: I changed the code to use print instead out output->writeln().  As a result, the commands will no longer obey the --quiet flag, even though they claim to:
```
rennie@rennie:~/gitrepos/php-docs-samples/translate/api$ php translate.php help detect
Usage:
  detect [options] [--] <text>

Arguments:
  text                   The text to examine.

Options:
  -k, --api-key=API-KEY  Your API key.
  -h, --help             Display this help message
  -q, --quiet            Do not output any message
  -V, --version          Display this application version
      --ansi             Force ANSI output
      --no-ansi          Disable ANSI output
  -n, --no-interaction   Do not ask any interactive question
  -v|vv|vvv, --verbose   Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

Help:
  The detect command detects which language text was written in using the Google Cloud Translate API.
  
      php translate.php detect -k YOUR-API-KEY "Your text here"
  
rennie@rennie:~/gitrepos/php-docs-samples/translate/api$ php translate.php detect -k XXX --quiet "Hello World."
Language code: en
Confidence: 0.7100697
```

What are the advantages of moving the code into a separate snippet?

SurferJeffAtGoogle:
Tests don't run https://coveralls.io/builds/8633651 because tests expect a GOOGLE_API_KEY.

TranslateClient requires an API key:
```
 public function __construct(array $config = [])
    {
        if (!isset($config['key'])) {
            throw new \InvalidArgumentException('A key is required.');
        }
```

So, I'll add one to the travis environment.

translate/api/src/DetectLanguageCommand.php
bshaffer: The main advantage here is the snippet can be copied and pasted. Having the `protected` designation means for the code to run, the user would have to wrap it in a class or know to leave that out. Having a separate snippet means the code can easily be copied/pasted to run.  In lieu of separate snippets consider doing this:

```php
/***
 * @param $apiKey string Your API key.
 * @param $text string The text for which to detect the language.
 */
protected function detectLanguage($apiKey, $text)
{
    // [START translate_detect_language]
    // $apiKey = 'YOUR API KEY';
    // $text = 'your text to translate';
    $translate = new TranslateClient([
        'key' => $apiKey,
    ]);
    $result = $translate->detectLanguage($text);
    print("Language code: $result[languageCode]\n");
    print("Confidence: $result[confidence]\n");
    // [END translate_detect_language]
}
```

Then, as long as it's clear to include the `use` statement in the documentation somewhere, this will run as expected.

In general, the greater context of the sample should not be required, since these will be displayed without that context.

SurferJeffAtGoogle:
Ready for another look.

Issue #193, opened 2016-10-12
Add e2e test for endpoint sample


tmatsuo:

tmatsuo:
Some tricky parts;

- you need to use gcloud beta app deploy, maybe we should tweak our deploytrait to make the command configurable
- you need to create an API key like curl -vv -d '{"message":"foo"}' -H 'Content-Type: application/json' "${ENDPOINTS_HOST}/echo?key=${ENDPOINTS_KEY}"
- ideally you should test it without the API key and check if it's correctly 404

bshaffer:
This was added [here](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/endpoints/getting-started/test/DeployTest.php), but we now need e2e for `GCE` and `GKE` as well.

jeffmendoza:
Awesome, I'm working on a doc for GKE testing.

meredithslota:
Closing due to capacity — will revisit if needed later on.

PR #236, opened 2016-11-04
Copy sample to /endpoints and make it generic.

sepehre:
We plan on cleaning up our docs for the various platforms and share this
sample throughout the various quickstart guides.

It no longer makes sense to have this under appgengine/flexible. Instead
it will live under endpoints/getting-started. The appengine/flexible
copy will be deleted after the cloud site docs have been updated and
released.

This also introduces how to deploy this sample using
GAE/GKE/GCE/GCE+Docker.


endpoints/getting-started/Dockerfile.custom
```diff
@@ -0,0 +1,15 @@
...
+    apt-get install -y build-essential curl php5 php5-cli && \
+    apt-get clean && \
+    rm /var/lib/apt/lists/*_*
+# Install Composer
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin/ --filename=composer
+RUN chmod 777 /usr/local/bin/composer
+
+COPY * /app/
+RUN cd /app/ && /usr/local/bin/composer install
+
```
jeffmendoza: You need `WORKDIR /app` or else the php command does not run in the /app directory.

endpoints/getting-started/Dockerfile.custom
```diff
@@ -0,0 +1,15 @@
...
+ENTRYPOINT ["/usr/bin/php", "-S", "localhost:8081"]
```
jeffmendoza: Change to `CMD ["/usr/bin/php", "-S", "0.0.0.0:8081"]` Entrypoint overrides alternative docker commands. Using `localhost` will not let the container accept connections from any other IP.

endpoints/getting-started/Dockerfile.custom
sepehre: done

jeffmendoza:
Looks good

@bshaffer please merge when ready.

bshaffer:
I'm not a fan of the directory name `getting-started`. The documentation uses `quickstart`, and the our other samples use `app`. We should use one of these instead. I do not believe `getting-started` is used in the other languages either.

sepehre:
1. The name was chosen consistently with other language specific repos.
https://github.com/GoogleCloudPlatform/ruby-docs-samples/tree/master/endpoints/getting-started
https://github.com/GoogleCloudPlatform/golang-samples/tree/master/endpoints/getting-started
https://github.com/GoogleCloudPlatform/nodejs-docs-samples/tree/master/endpoints/getting-started

Other samples will be added under /endpoints/ in the future.

2. We are removing the old appengine/flexible samples until the
documentation is updated. A second round of PRs will be sent to remove the
old samples.

bshaffer:
We should include a `README.md` in `appengine/flexible/endpoints` to point to this directory and the URL for the tutorial.

sepehre:
appengine/flexible/endpoints already has a README.md with links to the
cloud site documentation.

And as I mentioned, appengine/flexible/endpoints directory will be deleted
shortly once our cloud site documentation has been updated.
This sample will be referenced from these quickstarts:
https://cloud.google.com/endpoints/docs/quickstarts not new tutorials.

jeffmendoza:
@bshaffer do you have any other concerns? @sepehre covered the reasoning behind the copy, update docs, then delete plan, and the naming for the directory.

bshaffer:
Yeah it seems good, I would like to see some E2E tests also, but we can add those later

jeffmendoza:
I added myself to #193

bshaffer:
I've added e2e tests and updated for `env:flex` in #243

PR #237, opened 2016-11-08
removes servicebuilder for most BigQuery samples

bshaffer:
Also fixes a bug with table deletion


tmatsuo:
LGTM, but the test is failing.

PR #243, opened 2016-11-14
Updates new endpoints sample for env flex

bshaffer:
see #236


endpoints/getting-started/Dockerfile.custom
```diff
@@ -0,0 +1,16 @@
...
+    rm /var/lib/apt/lists/*_*
+# Install Composer
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin/ --filename=composer
+RUN chmod 777 /usr/local/bin/composer
+
+COPY * /app/
+WORKDIR /app
+RUN cd /app/ && /usr/local/bin/composer install
+
+CMD ["/usr/bin/php", "-S", "0.0.0.0:8081"]
```
tmatsuo: This is using PHP's built-in web server.

http://php.net/manual/en/features.commandline.webserver.php says:

> Warning
This web server was designed to aid application development. It may also be useful for testing purposes or for application demonstrations that are run in controlled environments. It is not intended to be a full-featured web server. It should not be used on a public network.

I don't think it's appropriate for the sample. WDYT?

tmatsuo: Maybe you can use `gcr.io/google_appengine/php`?

bshaffer: We for sure should not use the PHP built-in web server. Good catch!

bshaffer: It's just a lightweight example to show off how endpoints serves to an app, so it might be ok, but invariably someone will use it in production if we don't have a better example.

tmatsuo: I still don't think we should merge this unless stop using the built-in web server

sepehre: 1. Depending on gcr.io/google_appengine/php is not appropriate here because this container is meant to be generic and used for GCE/GKE as well. 

2. AFAIK, the production setup for running a php web server is to install xampp, or nginx and configure it with fastcgi etc, which are beyond the scope of this sample Dockerfile.

If you'd like to add a comment to this effect in the Dockerfile, go for it. 
Otherwise, if there's another simpler way that I'm unaware of, please do teach me.

jeffmendoza: 1. gcr.io/google_appengine/php is a generic php runtime and not specific to App Engine. We use these for GKE samples often.

jeffmendoza: Will this one work? https://github.com/GoogleCloudPlatform/getting-started-php/blob/master/optional-container-engine/Dockerfile

tmatsuo: Yeah the namespace is misleading, but `gcr.io/google_appengine/php` will be supposed to serve as a generic base image for our various platform.

Does the image need to listen to the port 8081? Unfortunately that image only listens on 8080 which might be bad for this use case.

Another option is to use php-fpm.

tmatsuo: @jeffmendoza 

Yeah I think it can be one liner:
```
FROM gcr.io/google_appengine/php
```

since it's only a single file app, the default should just work (but it's on 8080).

jeffmendoza: I just built and tested this code w/ the 1-line Dockerfile, works well. @tmatsuo can you update this PR w/ that change? I don't have access. Also, you will need to swap all instances of 8080 and 8081 in container-engine.yaml

tmatsuo: @jeffmendoza Thanks for the update! @bshaffer is also working on other minor changes and he will take care of that too.

endpoints/getting-started/container-engine.yaml
```diff
@@ -0,0 +1,56 @@
...
+# limitations under the License.
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: esp-echo
+spec:
+  ports:
+  - port: 80
+    targetPort: 8080
```
jeffmendoza: should be 8081

endpoints/getting-started/container-engine.yaml
```diff
@@ -0,0 +1,56 @@
...
+    metadata:
+      labels:
+        app: esp-echo
+    spec:
+      containers:
+      # [START esp]
+      - name: esp
+        image: b.gcr.io/endpoints/endpoints-runtime:0.3
+        args: [
+          "-p", "8080",
```
jeffmendoza: 8081

endpoints/getting-started/container-engine.yaml
```diff
@@ -0,0 +1,56 @@
...
+          "-a", "127.0.0.1:8080",
+          "-s", "SERVICE_NAME",
+          "-v", "SERVICE_VERSION",
+        ]
+      # [END esp]
+        ports:
+          - containerPort: 8080
```
jeffmendoza: 8081

sepehre:
I think the point of having two different ports for ESP and the App container is to be able to easily distinguish and reason between them in the docs. So in container-engine.yaml I'd like to see two different ports, and not 8080 for both. The issue is not whether having the same port "works", the issue is teaching the user how the different containers will interact with the docs and these samples.

I'm also concerned about having different ports in this sample compared to the other languages. That will make writing the documentation for these samples trickier and more error prone.

I'd ask that we take a pragmatic approach and
1. submit this PR with the sub-optimal PHP built-in web server so that the documentation work / review can be unblocked.
2. follow up with a second PR to work on making the gcr.io/google_appengine/php runtime port generic and configurable in order to replace the existing PHP built-in web server with it.

jeffmendoza:
> I think the point of having two different ports for ESP and the App container is to be able to easily distinguish and reason between them in the docs. So in container-engine.yaml I'd like to see two different ports, and not 8080 for both. The issue is not whether having the same port "works", the issue is teaching the user how the different containers will interact with the docs and these samples.

Correct, this was just fixed. ESP will be on 8081 and the app on 8080. We had to do this for Java as well for similar reasons.

> I'm also concerned about having different ports in this sample compared to the other languages. That will make writing the documentation for these samples trickier and more error prone.

The docs will be the same as the service still listens on port 80, so the curl commands will be consistent. Also the gke config will be git-included from the appropriate repo, so when the user selects 'php' they will see this php version.

sepehre:
>> Correct, this was just fixed. ESP will be on 8081 and the app on 8080.
We had to do this for Java as well for similar reasons.

Will you be updating all the other languages also? (Python, Ruby, Node, Go)
If not, we should revert the Java port change and keep Java/PHP consistent
with the other languages.

>> The docs will be the same as the service still listens on port 80, so
the curl commands will be consistent. Also the gke config will be
git-included from the appropriate repo, so when the user selects 'php' they
will see this php version.

I know the external port 80/443 is consistent. But we will have docs (if we
don't already) that explain the interaction between the ESP and App
containers and the product architecture so those ports should ideally be
consistent in our docs as well.

sepehre:
Jeff, Brad Friedman, Brent and I discussed this over email, and decided to
update all samples to use port 8080 for the app container, and 8081 for the
ESP container.

Jeff will send the PRs for the Python,Node,Go,Ruby repos.

Thanks!

PR #246, opened 2016-11-15
Revert "updates php-cs-fixer for 2.0"

bshaffer:
The `php-cs-fixer` team has rolled this change back, so we now have to roll it back as well 👎

### DIFF
--- a/datastore/api/src/functions/concepts.php
+++ b/datastore/api/src/functions/concepts.php
@@ -733,7 +733,7 @@ function unindexed_property_query(DatastoreClient $datastore)
     // [START unindexed_property_query]
     $query = $datastore->query()
         ->kind('Task')
-        ->filter('description', '=', 'A task description.');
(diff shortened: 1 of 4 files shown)


datastore/api/src/functions/concepts.php
```diff
@@ -733,7 +733,7 @@ function unindexed_property_query(DatastoreClient $datastore)
     // [START unindexed_property_query]
     $query = $datastore->query()
         ->kind('Task')
-        ->filter('description', '=', 'A task description.');
+        ->filter('description',  '=', 'A task description.');
```
tmatsuo: Do you need to add the space?

tmatsuo:
Sigh, basically LGTM

datastore/api/src/functions/concepts.php
bshaffer: No, it was just added back automatically from the revert.

PR #253, opened 2016-11-26
Replaced 'swagger.yaml' with openapi.yaml. Refer to b/32670418.

jasminejaksic:


bshaffer:
Tests are failing because this is not on the `master` branch.

PR #254, opened 2016-11-28
Modify link used to enable the translate api

sgreenberg:
translate.googleapis.com will stop working.

Eng team suggests forcing the user to explicitly select a project.

### DIFF
--- a/translate/api/README.md
+++ b/translate/api/README.md
@@ -6,7 +6,7 @@ These samples show how to use the [Google Cloud Translate API](
 https://cloud.google.com/translate/).
 
 ## Build and Run
-1.  **Enable APIs** - [Enable the Translate API](https://console.cloud.google.com/flows/enableapi?apiid=translate.googleapis.com)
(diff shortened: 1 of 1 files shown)


bshaffer:
Testing `&project=_`, all it achieves is adding an extra step for the user to activate the API (instead of selecting a project and activating on one page), they are required to click through two pages). What's the reasoning here?

sgreenberg:
Hong wrote back, "The &project=_ forces the user to choose the right project instead of landing on a random project. Latter is more convenient but also more error prone. I prefer safe than sorry, but YMMV."

I think that's a reasonable argument though admittedly the tradeoff is one that could go either way. I'm okay making the change this way for this api (ie. following the eng team's recommendation)

sgreenberg:
@bshaffer, @SurferJeffAtGoogle  - Do you guys want to merge this or give me permissions to do so? Either is totally fine with me.

bshaffer:
You don't land on a random project, you select your project from a dropdown:
screen shot 2016-11-29 at 5 05 53 pm

Remove the `&project=_` and we can merge

PR #256, opened 2016-11-28
Updating additional swagger API references

jasminejaksic:


bshaffer:
This is covered in #255

PR #257, opened 2016-11-29
Fix encrypted file downloads and adds encryption key rotation

bshaffer:


storage/api/test/EncryptionCommandTest.php
```diff
@@ -107,4 +109,103 @@ public function testEncryptedFile()
...
+                ['interactive' => false]
+            );
+            $this->fail('An exception should have been thrown');
+        } catch (BadRequestException $e) {
+            // Expected exception
+        }
+
+        $this->assertContains(
+            'CustomerEncryptionKeyIsIncorrect',
+            $e->getMessage()
```
tmatsuo: Is this `$e` available outside of the `catch` clause. Maybe yes, but it looks strange.

bshaffer: I like it outside because it's the only path allowed (the `fail()` call in the `try` block will ensure `$e` exists). I do not have strong opinions on this though.

PR #258, opened 2016-11-29
updates memcache sample for env:flex memcache changes

bshaffer:


bshaffer:
@tmatsuo e2e tests pass, but it required 768c88e

tmatsuo:
It is definitely a problem if we don't have a sample code, but soon we'll have it :)

bshaffer:
@tmatsuo what do you mean? We have sample code. Do you mean having instructions on cloud.google.com?

Issue #259, opened 2016-11-30
Update composer.lock to cloud-tools to v6.0


bshaffer:
This can be done by running the following bash command:

```sh
find . -name composer.lock -not -path "*/vendor/*" \
    -exec sh -c 'pushd "${0%/*}" && composer update && popd' {} \;
```

tmatsuo:
Also we'd like to update google/cloud to '0.20'.

```bash
find . -name composer.json -not -path "*/vendor/*" -exec sh -c 'pushd "${0%/*}" && grep \"google/cloud\" composer.json && composer require google/cloud:'~0.20' google/auth:'~0.11' && composer update && popd' {} \;
```

```bash
find . -name composer.json -not -path "*/vendor/*" -exec sh -c 'pushd "${0%/*}" && grep \"google/cloud-tools\" composer.json && composer require --dev google/cloud-tools:'~0.6' && composer update && popd' {} \; 
```

and I get #271

PR #260, opened 2016-11-30
updates all samples for google/cloud-tools and adds delay to endpoints

bshaffer:
see #259


bshaffer:
Also adds back the commit for `php-cs-fixer:2.0` because that release is finally stable :-/

bshaffer:
This is waiting on #263 which fixes a bunch of tests

PR #263, opened 2016-12-06
Updates php cs fixer 2.0

bshaffer:
also fixes mailgun tests


appengine/flexible/analytics/app.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Copyright 2015 Google Inc.
```
tmatsuo: The year should be the year when you add the file, so you should not update it

bshaffer: Ahh, that's true. Unfortunately, all the headers have to be the same for the linter to work. What if we changed them all to `Copyright 2015, 2016 Google Inc.`?

Also, [the vast majority of these files](https://github.com/GoogleCloudPlatform/php-docs-samples/commits/fac876bfbbc1310ca664852b5ba52ac5f6868ae6/appengine/flexible/analytics/app.php) were added in 2016, but the copyright header wasn't updated when copied.

tmatsuo: Can you have exception for the linter? I also feel like it's a linter's bug

bshaffer: We can just not check the header, if that's what you mean, but no there is no option to skip the copyright header. I submitted https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2363 for this. But in the interim, we should decide to remove it, change it to `2015, 2016`, or set all of them to `2016`

tmatsuo: Maybe skip checking the header?

keradus: a lot of packages do this like `initial_year - current_year`, like `2015 - 2016` here.
if you claimed copyrights in 2015 you probably didn't throw it away year later, right ? :)

bshaffer: Definitely true, although should new files have a previous year's copyright?

keradus: treat the copyright per project, not per file.
then, all files should have same copyright.

you put copyrights in one, common place, for easy use, like LICENSE file. and then, copy-paste it to all files if one would only see one file, then for sure you don't want to get under impression the file is copyleft.

you probably don't want to keep different files in one repo under different copyright. year, as trivial, but still, is one of that case, when you make different copyright for different files.

.php_cs.dist
```diff
@@ -27,5 +27,6 @@ return PhpCsFixer\Config::create()
         PhpCsFixer\Finder::create()
             ->notPath('appengine/wordpress/src/files/flexible/wp-config.php')
             ->notPath('appengine/wordpress/src/files/standard/wp-config.php')
+            ->in(__DIR__)
```
keradus: FYI: without this you would have to specify the path at every single `fix` run, like `php-cs-fixer fix .` instead of `php-cs-fixer fix`

bshaffer: Thank you! That is helpful.

appengine/flexible/analytics/app.php
tmatsuo: > treat the copyright per project, not per file.

I don't think that's feasible. Files should be able to have different year in their copyright headers.

bshaffer:
FYI, I am keeping the copyright changes, because all of these files are actually `Copyright 2016` anyway:

```sh
$ git ls-tree -r --name-only HEAD | \
while read filename; do
  echo "$(git log --format="%aD" -- $filename | tail -1) $filename";
done | grep -v 2016 | awk '{print $4, $7}' 
2015 .gitignore
2015 .travis.yml
2015 CONTRIBUTING.md
2013 LICENSE
2012 README.md
2015 TRAVIS.md
2015 bigquery/api/README.md
2015 bigquery/api/composer.json
2015 bigquery/api/composer.lock
2015 favicon.ico
2015 storage/api/README.md
2015 storage/api/composer.json
2015 storage/api/composer.lock
```

tmatsuo:
LGTM

appengine/flexible/analytics/app.php
keradus: ok, I took a look at this project more and I would say that i would treat different folders separately, like you have different composer files. for that, when you don't do submodule or other solution, just keep everything in one place, it may be a valid case to have different headers.

at the same time, I would not treat it as regular case nor bug of fixer. it could be extended to support that case for sure, feel free to contribute if you want to change current behaviour ;)

PR #268, opened 2016-12-08
Update Endpoints samples:

sepehre:
1. Change default service names to the new cloud.goog format.
2. Change app.yaml beta_settings to endpoints_api_service.

### DIFF
--- a/endpoints/getting-started/app.yaml
+++ b/endpoints/getting-started/app.yaml
@@ -1,8 +1,6 @@
 runtime: php
 env: flex
 
-beta_settings:
-  # Enable Google Cloud Endpoints API management.
-  use_endpoints_api_management: true
-  # Specify the Swagger API specification.
-  endpoints_swagger_spec_file: openapi.yaml
+endpoints_api_service:
+  name: ENDPOINTS SERVICE-NAME # To be replaced.
+  config_id: ENDPOINTS CONFIG-ID # To be replaced.
--- a/endpoints/getting-started/app.yaml
+++ b/endpoints/getting-started/app.yaml
@@ -2,5 +2,7 @@ runtime: php
 env: flex
 
 endpoints_api_service:
-  name: ENDPOINTS SERVICE-NAME # To be replaced.
-  config_id: ENDPOINTS CONFIG-ID # To be replaced.
+  # The following values are to be replaced by information from the output of
+  # 'gcloud service-management deploy openapi.yaml' command.
+  name: ENDPOINTS SERVICE-NAME
+  config_id: ENDPOINTS CONFIG-ID
--- a/endpoints/getting-started/openapi.yaml
+++ b/endpoints/getting-started/openapi.yaml
@@ -3,7 +3,7 @@ info:
   description: "A simple Google Cloud Endpoints API example."
   title: "Endpoints Example"
   version: "1.0.0"
-host: "YOUR-PROJECT-ID.appspot.com"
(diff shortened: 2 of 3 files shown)


sepehre:
Please do not merge until documentation updated have been coordinated.

endpoints/getting-started/app.yaml
```diff
@@ -1,8 +1,8 @@
...
-beta_settings:
-  # Enable Google Cloud Endpoints API management.
-  use_endpoints_api_management: true
-  # Specify the Swagger API specification.
-  endpoints_swagger_spec_file: openapi.yaml
+endpoints_api_service:
+  # The following values are to be replaced by information from the output of
+  # 'gcloud service-management deploy openapi.yaml' command.
+  name: ENDPOINTS SERVICE-NAME
+  config_id: ENDPOINTS CONFIG-ID
```
bshaffer: The space here is confusing, can we make this `ENDPOINTS-SERVICE-NAME` and `ENDPOINTS-CONFIG-ID`?

endpoints/getting-started/openapi.yaml
```diff
@@ -3,7 +3,7 @@ info:
   description: "A simple Google Cloud Endpoints API example."
   title: "Endpoints Example"
   version: "1.0.0"
-host: "YOUR-PROJECT-ID.appspot.com"
+host: "echo-api.endpoints.YOUR-PROJECT-ID.cloud.goog"
```
bshaffer: is `echo-api` arbitrary, or does it have to do with the `service name` above?

sepehre: "echo-api" is arbitrary, any name in .endpoints.YOUR-PROJECT-ID.cloud.goog would do. This sample is the "echo sample" so I've named the API echo-api.endpoints...

endpoints/getting-started/app.yaml
sepehre: I left the space (ENDPOINTS" "SERVICE-NAME) in there intentionally because I want the "gcloud app deploy" command to fail if the users (who should be going through the Quickstart) do not update their app.yaml for whatever reason. The command would fail due to the AppInfo regex validation.

We are in the process of adding a better client side validation to make sure the name and config_id provided are actually valid, but unfortunately adding such a validation causes other concerns, so we are still working on it. In the meanwhile, if you provide an invalid (but syntactically correct) config_id, the deployment would proceed and fail much later after spinning for 20 minutes... which is horrible.

Let me know if you'd like to discuss the details.

bshaffer:
Please rebase and resolve conflicts!

endpoints/getting-started/app.yaml
bshaffer: I understand the reasoning, but I do not agree with it. We have a lot of samples with placeholders, and nowhere do we have an arbitrary space. The steps for deployment are explained in the README and/or on the tutorial page. This is sufficient.

What about illegal characters like `<` or `[`? We could surround the placeholder name in those so an error is thrown.

sepehre: Feel free to replace it with any placeholder that you like, as long as it fails "gcloud app deploy" if the placeholder is not replaced. Does that make sense?

As the oncall for this product, I will be taking on the operational burden of answering customer questions when they forget to replace the placeholder and the tools/APIs don't complain. So please consider that when worrying about the consistency of spaces in our product's samples.

sepehre:
Abandoning this PR, please merge https://github.com/GoogleCloudPlatform/php-docs-samples/pull/276 instead.

PR #275, opened 2016-12-13
properly skip datastore tests from other repos

bshaffer:
note: this is from my fork, to test it works

### DIFF
--- a/appengine/flexible/datastore/tests/LocalTest.php
+++ b/appengine/flexible/datastore/tests/LocalTest.php
@@ -25,6 +25,9 @@ class LocalTest extends WebTestCase
         if (!getenv('GCLOUD_PROJECT')) {
             $this->markTestSkipped('Must set GCLOUD_PROJECT');
         }
+        if (!getenv('GOOGLE_APPLICATION_CREDENTIALS')) {
+            $this->markTestSkipped('Must set GOOGLE_APPLICATION_CREDENTIALS');
+        }
         parent::setUp();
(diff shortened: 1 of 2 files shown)


appengine/flexible/datastore/tests/LocalTest.php
```diff
@@ -25,6 +25,9 @@ public function setUp()
         if (!getenv('GCLOUD_PROJECT')) {
             $this->markTestSkipped('Must set GCLOUD_PROJECT');
         }
+        if (!getenv('GOOGLE_APPLICATION_CREDENTIALS')) {
```
tmatsuo: `GOOGLE_APPLICATION_CREDENTIALS` is actually set in the `.travis.yaml` so it doesn't skip. I think you need to check the file existence.

tmatsuo:
`GOOGLE_APPLICATION_CREDENTIALS` should be set somewhere if you want to remove it from `.travis.yaml` anyways. I think it's more appropriate to check the file existance.

bshaffer:
@tmatsuo I added `GOOGLE_APPLICATION_CREDENTIALS` to the repo settings' environment variables. 

Also I disagree, I think checking if the file exists is a pain because it's an additional check.

bshaffer:
We could possibly add the `SkipIfMissingCredentials` trait to the testing repo and then add that to all the tests.

bshaffer:
Closing for now, we'll find a better way to do this.

PR #276, opened 2016-12-13
Update Endpoints samples:

sepehre:
1. Change default service names to the new cloud.goog format.
2. Change app.yaml beta_settings to endpoints_api_service.

### DIFF
--- a/endpoints/getting-started/app.yaml
+++ b/endpoints/getting-started/app.yaml
@@ -1,8 +1,8 @@
 runtime: php
 env: flex
 
-beta_settings:
-  # Enable Google Cloud Endpoints API management.
-  use_endpoints_api_management: true
-  # Specify the Swagger API specification.
-  endpoints_swagger_spec_file: openapi.yaml
+endpoints_api_service:
+  # The following values are to be replaced by information from the output of
+  # 'gcloud service-management deploy openapi.yaml' command.
+  name: ENDPOINTS SERVICE-NAME
+  config_id: ENDPOINTS CONFIG-ID
--- a/endpoints/getting-started/app.yaml
+++ b/endpoints/getting-started/app.yaml
@@ -3,6 +3,10 @@ env: flex
 
 endpoints_api_service:
   # The following values are to be replaced by information from the output of
-  # 'gcloud service-management deploy openapi.yaml' command.
-  name: ENDPOINTS SERVICE-NAME
-  config_id: ENDPOINTS CONFIG-ID
+  # 'gcloud service-management deploy openapi.yaml' command. If you have
+  # previously run the deploy command, you can list your existing configuration
+  # ids using the 'configs list' command as follows:
+  # 'gcloud service-management configs list --service=echo-api.endpoints.[PROJECT-ID].cloud.goog'
+  # where echo-api.endpoints.[PROJECT-ID].cloud.goog is your Endpoints service name.
+  name: ENDPOINTS SERVICE NAME
+  config_id: ENDPOINTS CONFIG ID
--- a/endpoints/getting-started/openapi.yaml
+++ b/endpoints/getting-started/openapi.yaml
@@ -4,7 +4,7 @@ info:
   description: "A simple Google Cloud Endpoints API example."
   title: "Endpoints Example"
   version: "1.0.0"
-host: "YOUR-PROJECT-ID.appspot.com"
+host: "echo-api.endpoints.YOUR-PROJECT-ID.cloud.goog"
 # [END swagger]
 basePath: "/"
 consumes:


sepehre:
New PR based on https://github.com/GoogleCloudPlatform/php-docs-samples/pull/268

endpoints/getting-started/app.yaml
```diff
@@ -1,8 +1,8 @@
...
 env: flex
 
-beta_settings:
-  # Enable Google Cloud Endpoints API management.
-  use_endpoints_api_management: true
-  # Specify the Swagger API specification.
-  endpoints_swagger_spec_file: openapi.yaml
+endpoints_api_service:
+  # The following values are to be replaced by information from the output of
+  # 'gcloud service-management deploy openapi.yaml' command.
```
bshaffer: This needs to be `gcloud beta service-management deploy`

sepehre: No it doesn't. Please see CL 141482339 in google3 for the corresponding doc change.

bshaffer: The output of this command does not make it clear where the service name / config ID are:

```
Waiting for async operation operations/serviceConfigs.echo-api.endpoints.getting-started-php.cloud.goog:xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx to complete...
Operation finished successfully. The following command can describe the Operation details:
 gcloud beta service-management operations describe operations/serviceConfigs.echo-api.endpoints.getting-started-php.cloud.goog:xxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
Waiting for async operation operations/rollouts.echo-api.endpoints.getting-started-php.cloud.goog:2016-12-13r1 to complete...
Operation finished successfully. The following command can describe the Operation details:
 gcloud beta service-management operations describe operations/rollouts.echo-api.endpoints.getting-started-php.cloud.goog:2016-12-13r1

Service Configuration with version [2016-12-13r1] uploaded for service [echo-api.endpoints.getting-started-php.cloud.goog]
```

 Is this command changing to include the output?

endpoints/getting-started/app.yaml
```diff
@@ -1,8 +1,8 @@
...
+  name: ENDPOINTS SERVICE-NAME
+  config_id: ENDPOINTS CONFIG-ID
```
bshaffer: As mentioned in the last PR, please be consistent with placeholder names by making this `ENDPOINTS-SERVICE-NAME` and `ENDPOINTS-CONFIG-ID`.

endpoints/getting-started/app.yaml
sepehre: What version of gcloud are you running?

In the latest version, the last output line should be 
Service Configuration [2016-12-13r1] uploaded for service [echo-api.endpoints.getting-started-php.cloud.goog]

And the docs clarify further that 2016-12-13r1 is the configuration Id.

endpoints/getting-started/app.yaml
jeffmendoza: I thought the decision was:

> What about illegal characters like < or [? We could surround the placeholder name in those so an error is thrown.

sepehre: And as mentioned in the last PR, the space is left in there intentionally in order to avoid a horrible developer experience for customers some of whom will forget to replace these values when deploying our samples for the first time.

Please see apphosting/api/appinfo.py in google3. The automatic validation of these fields only looks for white space. < and [ are not going to fail the command. The real validation (client side and server side) to ensure that the values are correct (i.e. not just seemingly correct) will be done in in bugs 33010559 and 33542830. After that, I can come back and remove this white space. Deal?

jeffmendoza: The problem here though is, the docs need to be the same. There is a single quickstart page, with the same instructions for all languages. The code snippets are in a tabbed window.

https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/endpoints/getting-started/app.yaml
https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/endpoints/getting-started/app.yaml
https://github.com/GoogleCloudPlatform/golang-samples/blob/master/endpoints/getting-started/app.yaml
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/endpoints/getting-started/app.yaml

endpoints/getting-started/app.yaml
bshaffer: Until the aforementioned CL is merged, we can't merge this PR. So I will wait until `beta` is no longer required.

As for the config, I am running the latest version of gcloud. I followed [the docs link in the readme](https://cloud.google.com/endpoints/docs/quickstarts) and could not find the configuration id output explained. If we can add a link to where the docs clarify this in the comment, that will suffice.

endpoints/getting-started/app.yaml
bshaffer: I would prefer this read `ENDPOINTS SERVICE NAME` and `ENDPOINTS CONFIG ID`, but this is LGTM because you gotta pick your battles.

endpoints/getting-started/app.yaml
sepehre: I've updated the comment to include the 'configs list' command also.

endpoints/getting-started/app.yaml
sepehre: Changed to remove the - as you suggested.

PR #277, opened 2016-12-14
adds environment variable for analytics tracking

bshaffer:


appengine/flexible/analytics/app.yaml
```diff
@@ -1,5 +1,8 @@
 runtime: php
 env: flex
 
+env_variables:
+  GA_TRACKING_ID: "YOUR-GA-TRACKING-ID"
+
 runtime_config:
-  document_root: web
+  document_root: ""
```
tmatsuo: You can omit the entire `runtime_config`

bshaffer: done.

PR #279, opened 2016-12-15
Adds Cloud Storage sample for AppEngine Flex

bshaffer:


appengine/flexible/storage/README.md
```diff
@@ -0,0 +1,55 @@
+# Cloud Storage & Google App Engine
+
+This sample application demonstrates how to use [Cloud Storage with Google App Engine](https://cloud.google.com/appengine/docs/flexible/php/using-cloud-storage).
+
+## Setup
+
+Before running this sample:
```
tmatsuo: Probably worth adding project and bucket creation?

appengine/flexible/storage/app.php
```diff
@@ -0,0 +1,80 @@
...
+    <h1>Storage Example</h1>
+    <h3>Write [<a href="https://cloud.google.com/appengine/docs/flexible/php/using-cloud-storage">docs</a>]:</h3>
+    <form action="/write" method="post">
+        Some file content:<br />
+        <textarea name="content"></textarea><br />
+        <input type="submit" />
+    </form>
+EOF;
+    if ($content) {
+        $form .= "<p><strong>Your content:</strong><p><p>$content</p>";
```
tmatsuo: The `$content` is not sanitized, so it's basically XSS vulnerable.

I understand the rationale of using inline html for simplicity, but I would use html template library and let it sanitize the value.

appengine/flexible/storage/app.yaml
```diff
@@ -0,0 +1,7 @@
+runtime: php
+vm: true
+
+# [START env_variables]
+env_variables:
+  GOOGLE_BUCKET_NAME: "your-bucket-name"
+# [END env_variables]
```
tmatsuo: Small nit: add newline?

appengine/flexible/storage/app.php
bshaffer: How about we use `htmlspecialchars` instead?

tmatsuo: Sounds ok too

appengine/flexible/storage/README.md
bshaffer: added!

appengine/flexible/storage/app.yaml
bshaffer: done

PR #281, opened 2016-12-19
adds flex logging samples

bshaffer:


appengine/flexible/logging/app.php
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use Google\Cloud\Logging\LoggingClient;
+use Google\Cloud\Logging\AppEngineFlexHandler;
```
tmatsuo: This class is at `Google\Cloud\Logger\AppEngineFlexHandler`: https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/Logger/AppEngineFlexHandler.php

bshaffer: fixed

PR #282, opened 2016-12-21
Fix GCR reference

kyessenov:


endpoints/getting-started/container-engine.yaml
```diff
@@ -40,7 +40,7 @@ spec:
       containers:
       # [START esp]
       - name: esp
-        image: b.gcr.io/endpoints/endpoints-runtime:0.3
+        image: gcr.io/endpoints-release/endpoints-runtime:1
```
tmatsuo: Is it possible to use `latest` tag?

What happens if you release the next image with the tag `2`? Should we update our sample?

kyessenov: We do not have "latest" tag. :1 tracks 1.x.y, :1.0 track 1.0.x, and so on.
Next major release ":2" may not be backwards-compatible.

tmatsuo: makes sense

tmatsuo:
Re: test failure

The test is broken from PRs from other repos. I will create a new PR from the same repo (with the administrator's power) and we'll see the test result there.

tmatsuo:
@kyessenov I just created #283

tmatsuo:
The code LGTM, so I'll merge #283 once the test passes

kyessenov:
Thanks for taking care of the test!

tmatsuo:
Closing in favor of #283

PR #286, opened 2016-12-28
Fix typo in README.md

sgreenberg:
cd php-docs-samples/vision/api should be cd php-docs-samples/translate/api


tmatsuo:
Thanks @sgreenberg !

PR #291, opened 2017-01-05
adds region tags to flex memcache

bshaffer:


appengine/flexible/memcache/app.php
```diff
@@ -27,25 +27,30 @@
...
     $username = getenv('MEMCACHE_USERNAME');
     $password = getenv('MEMCACHE_PASSWORD');
+    # [START memcached]
+    // $host = 'YOUR_MEMCACHE_HOST';
+    // $port = 'YOUR_MEMCACHE_PORT';
+    // $username = 'OPTIONAL_MEMCACHE_USERNAME';
+    // $password = 'OPTIONAL_MEMCACHE_PASSWORD';
     $memcached = new Memcached;
     if ($username && $password) {
         $memcached->setOption(Memcached::OPT_BINARY_PROTOCOL, true);
```
tmatsuo: Is it(binary protocol) really needed? I'm skeptical because the memcached extension is compiled without binary protocol support.

bshaffer: It's optional, but I think it's good to show how to use it, since we should encourage it. 

Are you sure about it not being supported? In this sample, when I include the `php.ini` param, the SASL auth works as expected.

tmatsuo: Yeah look at the phpinfo or similar, you'll see it's disabled

tmatsuo: Memcached::setOption just returns false on failure. I suspect it just returns false.

Also iirc, I successfully communicate via sasl auth with redislab, without setting the `OPT_BINARY_PROTOCOL` option.

bshaffer: This was throwing an error w/o being set. I'll look into it though and see if I missed something.

bshaffer: @tmatsuo I'm not sure where the confusion is, but if I comment out the line above, the calls to Memcache fail. I have looked at the php.ini configuration, and do not see anything that implies binary protocol is missing.

tmatsuo: Sorry I was wrong. The memcached module doesn't have igbinary option enalbed, which is different from the binary protocol. I confused those two.

bshaffer: Thank you Takashi!

PR #299, opened 2017-01-12
adds error reporting quickstart

bshaffer:


error_reporting/quickstart/test/quickstartTest.php
```diff
@@ -0,0 +1,45 @@
...
+        );
+        file_put_contents($file, $contents);
+
+        // Invoke quickstart.php
+        ob_start();
+        passthru(sprintf('php %s', $file));
+        $output = ob_get_clean();
+
+        // Make sure it looks correct
+        $this->assertEquals(
```
tmatsuo: This is testing whether or not it succesfully wrote the log to Cloud Logging API.
Is there a way to make sure it's correctly handled by the Error Reporting?

waprin:
lgtm

error_reporting/quickstart/test/quickstartTest.php
bshaffer: @tmatsuo I can probably make an API call, yeah

bshaffer: Done!

error_reporting/quickstart/test/quickstartTest.php
```diff
@@ -41,5 +47,42 @@ public function testQuickstart()
             'Exception logged to Stack Driver Error Reporting' . PHP_EOL,
             $output
         );
+
+        // Ensure the log actually showed up
+        sleep(1);
```
tmatsuo: I think there is EventualConsistentTestRunner somewhere. Does it make more sense to use that?

bshaffer: Good call! Added.

PR #300, opened 2017-01-13
Fix vision samples

bshaffer:
adds region tags, static images, and updates authentication


bshaffer:
@tmatsuo we are waiting for a fix on the Speech API that should go out next week. I don't really want to update our samples to fix it. I agree it'd be pretty slick to only run select samples for tests!

I added logic to [write a bounding box](bf0c814) for vision samples. LMK what you think.

vision/api/src/DetectFaceCommand.php
```diff
@@ -48,19 +64,49 @@ protected function configure()
...
+                throw new \Exception('Unsupported image extension');
+            }
+            $outputImage = call_user_func($this->imageCreateFunc[$ext], $outFile);
+            # [START highlight_image]
+            foreach ($result->info()['faceAnnotations'] as $annotation) {
+                if (isset($annotation['boundingPoly'])) {
+                    $verticies = $annotation['boundingPoly']['vertices'];
+                    $x1 = isset($verticies[0]['x']) ? $verticies[0]['x'] : 0;
+                    $y1 = isset($verticies[0]['y']) ? $verticies[0]['y'] : 0;
+                    $x2 = isset($verticies[2]['x']) ? $verticies[2]['x'] : 0;
```
tmatsuo: For `$x2` and `$y2`, is it really 0 for the default value?

bshaffer: It may be a bug in the google-cloud-php library. I'll take a look.

PR #302, opened 2017-01-14
adding the KMS samples for creating keys, keyrings, encrypt and decrypt

bshaffer:
todo: 
 - [x] add tests
 - [ ] write samples for other KMS functions
 - [x] add README


kms/api/decrypt.php
```diff
@@ -0,0 +1,32 @@
...
+$ciphertext = file_get_contents($infile);
+
+$request = new Google_Service_CloudKMS_DecryptRequest();
+$request->setCiphertext($ciphertext);
+$response = $kms->projects_locations_keyRings_cryptoKeys->decrypt(
+    $keyName,
+    $request
+);
+
+// The plaintext response comes back base64 decoded.
```
tmatsuo: I think the comment here should be:

```
// The plaintext response comes back base64 encoded.
```

bshaffer:
Done! PTAL

kms/api/test/KmsTest.php
```diff
@@ -0,0 +1,167 @@
...
+        $client->setScopes([
+            'https://www.googleapis.com/auth/cloud-platform'
+        ]);
+        $this->kms = new Google_Service_CloudKMS($client);
+    }
+
+    public function testCreateCryptoKey()
+    {
+        if (!$keyRing = getenv('GOOGLE_KMS_KEYRING')) {
+            return $this->markTestSkipped('Set the GOOGLE_KMS_KEYRING environment variable');
```
tmatsuo: Can you run all the tests on travis? (or in other words, have you set the environment variables)

bshaffer: All set up!

PR #305, opened 2017-01-21
Add more kms samples

bshaffer:


kms/api/src/EncryptionCommand.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        $this
+            ->setName('encryption')
+            ->setDescription('Manage encryption for KMS')
+            ->setHelp(<<<EOF
+The <info>%command.name%</info> command manages KMS encryption.
+
+    <info>php %command.full_name% my-keyring my-cryptokey file.txt file.txt.encrypted</info>
+
+    <info>php %command.full_name% my-keyring my-cryptokey file.txt.encrypted file.txt.decrypted</info>
```
tmatsuo: Don't you need `--decrypt` in the second example?

kms/api/src/IamCommand.php
```diff
@@ -0,0 +1,144 @@
...
+    {
+        $this
+            ->setName('iam')
+            ->setDescription('Manage IAM for KMS')
+            ->setHelp(<<<EOF
+The <info>%command.name%</info> command manages KMS IAM policies.
+
+    <info>php %command.full_name% my-keyring</info>
+
+    <info>php %command.full_name% my-keyring my-cryptokey</info>
```
tmatsuo: Are these example correct?

kms/api/src/KeyCommand.php
```diff
@@ -0,0 +1,150 @@
...
+    {
+        $this
+            ->setName('key')
+            ->setDescription('Manage keys for KMS')
+            ->setHelp(<<<EOF
+The <info>%command.name%</info> command manages KMS keys.
+
+    <info>php %command.full_name% my-keyring</info>
+
+    <info>php %command.full_name% my-keyring my-cryptokey</info>
```
tmatsuo: Are these example correct?

kms/api/kms.php
```diff
@@ -0,0 +1,32 @@
+<?php
+/**
+ * Copyright 2016 Google Inc.
```
tmatsuo: 2017?

kms/api/src/EncryptionCommand.php
bshaffer: yes! good catch.

kms/api/src/IamCommand.php
bshaffer: Yes, they will list the IAM policies for those resources.

kms/api/src/KeyCommand.php
bshaffer: Yes, the first one lists the keys, the second one returns the info about that specific key.

kms/api/kms.php
bshaffer: fixed!

kms/api/src/EncryptionCommand.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        if (!$projectId = $input->getOption('project')) {
+            $projectId = $this->getProjectIdFromGcloud();
+        }
+        $keyRing = $input->getArgument('keyring');
+        $cryptoKey = $input->getArgument('cryptokey');
+        $infile = $input->getArgument('infile');
+        $outfile = $input->getArgument('outfile');
+        $location = $input->getOption('location');
+        if (!$input->getOption('decrypt')) {
```
tmatsuo: Small nit: I would not use `!` for this `if`.

kms/api/src/VersionCommand.php
```diff
@@ -0,0 +1,192 @@
...
+
+    protected function configure()
+    {
+        $this
+            ->setName('version')
+            ->setDescription('Manage key versions for KMS')
+            ->setHelp(<<<EOF
+The <info>%command.name%</info> command manages KMS keys.
+
+    <info>php %command.full_name% my-keyring my-cryptokey 1</info>
```
tmatsuo: What is `1` in the example?

tmatsuo:
In my opinion, some of the help text actually don't help me understand what the command is doing.

kms/api/src/VersionCommand.php
```diff
@@ -0,0 +1,192 @@
...
+{
+    use KmsCommandTrait;
+
+    protected function configure()
+    {
+        $this
+            ->setName('version')
+            ->setDescription('Manage key versions for KMS')
+            ->setHelp(<<<EOF
+The <info>%command.name%</info> command manages KMS keys.
```
tmatsuo: versions of the KMS keys?

tmatsuo:
Having hand written help is generally good! Although, if we have one, it should be something really help users to understand what it is doing.

For example, the encryption help is

```
The <info>%command.name%</info> command manages KMS encryption
```

Why not, `command encrypt or decrypt the given file` or something like that?

kms/api/src/VersionCommand.php
bshaffer: The cryptokey version

kms/api/src/EncryptionCommand.php
bshaffer: done!

kms/api/src/VersionCommand.php
bshaffer: added help text for this!

kms/api/src/VersionCommand.php
bshaffer: fixed

bshaffer:
I've added help text and more examples for every command! LMK what you think.

kms/api/src/IamCommand.php
```diff
@@ -0,0 +1,159 @@
...
+Add a service account to a CryptoKey:
+
+    <info>php %command.full_name% my-keyring my-cryptokey \
+        --service-account-email=example@my-project.gserviceaccount.com</info>
+
+Add a service account to a CryptoKey for a specific role:
+
+    <info>php %command.full_name% my-keyring my-cryptokey \
+        --service-account-email=example@my-project.gserviceaccount.com \
+        --role=roles\cloudkms.admin</info>
```
tmatsuo: `roles\cloudkms.admin`

What is the backslash above?

kms/api/src/VersionCommand.php
```diff
@@ -0,0 +1,200 @@
...
+    {
+        $this
+            ->setName('version')
+            ->setDescription('Manage key versions for KMS')
+            ->setHelp(<<<EOF
+The <info>%command.name%</info> command manages KMS key versions.
+
+List all versions of a CryptoKey:
+
+    <info>php %command.full_name% my-keyring my-cryptokey 1</info>
```
tmatsuo: Is `1` right here?

kms/api/src/IamCommand.php
bshaffer: It's so it can be copied and pasted into a shell. I think it's a good idea?

kms/api/src/VersionCommand.php
bshaffer: No, it isn't! Good catch.

kms/api/src/IamCommand.php
bshaffer: Oh sorry, I was thinking of the wrong backslash. We could remove this and automatically prepend `roles/` to the option.

PR #309, opened 2017-01-27
adds flex metadata example

bshaffer:


appengine/flexible/metadata/README.md
```diff
@@ -0,0 +1,45 @@
+# Google App Engine Metadata
```
tmatsuo: The title says `Google App Engine Metadata`, but we say `Compute Metadata` below

appengine/flexible/metadata/README.md
```diff
@@ -0,0 +1,45 @@
...
+## Setup
+
+Before running this sample:
+
+### Register your application
+
+- Go to
+  [Google Developers Console](https://console.developers.google.com/project)
+  and create a new project. This will automatically enable an App
+  Engine application with the same ID as the project.
```
tmatsuo: `This will automatically enable an App Engine application with the same ID as the project` Is it still true?

bshaffer: hmm, I don't actually know what this means.  I'll just remove this line.

PR #323, opened 2017-02-22
Made the datastore tests less flaky

tmatsuo:
TODO

- [x] Upgrade google/php-tools to 0.6.3

### DIFF
--- a/datastore/api/composer.lock
+++ b/datastore/api/composer.lock
@@ -100,16 +100,16 @@
         },
         {
             "name": "google/cloud",
-            "version": "v0.20.0",
+            "version": "v0.20.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/GoogleCloudPlatform/google-cloud-php.git",
-                "reference": "0a22f803246e64f398220df9abeb818f9183907f"
+                "reference": "7175deb3c4b07d790a52214ba209986404857d51"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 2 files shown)


bshaffer:
Personally, I'd be ok just hardcoding a number, or incrementing the default in https://github.com/GoogleCloudPlatform/php-tools

tmatsuo:
@bshaffer 

I think the current default `3` in the tool is somewhat appropriate. I looked at the logs of the recent failures and realized that we should use separate namespace for each test setup, which will likely be sufficient for stabler test result.

Also, I'd like to have an easy way to experiment with the retry count once if the datastore tests become flaky.

bshaffer:
We could still do better than this, i.e. have the Trait use a default static variable `eventuallyConsistentRetryCount` and then override it in the test classes. Passing it into every single call is not optimal.

tmatsuo:
@bshaffer Yeah I tried that path first and I agree. https://github.com/GoogleCloudPlatform/php-tools/pull/19

tmatsuo:
@bshaffer PTAL

tmatsuo:
Running the tests on jenkins

tmatsuo:
The tests on our jenkins are not green so far for several failure.
1) Datastore eventually consistency error
I hadn't set `DATASTORE_EVENTUALLY_CONSISTENT_RETRY_COUNT` when I saw this error, so I set it to 10.

2) Some 502 errors after flex deployment
Sigh, it is still happening. I raised `GOOGLE_DEPLOYMENT_DELAY` to 60.

tmatsuo:
The tests on jenkins are still somewhat flaky, but the last build was green.
Kicked off the travis build, and I'll merge once it's green too.

PR #324, opened 2017-02-22
add translate model premium feature

bshaffer:


tmatsuo:
@bshaffer FYI https://github.com/GoogleCloudPlatform/google-cloud-php/issues/343

bshaffer:
Yes, this is because our test project `cloud-samples-tests-php` is not yet whitelisted for the Translate Premium beta.

PR #325, opened 2017-02-23
Changes for env: flex

tmatsuo:
Now we're using the GCS plugin from the github release.


tmatsuo:
Stopped the travis build, I will run the tests on jenkins instead :)

tmatsuo:
Hmm our jenkins is lazy too.

The tests passed locally, so I'm merging this now.

PR #326, opened 2017-02-25
Download the WordPress GCS plugin from wordpress.org

tmatsuo:


tmatsuo:
@dpebot merge when green

tmatsuo:
hey @dpebot merge when green

PR #328, opened 2017-02-28
sadly, double-quotes in app.yaml break cloudsql v2

bshaffer:


tmatsuo:
Did this start happening recently and did our test catch it?

PR #334, opened 2017-03-18
improvements and tests for laravel

bshaffer:
And replaces README with community article.


appengine/flexible/laravel/app.yaml
```diff
@@ -4,26 +4,12 @@ env: flex
 runtime_config:
   document_root: public
 
+# required on some platforms so ".env" is not skipped
+skip_files: false
```
tmatsuo: Is it a standard way to do it? Other language might use `.env` file too, but do they do the similar thing like `skip_files: false`?
I'm asking because I feel it's too much to disable all the skip_files.

appengine/flexible/laravel/app-dbsessions.yaml
```diff
@@ -0,0 +1,29 @@
+runtime: php
```
tmatsuo: Why is this file named `app-dbsessions.yaml`? When I looked at the test code, it's like `app.yaml` template.

appengine/flexible/laravel/app.yaml
bshaffer: hmm, good question. We could do something like this to be more precise:

```yaml
skip_files:
- ^(.*/)?#.*#$
- ^(.*/)?.*~$
- ^(.*/)?.*\.py[co]$
- ^(.*/)?.*/RCS/.*$
- ^(.*/)?\.(?!env$).*$
```

This grabs the [default skip_files config](https://cloud.google.com/appengine/docs/standard/php/config/appref) and changes the regex matching the dot prefix to exclude `.env`.

Another option would be to add it as `env` or `env.prod` and then add a `post-deploy-cmd` that moved the file to `.env`. This wouldn't be terrible, since it would allow separate local configuration.

appengine/flexible/laravel/app-dbsessions.yaml
bshaffer: Because we have two `app.yaml` files to test - one for a simple deploy, and one for a deployment using DB sessions.

appengine/flexible/laravel/app.yaml
tmatsuo: How about to have the skip_files for now, and then file a feature request to change it to support `.env`? I think `.env` is widely used and worthwhile to whitelist.

WDYT?

appengine/flexible/laravel/app-dbsessions.yaml
tmatsuo: I see

appengine/flexible/laravel/app.yaml
bshaffer: Hmm, I tested it and the skip files above doesn't work, even though the regex is correct. I think a better solution would be to add a `skip_files` whitelist, as overriding the regex, instead of adding something simpler like below, is not a great user experience. I.e.:

```yaml
include_files: [.env]
```

bshaffer: Merging for now, since this matches our community article, and because the regex to exclude is not working.

PR #336, opened 2017-03-21
fixes appengine standard http tests

bshaffer:


bshaffer:
I had to explicitly add `phpunit:4` to every language because PHP7 comes with `phpunit 6` by default, which is not compatible with PHP5.6 or with our current tests.

I think this is the best way to solve the issue for now.

PR #340, opened 2017-04-04
Test client library dependencies are up-to-date

bshaffer:


tmatsuo:
@bshaffer Answering my own question. It only list the outdated packages :)

PR #344, opened 2017-04-07
defaults logger sample to the last 24 hours

bshaffer:


bshaffer:
Closing, as tests require both this PR and speech updates to pass. Combining into #342.

PR #346, opened 2017-04-11
Testing grpc install

bshaffer:
Testing if `pecl install grpc` runs w/o trusty dist


bshaffer:
Confirmed that removing `distr: trusty` fails to install grpc.

PR #358, opened 2017-05-04
Create openapi-appengine.yaml

p-buse:


bshaffer:
Have you verified this works? If it's for appengine we should be able to add testing to the existing deployment tests!

p-buse:
I don't think a new test is necessary. This is a cosmetic change that makes the code more accurately follow the Endpoints on App Engine quickstart.

bshaffer:
This is a copy of `openapi.yaml`, with the `host` field changed. @p-buse will add a comment to the existing `openapi.yaml` instead

PR #364, opened 2017-05-11
Refactor dependency checker and test scripts

bshaffer:


bshaffer:
@tmatsuo The value is that the makefile is a place for all relevant executables for the repository, and it's a standard, so people know where to look. What are the downsides to having it?

tmatsuo:
You don't need it because there's no dependencies. Maybe a composer script is more familiar to PHP developers.

tmatsuo:
Downside is that it is adding an unnecessary dependency. Unnecessary dependency is almost always bad. For example, does it work with [a Container Builder based test runner](https://github.com/GoogleCloudPlatform/cloudbuild-test-runner-example)?

bshaffer:
@tmatsuo I understand your point, but it doesn't really add a dependency as any of the scripts can be run individually. 

I added it for convenience, and also it provides a `make dependency-update` task which I don't really want to put in the `testing/` directory, and don't want to make another root directory `scripts`.

tmatsuo:
I mean `make` is a new dependency. I don't understand why adding a new script is worse than introducing a new dependency. If you want to have a central place for the tasks, why not use composer script?

bshaffer:
Composer runs groups of scripts, and these scripts don't need to be run together. It would look like this:

```json
"scripts": {
    "testing": ["testing/run_test_suite.sh"],
    "cs": ["testing/run_cs_check.sh"],
    "check-dependencies": ["testing/run_dependency_check.sh"],
    "update-dependencies": ["testing/update_dependencies.sh"],
}
```

```sh
composer run-script check-dependencies
```

This isn't really the right place for this, as usually composer scripts are used to hook into the dependencies they pull in, such as `post-install-cmd`.

tmatsuo:
Hooks. Yes. That's the most important use case of the composer scripts, but it's also legal to create a custom commands :)
https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands

tmatsuo:
The beloved google-cloud-php has the composer script:
https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/composer.json#L76

bshaffer:
Yes, I know it's possible to put custom scripts in there, and I don't have a problem with doing that, but the advantage to the makefile is we can execute more complex commands w/o having to have a script file, i.e. the `dependency-update` command.

bshaffer:
ok, I removed the Makefile to appease @tmatsuo  :-P

tmatsuo:
Great :) I believe it's not only me who please.

PR #368, opened 2017-05-15
Update language samples to dump all data returned from server.

SurferJeffAtGoogle:
According to an earlier discussion, we still want to show users a list of all
the features that will be returned by the server.  So, I added a sample
invocation to the help documentation that serves that purpose.

### DIFF
--- a/language/api/src/AllCommand.php
+++ b/language/api/src/AllCommand.php
@@ -60,103 +60,12 @@ tokens:
       tag: NOUN
       aspect: ASPECT_UNKNOWN
       case: CASE_UNKNOWN
-      form: FORM_UNKNOWN
-      gender: GENDER_UNKNOWN
-      mood: MOOD_UNKNOWN
-      number: SINGULAR
-      person: PERSON_UNKNOWN
-      proper: PROPER
-      reciprocity: RECIPROCITY_UNKNOWN
-      tense: TENSE_UNKNOWN
-      voice: VOICE_UNKNOWN
+      ...
     dependencyEdge:
       headTokenIndex: 1
       label: NSUBJ
     lemma: John
-  -
-    text:
-      content: took
-      beginOffset: 5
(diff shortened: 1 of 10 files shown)


language/api/src/AllCommand.php
```diff
@@ -41,6 +41,149 @@ protected function configure()
...
+      number: SINGULAR
+      person: PERSON_UNKNOWN
+      proper: PROPER
+      reciprocity: RECIPROCITY_UNKNOWN
+      tense: TENSE_UNKNOWN
+      voice: VOICE_UNKNOWN
+    dependencyEdge:
+      headTokenIndex: 1
+      label: NSUBJ
+    lemma: John
```
bshaffer: I do not think it's useful to have this much in the example output. How about a single token block with an ellipsis,  to indicate there will be more.

language/api/src/SyntaxCommand.php
```diff
@@ -40,7 +40,122 @@ protected function configure()
...
+      person: PERSON_UNKNOWN
+      proper: PROPER
+      reciprocity: RECIPROCITY_UNKNOWN
+      tense: TENSE_UNKNOWN
+      voice: VOICE_UNKNOWN
+    dependencyEdge:
+      headTokenIndex: 1
+      label: NSUBJ
+    lemma: John
+  -
```
bshaffer: Same here. We don't want the user to have to scroll up to get the full help text.

language/api/src/AllCommand.php
SurferJeffAtGoogle: Done.

language/api/src/SyntaxCommand.php
SurferJeffAtGoogle: Done.

SurferJeffAtGoogle:
Ready for another look.

Issue #373, opened 2017-05-22
Remove unnecessary test/bootstrap.php files


bshaffer:
If `phpunit.xml.dist` has the bootstrap as `test/bootstrap.php`, and `test/bootstrap.php` only contains a require for `vendor/autoload.php`, then we can eliminate the bootstrap file and set the bootstrap to load the composer autoloader directly.

There are quite a few samples where this is the case.

PR #386, opened 2017-06-05
standardize and clean up tests

bshaffer:


tmatsuo:
Can you summarize 1) changes and 2) how did you make those changes ? Should I really review all the 98 files?

bshaffer:
@tmatsuo thanks! yes, good idea.

1. Rename every `phpunit.xml` file to `phpunit.xml.dist`.
2. Add Google copyright header to every `phpunit.xml.dist`.
3. Indent every `phpunit.xml.dist` file to 4 spaces
4. Add `<testsuite name="...">` to every `phpunit.xml.dist`
5. If `test/bootstrap.php` was only including `vendor/autoload.php`, remove `test/bootstrap.php` and specify `vendor/autoload.php` as the bootstrap file in `phpunit.xml.dist` (see #373)

PR #401, opened 2017-07-11
adds all monitoring samples

bshaffer:


bshaffer:
These tests are failing because of an issue in Stackdriver Monitoring. Unfortunately there's not a straightforward way to fix this without switching everything over to a different project.

PR #403, opened 2017-07-13
Add video samples

bshaffer:


video/video.php
```diff
@@ -22,46 +22,53 @@
...
+$application = new Application('Cloud Video Intelligence');
 
-$application->add(new Command('shots'))
-    ->setDescription('Detect shot changes in video using '
-        . 'Google Cloud Video Intelligence API')
-    ->setHelp(<<<EOF
-The <info>%command.name%</info> command finds shot changes in a video using the 
-Google Cloud Video Intelligence API.
+$inputDefinition = new InputDefinition([
+    new InputArgument('uri', InputArgument::REQUIRED, 'URI pointing to a video.'),
```
ryanmats: Does this have to be a Cloud Storage URI? If so, might be good to specify here.

bshaffer: Yes, currently only Cloud Storage URIs are supported. Good call, I'll add that.

video/src/analyze_faces.php
```diff
@@ -0,0 +1,63 @@
...
+
+    # Wait for the request to complete.
+    $operation->pollUntilComplete();
+
+    # Print the result.
+    if ($operation->operationSucceeded()) {
+        $results = $operation->getResult()->getAnnotationResults()[0];
+        foreach ($results->getFaceAnnotations() as $face) {
+            foreach ($face->getLocations() as $location) {
+                if (!$box = $location->getBoundingBox()) {
```
bshaffer: This conditional can be removed ~7/17

PR #407, opened 2017-07-24
Add openapi-appengine.yaml file for App Engine deployments.

p-buse:
After conversation with the Endpoints team, this was the most effective solution we came up with to guide users through setting up Endpoints on App Engine. Having users uncomment code proved to be difficult. If you have another idea that could work better, please let me know. Otherwise, this will have to do.


bshaffer:
Seems fine, but if `openapi-appengine.yaml` is for App Engine, what is `openapi.yaml` for? Is it for compute engine, container engine, etc.

p-buse:
That's correct. GKE and GCE use the PROJECT-ID.cloud.goog host format, whereas App Engine uses PROJECT-ID.appspot.com.

bshaffer:
This is LGTM, is there anything else needed before merge?

p-buse:
Nope, this is ready to go.

PR #408, opened 2017-07-25
Code samples for Vision 1.1 features

ryanmats:


vision/api/src/snippets/detect_crop_hints.php
```diff
@@ -0,0 +1,41 @@
...
+    'projectId' => $projectId,
+]);
+$image = $vision->image(file_get_contents($path), ['CROP_HINTS']);
+$result = $vision->annotate($image);
+print("Crop Hints:\n");
+foreach ((array) $result->cropHints() as $hint) {
+    $boundingPoly = $hint->boundingPoly();
+    $vertices = $boundingPoly['vertices'];
+    foreach ((array) $vertices as $vertice) {
+        if (!isset($vertice['x'])) $vertice['x'] = 0;
```
ryanmats: @bshaffer It seems like I need this check here, otherwise I get an error if one of the hints has a 0-coordinate for x or y: 'Undefined index: x in /usr/local/google/home/ryanmats/php-docs-samples/vision/api/src/snippets/detect_crop_hints.php on line 38'.

However, it looks like none of the other code samples (https://cloud.google.com/vision/docs/detecting-crop-hints) require this check. Am I doing something wrong with the PHP syntax, or is our client library different from the others with regard to 0-coordinates?

bshaffer: This *might* be an issue with the client libraries, i.e. the API is coming back with values and the library isn't propagating them through. However, I've seen this with other APIs were fields were missing in the API requests.

Something you could do is try making a [REST request to the Vision API](https://developers.google.com/apis-explorer/#p/vision/v1/vision.images.annotate) and see if the proper values are coming back. If they are, then it's most likely an issue with the client library.

bshaffer: While worth looking into (because this is a bit hackish), this solution is ok for now.

vision/api/src/snippets/detect_document_text.php
```diff
@@ -0,0 +1,56 @@
...
+print('Document text: ' . $text . PHP_EOL);
+
+# Print out more detailed and structured information about document text
+foreach ($document->pages() as $page) {
+    foreach ($page['blocks'] as $block) {
+        $block_text = '';
+        foreach ($block['paragraphs'] as $paragraph) {
+            foreach ($paragraph['words'] as $word) {
+                foreach ($word['symbols'] as $symbol) {
+                    $block_text = $block_text . $symbol['text'];
```
bshaffer: Should be `$block_text .= $symbol['text'];`

vision/api/src/snippets/detect_document_text.php
```diff
@@ -0,0 +1,56 @@
...
+                }
```
bshaffer: Should add `$block_text .= ' ';` so there is a space between words.

vision/api/src/snippets/detect_document_text.php
```diff
@@ -0,0 +1,56 @@
...
+            }
```
bshaffer: Should add `$block_text .= "\n";` so there is a newline between paragraphs

vision/api/src/snippets/detect_document_text.php
```diff
@@ -0,0 +1,56 @@
...
+        }
+        print('Block text: ' . $block_text . PHP_EOL);
+        print('Block bounds:' . PHP_EOL);
+        foreach ($block['boundingBox']['vertices'] as $vertice) {
+            print('X: ' . $vertice['x'] . ' Y: ' . $vertice['y'] . PHP_EOL);
```
bshaffer: This would look cleaner as
```php
printf('X: %s Y: %s' . PHP_EOL, $vertice['x'], $vertice['y']);
```

vision/api/src/snippets/detect_document_text.php
ryanmats: fixed

vision/api/src/snippets/detect_document_text.php
ryanmats: fixed

vision/api/src/snippets/detect_document_text.php
ryanmats: fixed

vision/api/src/snippets/detect_document_text.php
ryanmats: fixed

ryanmats:
All the necessary samples / tests are now here. I still need to refactor the *Command.php classes and combine the api and quickstart directories - I'll do that next.

bshaffer:
Fixed the code styles error, so hopefully the tests will pass! If so this is LGTM.

PR #416, opened 2017-08-03
splits word-time-offset samples into their own files

bshaffer:


remi:
👍  LGTM

I cannot speak for the InputDefinition diff changes & some general PHP, but the transcribe sync/async & the transcribe sync/async words look great!

bshaffer:
Thanks! Yeah that's just a minor refactoring I did because I was sticking everything into a single command and wanted to make it more explicit (i.e. the README updates).

PR #419, opened 2017-08-04
Auth code samples

ryanmats:
Tests, README instructions, and Apiary samples coming soon!


auth/appengine/app.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+# [START auth_cloud_explicit_app_engine]
+use Google\Auth\Credentials\AppIdentityCredentials;
```
bshaffer: The sample shows Silex logic, which is not ideal. The sample should be copy/pasteable.

I suggest moving this into a file and create a function, like what we've done with the other samples, such as `function auth_explicit_app_engine`

auth/appengine/app.php
```diff
@@ -0,0 +1,58 @@
...
+use Silex\Application;
+use Symfony\Component\HttpFoundation\Request;
+
+// create the Silex application
+$app = new Application();
+
+$app->get('/', function () use ($app) {
+    # Explicitly use service account credentials by using App Engine credentials.
+    # Learn more about scopes at https://cloud.google.com/storage/docs/authentication#oauth-scopes
+    $projectId = $app['project_id'];
```
bshaffer: FYI The project ID in App Engine will be automatically detected (see [here](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/Core/ClientTrait.php#L167)) as of v0.33.  But since we're explicitly setting things, I suggest passing it into the function as an argument.

auth/appengine/phpunit.xml
```diff
@@ -0,0 +1,31 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<phpunit bootstrap="./vendor/autoload.php">
+    <testsuites>
+        <testsuite name="AppEngine Flex storage test">
```
bshaffer: This is actually an App Engine Standard test

auth/gce/app.yaml
```diff
@@ -0,0 +1,6 @@
+runtime: php
+env: flex
+
+runtime_config:
+  document_root: .
```
bshaffer: GCE doesn't use `app.yaml`, is this sample really App Engine Flex?

auth/src/auth_cloud_explicit.php
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/auth/README.md
+ */
+
+# [START auth_cloud_explicit]
```
bshaffer: I assume this is consistent with other languages, but this would be better named `auth_cloud_explicit_file`

auth/src/auth_cloud_explicit.php
```diff
@@ -0,0 +1,43 @@
...
+
+// Imports the Google Cloud Storage client library.
+use Google\Cloud\Storage\StorageClient;
+
+function auth_cloud_explicit()
+{
+    # Explicitly use service account credentials by specifying the private key
+    # file.
+    $config = [
+        'keyFilePath' => 'service-account.json',
```
bshaffer: Wouldn't this be better if passed in as an argument to the function, $keyFilePath?

If this is how the other languages are doing it then ignore me :)

auth/appengine/app.php
ryanmats: moved the App Engine Flex / Standard apps into functions

auth/appengine/app.php
ryanmats: changed it to accept an argument for projectId

auth/appengine/phpunit.xml
ryanmats: changed to use functions

auth/gce/app.yaml
ryanmats: changed to use functions

auth/src/auth_cloud_explicit.php
ryanmats: using what canonical sample uses

auth/src/auth_cloud_explicit.php
ryanmats: sure, changed it to accept an argument

auth/test/CommandTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+
+namespace Google\Cloud\Samples\Auth;
+
+use Symfony\Component\Console\Tester\CommandTester;
+
+/**
+ * Unit Tests for transcribe commands.
+ */
+class CommandTest extends \PHPUnit_Framework_TestCase
```
bshaffer: nit, but let's move this to `authTest.php` and call this class `authTest`, to be consistent with the other samples.

auth/auth.php
```diff
@@ -0,0 +1,100 @@
...
+
+EOF
+    )
+    ->setCode(function ($input, $output) {
+        auth_cloud_explicit_compute_engine($input->getArgument('projectId'));
+    })
+);
+
+// Create auth-cloud-explicit-app-engine Command.
+$application->add((new Command('auth-cloud-explicit-app-engine'))
```
bshaffer: This is a little odd since it's impossible to actually call this on appengine. You may be able to do something like:

```php
if ($onAppEngine) {
    return auth_cloud_explicit_app_engine(getenv('GCLOUD_PROJECT'));
}
```

It's hacky, so it may not be best.  It would also require an `app.yaml` in this case. 

Another option would be to add something like this to the README:

```md
## Run on App Engine Standard

To run this on App Engine Standard, create a yaml file like this

```yaml
runtime: php55
handlers:
- url: .*
  script: index.php
`` 

And create an index.php file that looks something like this:

```php
require_once __DIR__ . '/vendor/autoload.php';

auth_cloud_explicit_app_engine();
``
```

That way there is at least *some* instruction on how to do it.

auth/test/CommandTest.php
ryanmats: done

auth/auth.php
ryanmats: done

auth/app-standard.yaml
```diff
@@ -0,0 +1,10 @@
+runtime: php55
+api_version: 1
+threadsafe: true
+
+handlers:
+- url: /.*
+  script: index.php
+
+env_variables:
+  GCLOUD_PROJECT: your-project-id
```
bshaffer: Instead of this, try doing the following:

```php
use google\appengine\api\app_identity\AppIdentityService;
$projectId = AppIdentityService::getProjectId();
```

auth/test/authTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $this->assertContains('Invalid Credentials', $output);
+    }
+
+    private function runCommand($commandName, $serviceAccountPath=null, $projectId=null)
+    {
+        $application = require __DIR__ . '/../auth.php';
+        $command = $application->get($commandName);
+        $commandTester = new CommandTester($command);
+
+        ob_start();
```
bshaffer: This whole `runCommand` block is not optimal. All these conditionals for a function that should be making things easier to test. Consider instead passing in the $input as an array to this, so something like this:

```php
private function runCommand($commandName, $projectId = null, $serviceAccountPath)
{
    $args = array_filter([
            'projectId' => $projectId,
            'serviceAccountPath' => $serviceAccountPath,
    ];

    ob_start();
    try {
        $commandTester->execute(
            $args
            ['interactive' => false]
        );
    } catch (\Google\Cloud\Core\Exception\ServiceException $e) {
        ob_clean();
        $application->renderException($e, $commandTester->getOutput());
        return $commandTester->getDisplay();
    }
    return ob_get_clean();
}
```

auth/test/authTest.php
```diff
@@ -0,0 +1,143 @@
...
+        } else if ($projectId) {
+            try {
+                $commandTester->execute(
+                    [
+                        'projectId'=> $projectId
+                    ],
+                    ['interactive' => false]
+                );
+            } catch (\Google\Cloud\Core\Exception\ServiceException $e) {
+                ob_get_clean();
```
bshaffer: use `ob_clean` here, which is the same function but doesn't return the current buffer. This is more appropriate because we're not doing anything with the return value.

auth/src/auth_cloud_explicit_compute_engine.php
```diff
@@ -0,0 +1,44 @@
...
+# [START auth_cloud_explicit_compute_engine]
+namespace Google\Cloud\Samples\Auth;
+
+// Imports GCECredentials and the Google Cloud Storage client library.
+use Google\Auth\Credentials\GCECredentials;
+use Google\Cloud\Storage\StorageClient;
+
+function auth_cloud_explicit_compute_engine($projectId)
+{
+    $gce_credentials = new GCECredentials();
```
bshaffer: To stay consistent with camel vs snake case, let's use `$gceCredentials` here.

auth/src/auth_cloud_explicit_app_engine.php
```diff
@@ -0,0 +1,46 @@
...
+
+// Imports AppIdentityCredentials and the Google Cloud Storage client library.
+use Google\Auth\Credentials\AppIdentityCredentials;
+use Google\Cloud\Storage\StorageClient;
+
+function auth_cloud_explicit_app_engine($projectId)
+{
+    # Learn more about scopes at https://cloud.google.com/storage/docs/authentication#oauth-scopes
+    $scope = 'https://www.googleapis.com/auth/devstorage.read_only';
+    $gae_credentials = new AppIdentityCredentials($scope);
```
bshaffer: change to `$gaeCredentials`

auth/src/auth_api_explicit_compute_engine.php
```diff
@@ -0,0 +1,58 @@
...
+use GuzzleHttp\Client;
+use GuzzleHttp\HandlerStack;
+
+// Imports the Google Cloud Storage client library.
+use Google_Client;
+use Google_Service_Storage;
+
+function auth_api_explicit_compute_engine($projectId)
+{
+    $gce_credentials = new GCECredentials();
```
bshaffer: change to `$gceCredentials`

auth/src/auth_api_explicit_app_engine.php
```diff
@@ -0,0 +1,60 @@
...
+
+// Imports the Google Cloud Storage client library.
+use Google_Client;
+use Google_Service_Storage;
+
+function auth_api_explicit_app_engine($projectId)
+{
+    # Learn more about scopes at https://cloud.google.com/storage/docs/authentication#oauth-scopes
+    $scope = 'https://www.googleapis.com/auth/devstorage.read_only';
+    $gae_credentials = new AppIdentityCredentials($scope);
```
bshaffer: change to `$gaeCredentials`

auth/index.php
```diff
@@ -0,0 +1,43 @@
...
+    printf("Buckets retrieved using the api client:\n");
+    auth_api_explicit_compute_engine(getenv('GCLOUD_PROJECT'));
+} else {
+    printf("Buckets retrieved using the cloud client library:\n");
+    auth_cloud_explicit_app_engine(getenv('GCLOUD_PROJECT'));
+    printf("\n");
+    printf("Buckets retrieved using the api client:\n");
+    auth_api_explicit_app_engine(getenv('GCLOUD_PROJECT'));
+}
+print('</pre>');
```
bshaffer: This is great, but now that the logic is expanding, I think we actually DO want to make this a template.  Try the following:

```php
<?php
// ... 
namespace Google\Cloud\Samples\Auth;

use Google\Auth\Credentials\GCECredentials;
use google\appengine\api\app_identity\AppIdentityService;

require __DIR__ . '/vendor/autoload.php';

$onGce = GCECredentials::onGce();
$projectId = $onGce
   ? getenv('GCLOUD_PROJECT')
   : AppIdentityService::getProjectId();;
?>

<h1>Buckets retrieved using the cloud client library:</h1>
<pre>
<?php if ($onGce): ?>
    <?php auth_cloud_explicit_compute_engine($projectId) ?>
<?php else: ?>
    <?php auth_cloud_explicit_app_engine($projectId) ?>
<?php endif ?>
</pre>

<h1>Buckets retrieved using the api client:</h1>
<pre>
<?php if ($onGce): ?>
    <?php auth_api_explicit_compute_engine($projectId) ?>
<?php else: ?>
    <?php auth_api_explicit_app_engine($projectId) ?>
<?php endif ?>
</pre>
```

auth/composer.json
```diff
@@ -0,0 +1,27 @@
+{
+    "require": {
+        "google/apiclient": "^2.1",
+        "google/cloud-storage": "^1.0",
+        "paragonie/random_compat": "^2.0",
```
bshaffer: Why is this explicitly set?

auth/composer.json
```diff
@@ -0,0 +1,27 @@
...
+        "symfony/console": " ^3.0",
+        "google/auth":"^0.11"
```
bshaffer: This should be `google/auth:^1.0`

auth/app-standard.yaml
ryanmats: done

auth/test/authTest.php
ryanmats: done

auth/src/auth_cloud_explicit_compute_engine.php
ryanmats: done

auth/src/auth_cloud_explicit_app_engine.php
ryanmats: done

auth/src/auth_api_explicit_compute_engine.php
ryanmats: done

auth/src/auth_api_explicit_app_engine.php
ryanmats: done

auth/composer.json
ryanmats: removed

auth/composer.json
ryanmats: done

auth/test/authTest.php
ryanmats: ob_clean results in a risky test. keeping the same.

auth/test/authTest.php
ryanmats: done. but i did need an extra if/else statement to create $args differently based on whether or not there is a serviceAccountPath.

ryanmats: done. but i did need an extra if/else to create $args differently depending on whether or not we have a serviceAccountPath.

auth/test/authTest.php
bshaffer: Really? What do you mean?

auth/test/authTest.php
bshaffer: No, the `array_filter` will filter out null arguments.

bshaffer: Another option would be to make the function declaration like this:

```php
private function runCommand($commandName, $args)
```

And have the tests pass in the args themselves, i.e.

```
$this->runCommand('app-engine-explicit', [
    'projectId' => $this->projectId,
    'serviceAccountPath' => $serviceAccountPath
]);
```

But in general `runCommand` should be simple and not contain any conditionals.

auth/index.php
ryanmats: done

auth/test/authTest.php
ryanmats: never mind, looks like it was an unrelated issue before. changed it back to the shorter version.

auth/test/authTest.php
```diff
@@ -0,0 +1,119 @@
...
+        $serviceAccountPath = getenv('GOOGLE_APPLICATION_CREDENTIALS');
+        $output = $this->runCommand('auth-cloud-explicit', $this->projectId, $serviceAccountPath);
+        $this->assertContains($this->bucketName, $output);
+    }
+
+    public function testAuthCloudExplicitComputeEngineCommand()
+    {
+        $output = $this->runCommand(
+            'auth-cloud-explicit-compute-engine', $this->projectId);
+        $this->assertContains('Undefined index: access_token', $output);
```
bshaffer: This is a PHP warning, I don't think we want to assert this

bshaffer:
Let's add two e2e tests for GCE and GAE (standard).  Name them `StandardDeployTest.php` and `FlexDeployTest.php`, and then add this to each of them (after using the `AppEngineDeploymentTestTrait` in each of them, see ([this test](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/standard/memcache/test/DeployTest.php) for standard, and [this test](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/flexible/analytics/test/DeployTest.php) for flex)

```php
    // for FlexDeployTest.php
    private static function doDeploy()
    {
        return self::$gcloudWrapper->deploy('app-flex.yaml');
    }
    // for StandardDeployTest.php
    private static function doDeploy()
    {
        return self::$gcloudWrapper->deploy('app-standard.yaml');
    }
```

bshaffer:
Added #424 for e2e tests

PR #422, opened 2017-08-08
set grpc 1.4.1 in travis

bshaffer:
see https://github.com/grpc/grpc/issues/12109


tmatsuo:
FYI, there is no `1.4.2` ever released. You may want to use `1.4.1`
https://pecl.php.net/package/gRPC

PR #427, opened 2017-08-13
Updates a comment

shawnbayer:
Updates a comment to say "export" instead of "import" on the export table function for biquery.


shawnbayer:
Signed cla

PR #431, opened 2017-08-16
Updated existing Language samples and added entity_sentiment samples

ryanmats:


language/api/composer.json
```diff
@@ -3,7 +3,11 @@
         "google/cloud-language": "^0.4",
         "symfony/console": "^3.0",
         "google/cloud-storage": "^1.0",
-        "symfony/yaml": "^3.2"
+        "symfony/yaml": "^3.2",
```
bshaffer: We can remove this, since you've gotten rid of the `Yaml` calls below.

language/api/composer.json
```diff
@@ -3,7 +3,11 @@
...
+        "google/gax": "^0.21.2",
+        "grpc/grpc": "^v1.1.0",
+        "google/protobuf": "^v3.3.0",
```
bshaffer: The dependency `google/gax` is managed by [google/proto-client](https://github.com/googleapis/proto-client-php/blob/master/composer.json), and the dependencies `grpc/grpc` and `google/protobuf` are maintained by [google/gax](https://github.com/googleapis/gax-php/blob/master/composer.json). So, we can remove these three dependencies. The only new one we need is `google/proto-client`.

language/api/src/analyze_entities.php
```diff
@@ -36,16 +35,37 @@
...
+    // Print out information about each entity
+    $entities = $annotation->entities();
+    foreach ($entities as $entity) {
+        printf('Name: %s' . PHP_EOL, $entity['name']);
+        printf('Type: %s' . PHP_EOL, $entity['type']);
+        printf('Salience: %s' . PHP_EOL, $entity['salience']);
+        if (array_key_exists('wikipedia_url', $entity['metadata'])) {
+            printf('Wikipedia URL: %s' . PHP_EOL, $entity['metadata']['wikipedia_url']);
+        }
+        if (array_key_exists('mid', $entity['metadata'])) {
```
bshaffer: We don't need to print every single possible property - maybe we should exclude this one? What do the other languages do? As a base line, for simplicity, we could leave out all "optional" properties.

language/api/src/analyze_entity_sentiment_from_file.php
```diff
@@ -0,0 +1,80 @@
...
+
+    // Print out information about each entity
+    foreach ($entities as $entity) {
+        printf('Entity Name: %s' . PHP_EOL, $entity->getName());
+        printf('Entity Type: %s' . PHP_EOL, $entity_types[$entity->getType()]);
+        printf('Entity Salience: %s' . PHP_EOL, $entity->getSalience());
+        printf('Entity Magnitude: %s' . PHP_EOL, $entity->getSentiment()->getMagnitude());
+        printf('Entity Score: %s' . PHP_EOL, $entity->getSentiment()->getScore());
+        printf('Mentions: ' . PHP_EOL);
+        foreach ($entity->getMentions() as $mention) {
```
bshaffer: dumb question - what is a mention? For simplicity, can we exclude printing them out every time, or do the other languages do this also?

bshaffer:
You will need to run `composer update` on PHP 5.6 in order to get rid of the dependency error on Travis. Once the tests pass this is good 2 go 👍 👍

ryanmats:
thanks Brent. Yeah - I noticed the Travis dependency error. Running into a small roadblock reverting back to PHP 5.6 with phpenv but hopefully I'll be able to fix it soon!

PR #432, opened 2017-08-17
updating Mailjet code samples for Send API v3.1

adamyanliev:
Mailjet Send API v3.1 is being released, so updating the code samples to match the new Send API v3.1 configuration. Reference can be found [here](https://dev.mailjet.com/beta/).

### DIFF
--- a/appengine/flexible/mailjet/app.php
+++ b/appengine/flexible/mailjet/app.php
@@ -40,23 +40,27 @@ $app->post('/send', function () use ($app) {
     /** @var Mailjet\Client $mailjet */
     $mailjet = $app['mailjet'];
     $recipient = $request->get('recipient');
-
     # [START send_email]
     $body = [
-        'FromEmail' => "test@example.com",
-        'FromName' => "Testing Mailjet",
-        'Subject' => "Your email flight plan!",
-        'Text-part' => "Dear passenger, welcome to Mailjet! May the delivery force be with you!",
-        'Html-part' => "<h3>Dear passenger, welcome to Mailjet!</h3><br/>May the delivery force be with you!",
-        'Recipients' => [
-            [
-                'Email' => $recipient,
-            ]
+    'Messages' => [
+        [
+            'From' => [
+                'Email' => "pilot@mailjet.com",
+                'Name' => "Mailjet Pilot"
+            ],
+            'To' => [
+                [
+                    'Email' => $recipient
+                ]
+            ],
+            'Subject' => "Your email flight plan!",
+            'TextPart' => "Dear passenger, welcome to Mailjet! May the delivery force be with you!",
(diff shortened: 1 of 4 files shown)


appengine/standard/mailjet/app.php
```diff
@@ -40,31 +40,34 @@
...
-        'FromEmail' => "test@example.com",
-        'FromName' => "Testing Mailjet",
-        'Subject' => "Your email flight plan!",
-        'Text-part' => "Dear passenger, welcome to Mailjet! May the delivery force be with you!",
-        'Html-part' => "<h3>Dear passenger, welcome to Mailjet!</h3><br/>May the delivery force be with you!",
-        'Recipients' => [
-            [
-                'Email' => $recipient,
-            ]
+    'Messages' => [
```
bshaffer: This is the wrong indentation, everything after body needs another indent.

appengine/flexible/mailjet/app.php
```diff
@@ -40,31 +40,34 @@
...
-        'FromEmail' => "test@example.com",
-        'FromName' => "Testing Mailjet",
-        'Subject' => "Your email flight plan!",
-        'Text-part' => "Dear passenger, welcome to Mailjet! May the delivery force be with you!",
-        'Html-part' => "<h3>Dear passenger, welcome to Mailjet!</h3><br/>May the delivery force be with you!",
-        'Recipients' => [
-            [
-                'Email' => $recipient,
-            ]
+    'Messages' => [
```
bshaffer: Same here

adamyanliev:
hi @bshaffer, thanks for the comments. I fixed the indentation, it should be ok now. Also, yes, the samples were manually deployed, tested and working.

adamyanliev:
@bshaffer the Travis CI build fails now, seems like the PHP CS Fixer needs to be updated on your end?

bshaffer:
@adamyanliev working on it... https://github.com/GoogleCloudPlatform/php-tools/pull/21

bshaffer:
Before I merge this can you mention the reasons for opting into Send API v3?

bshaffer:
It appears the client library does not default to v3 (we explicitly set the version), so we are opting into V3. Why?

adamyanliev:
@bshaffer v3 is indeed the default - see the below: 
https://github.com/mailjet/mailjet-apiv3-php/blob/master/src/Mailjet/Client.php#L35
https://github.com/mailjet/mailjet-apiv3-php/blob/master/src/Mailjet/Config.php#L21

bshaffer:
If it's the default why are we specifying `['version' => 'v3']` when making the call?

adamyanliev:
@bshaffer sorry, our mistake, version call should be 3.1. I updated the files to match, thank you for pointing it out.

bshaffer:
This is now returning 400 Bad Request. When I revert this PR, it works as expected.

PR #434, opened 2017-08-18
Read stale data sample for spanner.

SurferJeffAtGoogle:


spanner/src/read_stale_data.php
```diff
@@ -0,0 +1,62 @@
...
+    $instance = $spanner->instance($instanceId);
+    $database = $instance->database($databaseId);
+    // maxStaleness and minReadTimestamp are only available in single-use transactions.
+    $snapshot = $database->snapshot();
+    $keySet = $spanner->keySet(['all' => true]);
+    $results = $snapshot->read(
+        'Albums',
+        $keySet,
+        ['SingerId', 'AlbumId', 'AlbumTitle'],
+        ['maxStaleness' => new Duration(3)]
```
vkedia: can you use exact staleness with 10 seconds. We recommend in our docs to use atleast 10 seconds to get maximum performance benefits. And exact staleness in general would be faster than maxStaleness.

SurferJeffAtGoogle: Done.

SurferJeffAtGoogle:
Brent, what error do you see?

bshaffer:
@SurferJeffAtGoogle I'm talking about the error being thrown in travis

PR #436, opened 2017-08-22
KMS: Make samples consistent across languages.

white0ut:
This change:

* Uses a consistent "id" suffix for resource name components.
* Uses ciphertext and plaintext instead of inFile and outFile.
* Correctly handles base64 decoding and encoding of responses.
	* Data written to files should never be base64 encoded.

Tracking bug: http://b/64758639

### DIFF
--- a/kms/api/src/functions.php
+++ b/kms/api/src/functions.php
@@ -35,16 +35,16 @@ use Google_Service_CloudKMS_UpdateCryptoKeyPrimaryVersionRequest;
  * Add a member to a CryptoKey IAM policy.
  *
  * @param string $projectId
- * @param string $ring
- * @param string $key
+ * @param string $keyRingId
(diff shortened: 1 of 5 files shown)


bshaffer:
None of the tests have been updated to accommodate for these changes.

white0ut:
Sorry about that. Tests updated

white0ut:
Brent, I'm seeing failures related to ADC, is that because this is a fork?

```
Unable to read the credential file specified by  GOOGLE_APPLICATION_CREDENTIALS: file /home/travis/build/GoogleCloudPlatform/php-docs-samples/credentials.json does not exist
```

bshaffer:
Yes, unfortunately these need to run on the same branch in order for travis to use the credentials.

I ran the tests locally and have found a lot of errors. I'll push a separate branch and fix the issues.

PR #438, opened 2017-09-05
Speech: Print out all results.

jerjou:
There was some confusion where folks thought the API was only transcribing the first portion of the audio, when in fact it was because they were only using the first result. I'm updating the samples to try to reduce that confusion. See b/64934785

### DIFF
--- a/speech/quickstart.php
+++ b/speech/quickstart.php
@@ -43,8 +43,8 @@ $options = [
 # Detects speech in the audio file
 $results = $speech->recognize(fopen($fileName, 'r'), $options);
 
-foreach ($results[0]->alternatives() as $alternative) {
-    echo 'Transcription: ' . $alternative['transcript'] . PHP_EOL;
+foreach ($results as $result) {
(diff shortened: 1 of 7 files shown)


jerjou:
Test failures look unrelated to this PR, and I'm having trouble running them manually (it skips 3/9, and tells me to `pecl install grpc`, which I did but it didn't help)

bshaffer:
Thanks for the feedback, you also will need to enable grpc by adding `extension=grpc.so` to your `php.ini` file (which you can get by running `php -i | grep ini`). We will make the error message better!

jerjou:
Ah - that fixed it. Okay - confirmed that all the tests pass.

jerjou:
@bshaffer  ping? I've updated all the other language samples, and just waiting on this one so I can refresh the doc pages.

jerjou:
Yay! Would you mind merging? I don't appear to have permissions to..

tmatsuo:
Thanks @jerjou !

PR #443, opened 2017-09-07
WordPress on App Engine README Fixes + Improvements (ready for review + merge)

ryanmats:
I plan on compiling all updates here in this pull request and then testing everything end-to-end to ensure that it everything works as expected.


bshaffer:
Assign to me when it's ready for me to have a look!

ryanmats:
This GitHub tutorial now works end-to-end, with the exception of local testing. There are a few issues there, one of which may require fixing something with the Google App Engine plugin, so let's try to merge this one first and then add the local testing instructions later once they work.

bshaffer:
As we discussed, let's get an issue open for local testing!

appengine/wordpress/src/WordPressSetup.php
```diff
@@ -29,7 +29,7 @@ class WordPressSetup extends Command
 {
     const DEFAULT_DIR = 'my-wordpress-project';
     const DEFAULT_ERROR = 1;
-    const LATEST_WP = 'https://wordpress.org/latest.tar.gz';
+    const LATEST_WP = 'https://wordpress.org/4.8.1.tar.gz';
```
tmatsuo: Why `4.8.1` is better than `latest`?

jgacuca567: @tmatsuo @bshaffer It should stay as latest, so that you don't keep changing the constant every time 
`4.8.2` was just released

ryanmats: There seems to be a temporary problem with using latest (or at least there was when I tried yesterday): https://github.com/wp-cli/wp-cli/issues/4370

I was thinking we could leave it at 4.8.1 and then switch it back to latest once the WordPress issue has been resolved. Thoughts?

ryanmats: @tmatsuo And is it okay to leave this version until the 'latest' issue is fixed? And does this look good otherwise?

ryanmats:
Opened the issue for local testing:  https://github.com/GoogleCloudPlatform/php-docs-samples/issues/454

PR #447, opened 2017-09-15
Storage requester pays sample code + tests

ryanmats:
Code samples + tests are complete and ready for review.

HOWEVER, we should not merge this until a bug in the client library is resolved (https://github.com/GoogleCloudPlatform/google-cloud-php/issues/667). The bug prevents the download function from working properly when using different accounts for the Storage owner and the requester.

### DIFF
--- a/storage/api/src/functions/disable_requester_pays.php
+++ b/storage/api/src/functions/disable_requester_pays.php
@@ -42,7 +42,7 @@ function disable_requester_pays($projectId, $bucketName)
     $bucket = $storage->bucket($bucketName);
     $bucket->update([
         'billing' => [
-            'requesterPays' => False
+            'requesterPays' => false
         ]
     ]);
     printf('Requester pays has been disabled for %s' . PHP_EOL, $bucketName);
--- a/storage/api/src/functions/disable_requester_pays.php
+++ b/storage/api/src/functions/disable_requester_pays.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/storage/api/README.md
+ */
+
+namespace Google\Cloud\Samples\Storage;
+
+# [START disable_requester_pays]
+use Google\Cloud\Storage\StorageClient;
+
+/**
+ * Disable a bucket's requesterpays metadata.
+ *
+ * @param string $projectId Your Google Cloud project ID.
+ * @param string $bucketName Name of your Google Cloud Storage bucket.
+ *
+ * @return void
+ */
+function disable_requester_pays($projectId, $bucketName)
+{
+    $storage = new StorageClient([
+        'projectId' => $projectId
+    ]);
+    $bucket = $storage->bucket($bucketName);
+    $bucket->update([
+        'billing' => [
+            'requesterPays' => False
+        ]
+    ]);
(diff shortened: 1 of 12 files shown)


storage/api/src/functions/disable_requester_pays.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+function disable_requester_pays($projectId, $bucketName)
+{
+    $storage = new StorageClient([
+        'projectId' => $projectId
+    ]);
+    $bucket = $storage->bucket($bucketName);
+    $bucket->update([
+        'billing' => [
+            'requesterPays' => False
```
bshaffer: `False` should be lowercase

storage/api/src/functions/enable_requester_pays.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+function enable_requester_pays($projectId, $bucketName)
+{
+    $storage = new StorageClient([
+        'projectId' => $projectId
+    ]);
+    $bucket = $storage->bucket($bucketName);
+    $bucket->update([
+        'billing' => [
+            'requesterPays' => True
```
bshaffer: again here

storage/api/test/RequesterPaysCommandTest.php
```diff
@@ -0,0 +1,152 @@
...
+
+use Google\Cloud\Samples\Storage\RequesterPaysCommand;
+use Google\Cloud\Storage\StorageClient;
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Tester\CommandTester;
+
+/**
+ * Unit Tests for RequesterPaysCommand.
+ */
+class BucketsCommandTest extends \PHPUnit_Framework_TestCase
```
bshaffer: should be `RequesterPaysCommandTest`

storage/api/test/RequesterPaysCommandTest.php
```diff
@@ -0,0 +1,152 @@
...
+                'bucket' => $bucketName,
+                '--enable' => true,
+            ],
+            ['interactive' => false]
+        );
+
+        $this->expectOutputRegex("/Requester pays has been enabled/");
+    }
+
+    public function testDisableRequesterPays()
```
bshaffer: This should have `/** @depends testEnableRequesterPays */` so it only runs if the above test passes

storage/api/test/RequesterPaysCommandTest.php
```diff
@@ -0,0 +1,152 @@
...
+                'bucket' => $bucketName,
+                '--disable' => true,
+            ],
+            ['interactive' => false]
+        );
+
+        $this->expectOutputRegex("/Requester pays has been disabled/");
+    }
+
+    public function testGetRequesterPaysStatus()
```
bshaffer: This should have `/** @depends testDisableRequesterPays */` so it only runs if the above test passes

storage/api/test/RequesterPaysCommandTest.php
```diff
@@ -0,0 +1,152 @@
...
+                'project' => $projectId,
+                'bucket' => $bucketName,
+                'object' => $objectName,
+                'download-to' => $destination
+            ],
+            ['interactive' => false]
+        );
+
+        $this->expectOutputRegex("/Downloaded/");
+        $this->expectOutputRegex("/using requester-pays requests/");
```
bshaffer: The `expectOutputRegex` function only uses the last value its called with, so `"/Downloaded/"` should be removed.

storage/api/src/functions/download_file_requester_pays.php
```diff
@@ -0,0 +1,50 @@
...
+ * @param string $destination Path to local file to save.
+ *
+ * @return void
+ */
+function download_file_requester_pays($projectId, $bucketName, $objectName, $destination)
+{
+    $storage = new StorageClient([
+        'projectId' => $projectId
+    ]);
+    $bucket = $storage->bucket($bucketName, true);
```
bshaffer: We should make this example more obvious as to how this applies to Requester Pays, like this:

```php
$storage = new StorageClient([
    'projectId' => $projectId
]);
$requesterPays = true;
$bucket = $storage->bucket($bucketName, $requesterPays);
```

bshaffer:
Hmm, the 5.6 build errored with "Your requirements could not be resolved to an installable set of packages". I restarted the build because I've never seen this error before. 🤞

frankyn:
Pending open issue. 

Sample design LGTM!

ryanmats:
It could be an issue where I've been testing everything using PHP version 7.1 on my computer, and that updates the composer files to use something that is unusable in version 5.6.

If this build fails again I can try moving to 5.6 and push any updates to the composer files.

PR #452, opened 2017-09-21
update flex to use post-install-cmd

bshaffer:
See https://github.com/GoogleCloudPlatform/community/issues/248


bshaffer:
I ran all 3 e2e tests and they succeeded! Woohoo.

PR #453, opened 2017-09-23
Simplifies Logging Samples and adds PSR logging example

bshaffer:


bshaffer:
The tests are failing because of https://github.com/GoogleCloudPlatform/google-cloud-php/pull/684

PR #455, opened 2017-09-26
Error Reporting Samples

bshaffer:
Here is a possible Error Reporting sample, although I'm not sure our Client Libraries are ready for it. I am also debating using `google/cloud-logging` instead, since this can accomplish the same task in a more user friendly way.


bshaffer:
After looking into it, the integration with Logging is quite minimal, and seems to be very hackish. The error message **must** be prefixed with `PHP Warning: ` and it MUST be an exception to string, e.g.:

```php
$exceptionToString = (string) (new Exception('here is my message');
$logMessage = sprintf('PHP Warning: %s', $exceptionToString);
```

There is also no way to log additional context, such as the user, and there are other restrictions such as a requirement for **module** and **version**. Because these rules don't really make sense in many error reporting cases, having a dedicated error reporting sample will be a good investment.

bshaffer:
@tmatsuo I think this is a great idea! We can follow [Node.js's example](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/error-reporting/snippets.js#L47)

@dwsupplee is this something we can do by submitting PRs to `ReportErrorsServiceClient`?

dwsupplee:
@bshaffer Definitely! Those classes exist solely for us to be able to include utility code outside of the autogen GAPIC layer. We've done something similar with [Spanner](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/Spanner/V1/SpannerClient.php#L40-L65).

bshaffer:
@dwsupplee Why are the stubs and credentials helpers being exposed? Those aren't the kinds of functions I had in mind.

dwsupplee:
Those are unique to Spanner as we needed to expose them in order to fulfill some product requirements. I was only pointing to them as an example of a place we have exposed added functionality before.

bshaffer:
@dwsupplee I am thinking of helper functions like `$logger->logException()`. The logging and error reporting APIs are essentially identical, with the Error Reporting API being a derivative of the Logging API. I am not sure there's a really good reason to recommend using both, especially if we can build one into the other.

ryanmats:
looks good with the exception of the linting issues

PR #458, opened 2017-10-09
Fix link to latest WordPress version

bionicbrian:
As it is, without this fix, WordPress setup is broken as the link is not valid without the `wordpress-` bit. Updated to the latest patch because might as well. Thanks!


bionicbrian:
I signed it!

bshaffer:
Great catch, thank you @bionicbrian!

tmatsuo:
Can we revert it back to `https://wordpress.org/latest.tar.gz`?
@ryanmats WDYT?

ryanmats:
@tmatsuo i'll run through everything again soon to see if the issue with 'latest' has been resolved!

PR #484, opened 2017-11-17
Fixes the Endpoints sample.

wmwong:
The app.yaml file now requires a document_root to be specified.

### DIFF
--- a/endpoints/getting-started/app.yaml
+++ b/endpoints/getting-started/app.yaml
@@ -1,6 +1,9 @@
 runtime: php
 env: flex
 
+runtime_config:
+  document_root: web
+
 endpoints_api_service:
   # The following values are to be replaced by information from the output of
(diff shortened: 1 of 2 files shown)


bshaffer:
Moving index.php into `web` will break the sample, why do you think this change is necessary?

See https://github.com/GoogleCloudPlatform/php-docs-samples/pull/483, where we make document root the root directory `.`

wmwong:
I was looking at this documentation:
https://cloud.google.com/appengine/docs/flexible/php/configuring-your-app-with-app-yaml#general

And this note specifically:
Note: Although you can specify the project root directory as the document root, we recommend that you use a subdirectory for the document_root setting.

I was able to deploy and hit the endpoint with this particular fix (with index.php under web/). However, it looks like you've already fixed the issue.

Issue #493, opened 2017-11-28
Add "Open In Cloudshell" to samples


bshaffer:
Add the button "Open in CloudShell" to applicable samples, namely API samples.

Example: [NodeJS Spanner samples](https://github.com/googleapis/nodejs-spanner/tree/master/samples)

The URL should look like this (for spanner):

https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googlecloudplatform/php-docs-samples&page=editor&working_dir=spanner&open_in_editor=README.md

bshaffer:
Add this to the top of every applicable `README.md`:

```
[![Open in Cloud Shell][shell_img]][shell_link]

[shell_img]: http://gstatic.com/cloudssh/images/open-btn.png
[shell_link]: https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googlecloudplatform/php-docs-samples&page=editor&working_dir=DIR&open_in_editor=README.md
```

PR #496, opened 2017-11-29
adds Open In Cloudshell to all READMEs with one depth

bshaffer:
#493

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -3,7 +3,7 @@
 [![Open in Cloud Shell][shell_img]][shell_link]
 
 [shell_img]: http://gstatic.com/cloudssh/images/open-btn.png
-[shell_link]: https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googlecloudplatform/php-docs-samples&page=editor&working_dir=auth&open_in_editor=README.md
+[shell_link]: https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googlecloudplatform/php-docs-samples&page=editor&working_dir=auth
 
 ## Description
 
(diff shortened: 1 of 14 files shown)


auth/README.md
```diff
@@ -1,5 +1,10 @@
 # Google Auth PHP Sample Application
 
+[![Open in Cloud Shell][shell_img]][shell_link]
+
+[shell_img]: http://gstatic.com/cloudssh/images/open-btn.png
+[shell_link]: https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googlecloudplatform/php-docs-samples&page=editor&working_dir=auth&open_in_editor=README.md
```
tmatsuo: I tried the link. There are basically 4 panes; file list, editor, instruction pane on the right, the shell at the bottom. I have feedback:

1. The editor and the instruction pane both shows the same file, but with the different format. I feel it's redundant.
2. The README is not perfect for this use case, the step for cloning the repo is redundant
3. Cloudshell doesn't have composer

Can we

1. Create a specialized readme file for this purpose and show it?
2. Can we install composer by default in the cloudshell environment?

bshaffer: 1. If we remove "open_in_editor", it displays just the rendered README as expected, which I think should suffice. I'll get to updating that.

2. Agreed that we need `composer` in the cloud shell environment, and that it is a requirement for this PR. I'll get started on this.

bshaffer:
This PR is dependent on composer being added to CloudShell, and CloudShell being upgraded to PHP 7.0.

bshaffer:
Composer has been added to Cloud Shell and PHP has been upgraded to 7.0!

PR #498, opened 2017-11-30
adds php 72 to travis

bshaffer:


bshaffer:
This is erroring because of issues with the DLP API. We are working to resolve them.

tmatsuo:
Adding PHP7.2 in #530 
closing

PR #510, opened 2017-12-08
Cloud Tasks Code Samples - work in progress

ryanmats:
Currently have the App Engine Queue create_task function done. Still need to add a PHP web app to log the payload when it's retrieved from the queue, do the Pull Queue samples, and add tests/README. However, the create_task function does work with a deployed Python app.

Canonical Python samples:

App Engine queue sample: 
https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/appengine/flexible/tasks

Pull queue sample:
https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/tasks


ryanmats:
@bshaffer These Cloud Tasks code samples are ready for review - all functionality is done and all tests are passing on Travis. I just need to add two README instruction files.

appengine/flexible/tasks/index.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Google\Cloud\Logging\LoggingClient;
+
+$app = new Silex\Application();
+
+$app->get('/', function () use ($app) {
+    return 'Hello World';
+});
+
+$app->post('/log_payload', function (Request $request) use ($app) {
```
steren: Please replace "log_payload" with "example_task_handler" in this sample. 
This is a feedback coming from user studies. The Python master samples are changing: https://github.com/GoogleCloudPlatform/python-docs-samples/pull/1288/files

ryanmats: thanks, makes sense.

appengine/flexible/tasks/src/create_task.php
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+# [START create_task]
+namespace Google\Cloud\Samples\Tasks;
+
+use Google_Client;
+use Google_Service_CloudTasks;
+use Google_Service_CloudTasks_AppEngineHttpRequest;
+use Google_Service_CloudTasks_Task;
+use Google_Service_CloudTasks_CreateTaskRequest;
```
bshaffer: Let's move the `# [START create_task]` region tag below the use statements. The users don't need to copy-paste our namespacing, and since the Apiary client library isn't namespaced, there's no reason to include the import statements. 

This is similar to what we do for the [KMS samples](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/kms/src/functions.php#L33)

This should be done for all the samples!

appengine/flexible/tasks/src/create_task.php
```diff
@@ -0,0 +1,92 @@
...
+    $appEngineHttpRequest->setRelativeUrl('/log_payload');
+
+    // Create a Cloud Task object.
+    $task = new Google_Service_CloudTasks_Task();
+    $task->setAppEngineHttpRequest($appEngineHttpRequest);
+
+    // If in_seconds variable is set, set the future time for when the task will be attempted.
+    if ($in_seconds != null) {
+        $seconds_string = sprintf('+%s seconds', $in_seconds);
+        $future_time = date(\DateTime::RFC3339, strtotime($seconds_string));
```
bshaffer: Let's also add `DateTime` to the use statements above the region tag. This way, when the sample is copied and pasted it will still work as expected.

tasks/test/tasksTest.php
```diff
@@ -0,0 +1,95 @@
...
+    public function setUp()
+    {
+        if (!self::$hasCredentials) {
+            $this->markTestSkipped('No application credentials were found.');
+        } elseif (!self::$project) {
+            $this->markTestSkipped('No project ID was found.');
+        } elseif (!self::$queue) {
+            $this->markTestSkipped('No Pull Queue was found.');
+        } elseif (!self::$location) {
+            $this->markTestSkipped('No location was found.');
```
bshaffer: These would be better if they included the name of the env var in the skipped message:

```
if (!self::$location) {
    $this->markTestSkipped('No location was found. Please set the CLOUD_TASKS_LOCATION environment variable.');
}
```

tasks/test/tasksTest.php
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    public function testCreateTask()
+    {
+        $output = $this->runCommand('create-task', ['project' => self::$project, 'queue' => self::$queue, 'location' => self::$location]);
```
bshaffer: nit, but for readability let's make this:

```php
$this->runCommand('create-task', [
    'project' => self::$project,
    'queue' => self::$queue,
    'location' => self::$location
]);
```

tasks/src/create_task.php
```diff
@@ -0,0 +1,81 @@
...
+use Google_Service_CloudTasks_CreateTaskRequest;
+
+/**
+ * Create a task for a given App Engine queue
+ * ```
+ * create_task(PROJECT_ID, QUEUE_ID, LOCATION, PAYLOAD, SECONDS)
+ * ```
+ *
+ * @param string $projectId Project of the queue to add the task to.
+ * @param string $queueID ID (short name) of the queue to add the task to.
```
bshaffer: Is there a method to get or create the queues? If so I think we should add them here, or document the `gcloud` command used to get them. If I am browsing these samples, not having the queue ID is a blocker and I'd have to dig around for it.

tasks/src/pull_task.php
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Pull a task from a given Pull Queue
+ * ```
+ * pull_task(PROJECT_ID, QUEUE_ID, LOCATION)
+ * ```
+ *
+ * @param string $projectId Project of the queue to pull the task from.
+ * @param string $queueID ID (short name) of the queue to pull the task from.
+ * @param string $location Location of the queue to pull the task from.
```
bshaffer: it might make sense to put location before queueId in these samples, since that's the order they are in the URL. Another option is to make them default to `global`, in which case they should be last, so I don't know. At the very least we should document where to see possible locations, and make `global` the default in `tasks.php`.

appengine/flexible/tasks/index.php
ryanmats: done

appengine/flexible/tasks/index.php
```diff
@@ -0,0 +1,50 @@
...
+    $content = $request->getContent();
+    $logging = new LoggingClient();
+    $logName = 'my-log';
+    $logger = $logging->logger($logName);
+    $logging_text = sprintf('Received task with payload: %s', $content);
```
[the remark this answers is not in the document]
ryanmats: done

appengine/flexible/tasks/index.php
```diff
@@ -0,0 +1,50 @@
...
+    $entry = $logger->entry($logging_text);
+    $logger->write($entry);
+    return sprintf('Received task with payload: %s', $content);
```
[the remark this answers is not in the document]
ryanmats: done

appengine/flexible/tasks/src/create_task.php
ryanmats: done

appengine/flexible/tasks/src/create_task.php
ryanmats: done

tasks/src/create_task.php
ryanmats: Added README instructions for both samples.

tasks/src/pull_task.php
ryanmats: Added README instructions to clarify.

tasks/test/tasksTest.php
ryanmats: done

tasks/test/tasksTest.php
ryanmats: done

appengine/flexible/tasks/composer.json
```diff
@@ -0,0 +1,21 @@
+{
+    "require": {
+        "symfony/console": "^3.0",
+        "silex/silex": "^1.3",
+        "google/apiclient": "^2.1",
+        "google/cloud-logging": "^1.5"
+    },
+    "require-dev": {
+        "phpunit/phpunit": "~4.0",
```
tmatsuo: Please don't use phpunit4 any more

appengine/flexible/tasks/test/LocalTest.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+use Silex\WebTestCase;
+
+class LocalTest extends WebTestCase
```
tmatsuo: WebTestCase depends on the old style PHPUnit classes, but you can use the Shim to avoid the error
An example:
https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/standard/cloudsql/test/bootstrap.php

tasks/test/tasksTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+namespace Google\Cloud\Samples\Tasks\Tests;
+
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Tester\CommandTester;
+
+/**
+ * Unit Tests for tasks commands.
+ */
+class tasksTest extends \PHPUnit_Framework_TestCase
```
tmatsuo: Let's use `PHPUnit\Framework\TestCase`

appengine/flexible/tasks/composer.json
ryanmats: Which phpunit version do we want our samples to use? I think most of them are using version 4 at the moment.

tmatsuo: I would suggest just removing it from the `require-dev` section. Travis provides appropriate version

tmatsuo:
Thanks @ryanmats LGTM for the test change

tasks/src/acknowledge_task.php
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Acknowledge a task from a given Pull Queue
+ * ```
+ * acknowledge_task(TASK)
+ * ```
+ *
+ * @param Google_Service_CloudTasks_Task $task A task that was pulled from a Pull Queue.
+ *
+ */
+function acknowledge_task($task)
```
bshaffer: Let's add a typehint here,

```php
function acknowledge_task(Google_Service_CloudTasks_Task  $task)
```

PR #515, opened 2017-12-14
BigQuery: rewrite simple app tutorial.

tswast:
- Use latest version of `google/cloud-bigquery` package.
- Add additional region tags to support tutorial rewrite.
- Use more relevant query from public datasets.
- Use more direct associative array for printing results.


tmatsuo:
The test for this PR skipped the test for bigquery/stackoverflow:

```
+grep -q '^bigquery/stackoverflow'
+echo 'Skipping tests in bigquery/stackoverflow\n'
Skipping tests in bigquery/stackoverflow\n
+continue
+read DIR
```

I encounter the error when I try to fix the test failure in https://github.com/GoogleCloudPlatform/php-docs-samples/pull/530

tswast:
Any guesses as to why that would be? I don't think I did anything special that it would skip it.

Also, note that this sample is already a few versions out of date. :-(

tmatsuo:
Yeah it's not your fault, but I think the script for skipping the irrelevant tests have a bug :(
I'm updating the client lib versions in https://github.com/GoogleCloudPlatform/php-docs-samples/pull/530

PR #516, opened 2017-12-14
BigQuery: update quickstart to latest package versions.

tswast:
Version 0.3.1 is the latest version of the google/cloud-bigquery package.

It includes some breaking changes, so the api samples will need to be updated separately.

### DIFF
--- a/bigquery/quickstart/composer.json
+++ b/bigquery/quickstart/composer.json
@@ -1,7 +1,7 @@
 {
     "require": {
         "php": ">=5.4",
-        "google/cloud-bigquery": "^0.2"
+        "google/cloud-bigquery": "^0.3"
     },
     "require-dev": {
         "phpunit/phpunit": "~4"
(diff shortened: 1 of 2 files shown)


tswast:
https://github.com/GoogleCloudPlatform/php-docs-samples/pull/515 updates the "shakespeare" sample.

I started branch https://github.com/GoogleCloudPlatform/php-docs-samples/tree/bq-pkgs/bigquery/api for the other updates, but got a little stuck and moving on for now.

tswast:
Build for PHP 5.6 fails with

```
+composer install -q
Your requirements could not be resolved to an installable set of packages.

The command "testing/run_all_tests.sh" exited with 1.
```

I don't know what to do about that.

bshaffer:
It's probably because you ran `composer install` on a machine running a PHP version above 5.6, and so some dependencies are locked with versions that are only compatible with PHP > 5.6

tswast:
Ah, yeah that's what I did. I was using PHP 7.2. Just pushed a commit after trying again with PHP 5.6.

PR #522, opened 2017-12-27
[WorkInProgress] update bigquery/api samples to latest package version.

bshaffer:


bshaffer:
BigQuery samples were updated in #543

PR #544, opened 2018-01-24
Add PHP 7.2 support for IAP

bshaffer:
Adds PHP 7.2 support to IAP samples by using a different JWT library which supports PHP 7.2

### DIFF
--- a/iap/src/validate_jwt.php
+++ b/iap/src/validate_jwt.php
@@ -24,10 +24,8 @@
 namespace Google\Cloud\Samples\Iap;
 
 # Imports OAuth Guzzle HTTP libraries.
-use GuzzleHttp\Client;
-use Lcobucci\JWT\Parser;
-use Lcobucci\JWT\ValidationData;
-use Lcobucci\JWT\Signer\Ecdsa\Sha256;
+use Jose\Factory\JWKFactory;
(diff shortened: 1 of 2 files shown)


bshaffer:
This PR is dependent on #543 for tests to pass

iap/src/validate_jwt.php
```diff
@@ -76,27 +74,23 @@ function validate_jwt_from_compute_engine($iap_jwt, $cloud_project_number, $back
...
-    $data->setIssuer('https://cloud.google.com/iap');
-    $data->setAudience($expected_audience);
-    assert($token->validate($data));
+    // Validate the signature using the key set and ES256 algorithm.
+    $loader = new Loader();
+    $jws = $loader->loadAndVerifySignatureUsingKeySet(
+        $iap_jwt,
+        $jwk_set,
+        ['ES256'],
+        $signature_index
```
ryanmats: what is this variable and where is it coming from?

bshaffer: Yep! It's kinda silly, but if the validation is successful, it returns the index in the `$jwk_set` that is used by the key. Since it's not important to our sample, and it's optional, I'll remove it.

PR #545, opened 2018-01-25
Adding sample applications for supervisord config

tmatsuo:


appengine/flexible/supervisord/replacement/test/bootstrap.php
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+require __DIR__ . '/../vendor/autoload.php';
+
+use Google\Cloud\TestUtils\PHPUnit4To6Shim;
+
+PHPUnit4To6Shim::install();
```
bshaffer: I am not sure why this is necessary, as it seems like just using `vendor/bin/phpunit` and installing `phpunit/phpunit:~4.8` works fine.

tmatsuo:
Moved to `additional`.

Combining the directory is tricky because if you have both `supervisord.conf` and `additional-supervisord.conf`, only the former will take effect, and there's no way to e2e test `additional-supervisord.conf`.

tmatsuo:
@bshaffer Because I want to show both examples on the documentation site with includecode.

bshaffer:
I meant it rhetorically, as in, Q: Why would an application need both? A: It wouldn't.

tmatsuo:
Sometimes you don't need nginx and php-fpm at all
Sometimes you just want to add your worker to the nginx + php-fpm

bshaffer:
Sorry, I mean Q: Why would a *single* application need to have both files at the same time? A: they wouldn't.

tmatsuo:
Ah sorry yeah I got it

PR #546, opened 2018-01-26
Misc test fixes

bshaffer:


error_reporting/composer.json
```diff
@@ -7,5 +7,10 @@
     "require-dev": {
         "phpunit/phpunit": "~4.8",
         "google/cloud-tools": "^0.6.9"
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Google\\Cloud\\Samples\\Monitoring\\": "test/lib"
```
tmatsuo: The directory is `error_reporting` while the namespace is `Monitoring`?

error_reporting/test/error_reportingTest.php
```diff
@@ -17,19 +17,13 @@
 
 namespace Google\Cloud\Samples\Monitoring;
```
tmatsuo: Should we match the namespace to the directory?

error_reporting/test/quickstartTest.php
```diff
@@ -15,17 +15,13 @@
  * limitations under the License.
  */
 
-use Google\Auth\ApplicationDefaultCredentials;
-use GuzzleHttp\Client;
-use GuzzleHttp\HandlerStack;
-use Google\Cloud\TestUtils\EventuallyConsistentTestTrait;
+namespace Google\Cloud\Samples\Monitoring;
```
tmatsuo: ditto

error_reporting/composer.json
bshaffer: Good catch! These should all be fixed now.

tmatsuo:
The code looks good.

Do you know why one of the tests in logging was skipped?
https://travis-ci.org/GoogleCloudPlatform/php-docs-samples/jobs/337274790#L4223

bshaffer:
@tmatsuo amazing!! Yes, I renamed a test from `testWrite` to `testWriteAndList` and forgot to update the next test which had `@depends testWrite`. This is now fixed.

PR #550, opened 2018-02-08
Add sample for batch/async enabled PSR3 logger

tmatsuo:


tmatsuo:
Oh the tests are failing because there's no secret available...

I will close this and create another PR from the same repo

PR #553, opened 2018-02-13
Added region tags

ellenevans:


appengine/flexible/tasks/tasks.php
```diff
@@ -35,7 +35,7 @@
     new InputOption('payload', 'helloworld', InputOption::VALUE_OPTIONAL, 'Optional payload to attach to the App Engine HTTP request.'),
     new InputOption('seconds', 0, InputOption::VALUE_OPTIONAL, 'The number of seconds from now to schedule task attempt.'),
 ]);
-
+// [START example_task_handler]
```
bshaffer: We don't really want to add this sample to our docs, so we don't need these region tags

bshaffer: This is the sample you're looking for: https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/flexible/tasks/src/create_task.php#L32

PR #554, opened 2018-02-13
Updated region tags for create task

ellenevans:


bshaffer:
The region you're looking for is here: https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/flexible/tasks/src/create_task.php#L32

PR #555, opened 2018-02-13
Add region tags for create and pull/acknowledge

ellenevans:


bshaffer:
The code you're looking for is here: https://github.com/ellenevans/php-docs-samples/tree/4834072f82fc200ed103ab101178d57b8463545b/tasks/src

PR #556, opened 2018-02-20
Add logging configuration to the symfony sample

tmatsuo:


appengine/flexible/symfony/README.md
```diff
@@ -20,6 +20,13 @@ Use composer to download Symfony Standard and its dependencies
 composer create-project symfony/symfony:^3.0
 ```
 
+# Add dependency
```
bshaffer: This would be more clear as `Add Composer Dependencies`, but this enhancement would be better as a second section of the README, marked `# Integrate Stackdriver`

appengine/flexible/symfony/README.md
```diff
@@ -31,13 +38,15 @@ git clone https://github.com/GoogleCloudPlatform/php-docs-samples /path/to/php-d
 
 # copy the two files below to the root directory of your Symfony project
 cd /path/to/php-docs-samples/appengine/flexible/symfony/
-cp ./{app.yaml,nginx-app.conf} /path/to/symfony
+cp app.yaml app/config/config_prod.yml \
+  src/AppBundle/EventSubscriber/ExceptionSubscriber.php /path/to/symfony
```
bshaffer: This isn't going to work because it will copy these files into the root of their project, and so `ExceptionSubscriber.php` will not be autoloaded, and `config_prod.yaml` will not be in the config directory.

It would be better to add `ExceptionSubscriber` to our client libraries, or instead of copying it over, explain in this tutorial how to add the exception Subscriber in symfony, and link to [this doc](https://symfony.com/doc/current/event_dispatcher.html#creating-an-event-subscriber)

appengine/flexible/symfony/README.md
tmatsuo: Done (with the next commit)

appengine/flexible/symfony/README.md
tmatsuo: Fixed in the next commit

tmatsuo:
@bshaffer Ready for another look. The deployment test is passing locally

appengine/flexible/symfony/README.md
```diff
@@ -29,28 +38,20 @@ directory:
...
 
-# copy the two files below to the root directory of your Symfony project
+# copy the three files below to your Symfony project
 cd /path/to/php-docs-samples/appengine/flexible/symfony/
-cp ./{app.yaml,nginx-app.conf} /path/to/symfony
+mkdir -p /path/to/symfony/src/AppBundle/EventSubscriber
+cp app.yaml /path/to/symfony
+cp app/config/config_prod.yml /path/to/symfony/app/config
+cp src/AppBundle/EventSubscriber/ExceptionSubscriber.php \
+    /path/to/symfony/src/AppBundle/EventSubscriber
```
bshaffer: By following this tutorial, the app won't have an `AppBundle/EventSubscriber` directory. So we'll need to add a step to create that directory. Also, I would think we'd need to add `AppBundle` to `composer.json` for autoloading, but strangely the tests do not seem to do that, so I'm not totally sure how that class is being loaded and integrated.

tmatsuo: there is a line for mkdir. The file is automatically loaded by symfony.

bshaffer: Ahh I missed it. Let's call out "Create a directory for the Exception Subscriber" in a comment and separate it from "copy the three files" 

Looking at the [`composer.json`](https://github.com/symfony/symfony/blob/master/composer.json) generated by symfony, I do not see anywhere that autoloading would occur.

tmatsuo: @bshaffer Done for the README.

It's automatically loaded. See:
http://symfony.com/doc/3.4/event_dispatcher.html

> That's it! Your services.yml file should already be setup to load services from the EventSubscriber directory. Symfony takes care of the rest.

PR #557, opened 2018-02-22
Add logging configuration to the laravel sample

tmatsuo:


appengine/flexible/laravel/composer.json
```diff
@@ -1,6 +1,6 @@
 {
     "require-dev": {
-        "phpunit/phpunit":"~4.8",
+        "google/cloud-logging": "^1.9",
```
bshaffer: We should also add `google/cloud-error-reporting` for consistency

tmatsuo: No it's solely for testing. `google/cloud-error-reporting` is not needed for the test.

tmatsuo:
@bshaffer 

Re: README
When you added this directory, apparently you made a decision not to duplicate the same contents in the README as our community tutorial. Let's follow this decision here. Once I update our Stackdriver documentation, I'll update the README with links to those documentation.

tmatsuo:
@bshaffer The DeployTest is passing locally.

PR #561, opened 2018-02-28
Add Stackdriver Trace sample code for the doc

tmatsuo:
cc @chingor13


testing/run_test_suite.sh
```diff
@@ -65,6 +65,7 @@ do
         fi
     fi
     pushd ${DIR}
+    mkdir -p build/logs
```
bshaffer: 🎉

trace/trace-sample.php
```diff
@@ -0,0 +1,56 @@
...
+
+# [END trace_use_statement]
+
+$projectId = getenv('GOOGLE_CLOUD_PROJECT');
+if ($projectId === false) {
+    die('Set GOOGLE_CLOUD_PROJECT envvar');
+}
+
+# [START exporter_setup]
+$exporter = new StackdriverExporter(
```
bshaffer: It's not in the PSR coding standards, but I've been doing these like this, so it'd be better for internal sample consistency:

```php
$exporter = new StackdriverExporter([
    'clientConfig' => [
        'projectId' => $projectId
    ]
]);
```

trace/trace-sample.php
```diff
@@ -0,0 +1,56 @@
...
+if (getenv('USE_NULL_EXPORTER')) {
+    $exporter = new NullExporter();
+}
+# [START tracer_start]
+Tracer::start($exporter);
+# [END tracer_start]
+
+function use_trace()
+{
+    # [START span_with_closure]
```
bshaffer: The reference block here is off.. the `END` ref should be inside the function

trace/trace-sample.php
```diff
@@ -0,0 +1,56 @@
...
+    die('Set GOOGLE_CLOUD_PROJECT envvar');
+}
+
+# [START exporter_setup]
+$exporter = new StackdriverExporter(
+    ['clientConfig' => [
+        'projectId' => $projectId
+    ]]);
+# [END exporter_setup]
+// For testing
```
bshaffer: How about this comment instead? 

```php
// When running tests, use a null exporter instead.
```

But a better question is, why not just use `StackdriverExporter`?

Or, in the tests, just call `Trace::start(new NullExporter);` before calling `use_trace()`?

trace/trace-sample.php
```diff
@@ -0,0 +1,56 @@
...
+if (getenv('USE_NULL_EXPORTER')) {
+    $exporter = new NullExporter();
+}
+# [START tracer_start]
+Tracer::start($exporter);
+# [END tracer_start]
+
+function use_trace()
```
bshaffer: I would like a more descriptive name for this function here. How about `trace_callable()`?

trace/trace-sample.php
tmatsuo: Done (with the next commit)

trace/trace-sample.php
tmatsuo: done

trace/trace-sample.php
tmatsuo: Having integration test is good in general, but unfortunately now there's no way to fetch the trace from the API. I would just have the unit test like this for now.

trace/trace-sample.php
tmatsuo: Done

tmatsuo:
@bshaffer Thanks, PTAL

chingor13:
LGTM

PR #563, opened 2018-03-02
Spanner Batch Query Sample

ryanmats:


bshaffer:
Travis is failing on the code standards check.

PR #564, opened 2018-03-02
[WIP, DO NOT MERGE] DLP v2 samples

ace-n:
**DO NOT MERGE - This is a work in progress.**


ace-n:
Brent - can you take a look at these? They should be reviewable (though not yet mergeable).

dlp/dlp.php
```diff
@@ -27,81 +27,434 @@
 
 $application->add(new Command('inspect-string'))
     ->addArgument('string', InputArgument::REQUIRED, 'The text to inspect')
-    ->setDescription('Inspect a string using the Data Loss Prevention (DLP) API.')
+    ->addArgument('calling-project', InputArgument::OPTIONAL, 'The GCP Project ID to run the API call under', getenv('GOOGLE_PROJECT_ID'))
```
bshaffer: just make this `project` to be consistent with all the other samples. And do this to all the other `calling-project` options as well.

dlp/dlp.php
```diff
@@ -27,81 +27,434 @@
...
+    ->setDescription('Computes the k-anonymity of a column set in a Google BigQuery table.')
+    ->setCode(function ($input, $output) {
+        k_anonymity(
+            (string) $input->getArgument('calling-project'),
+            (string) $input->getArgument('data-project'),
+            $input->getArgument('topic-id'),
+            $input->getArgument('subscription-id'),
+            $input->getArgument('dataset'),
+            $input->getArgument('table'),
+            (array) $input->getArgument('quasi-ids')
```
bshaffer: This is just going to cast whatever string is supplied into an array surrounding that string. So if more than one is supplied, they'll still be in a single string. What you probably want is to use a delimiter and the `explode` method.

dlp/dlp.php
```diff
@@ -27,81 +27,434 @@
...
+        k_map(
+            (string) $input->getArgument('calling-project'),
+            (string) $input->getArgument('data-project'),
+            $input->getArgument('topic-id'),
+            $input->getArgument('subscription-id'),
+            $input->getArgument('dataset'),
+            $input->getArgument('table'),
+            $input->getArgument('region-code'),
+            (array) $input->getArgument('quasi-ids'),
+            (array) $input->getArgument('info-types')
```
bshaffer: Same issue here. This allows for more than one info types, but there is no delimiter mentioned and the array cast will not have the appropriate effect.

dlp/src/deidentify_dates.php
```diff
@@ -0,0 +1,182 @@
...
+                $date->setMonth((int) $csvDate['tm_mon']);
+                $date->setDay((int) $csvDate['tm_mday']);
+
+                $protoDate = new Value();
+                $protoDate->setDateValue($date);
+                return $protoDate;
+            } else {
+                $protoString = new Value();
+                $protoString->setStringValue($csvValue);
+                //var_dump($protoString->getStringValue());
```
bshaffer: remove this line!

dlp/src/deidentify_dates.php
```diff
@@ -0,0 +1,182 @@
...
+        return (new FieldId)->setName($csvHeader);
+    }, $csvHeaders);
+
+    $tableRows = array_map(function ($csvRow) {
+        $rowValues = array_map(function ($csvValue) {
+            if ($csvDate = strptime($csvValue, '%m/%d/%Y')) {
+                $date = new Date();
+                $date->setYear((int) $csvDate['tm_year']);
+                $date->setMonth((int) $csvDate['tm_mon']);
+                $date->setDay((int) $csvDate['tm_mday']);
```
bshaffer: These methods are fluid so we can make this more readable by writing it like:

```php
$date = (new Date())
    ->setYear((int) $csvDate['tm_year'])
    ->setMonth((int) $csvDate['tm_mon'])
    ->setDay((int) $csvDate['tm_mday']);
```

dlp/src/deidentify_dates.php
```diff
@@ -0,0 +1,182 @@
...
+
+                $protoDate = new Value();
+                $protoDate->setDateValue($date);
+                return $protoDate;
```
bshaffer: could write this:
```php
return (new Value())
    ->setDateValue($date);
```

dlp/src/deidentify_dates.php
```diff
@@ -0,0 +1,182 @@
...
+        return (new FieldId())->setName($dateFieldName);
+    }, $dateFieldNames);
+
+    // Construct the table object
+    $table = new Table();
+    $table->setHeaders($tableHeaders);
+    $table->setRows($tableRows);
+
+    $item = new ContentItem();
+    //$item->setType('text/csv');
```
bshaffer: Is this comment here intentionally?

dlp/src/deidentify_fpe.php
```diff
@@ -0,0 +1,115 @@
...
+    $parent = $dlp->projectName($callingProjectId);
+
+    // Run request
+    $response = $dlp->deidentifyContent($parent, [
+        'deidentifyConfig' => $deidentifyConfig,
+        'item' => $content
+    ]);
+
+    $likelihoods = ['Unknown', 'Very unlikely', 'Unlikely', 'Possible',
+                  'Likely', 'Very likely'];
```
bshaffer: whitespace is off

dlp/src/create_trigger.php
```diff
@@ -0,0 +1,124 @@
...
+
+    // Specify finding limits
+    $limits = new InspectConfig_FindingLimits();
+    $limits->setMaxFindingsPerRequest($maxFindings);
+
+    // Create the inspectConfig object
+    $inspectConfig = new InspectConfig();
+    $inspectConfig->setMinLikelihood($minLikelihood);
+    $inspectConfig->setLimits($limits);
+    $inspectConfig->setInfoTypes($infoTypes);
```
bshaffer: Take advantage of fluent methods here:

```php
$inspectConfig = (new InspectConfig())
    ->setMinLikelihood($minLikelihood)
    ->setLimits($limits)
    ->setInfoTypes($infoTypes);
```

dlp/src/create_trigger.php
```diff
@@ -0,0 +1,124 @@
...
+    $jobConfig->setInspectConfig($inspectConfig);
+    $jobConfig->setStorageConfig($storageConfig);
+
+    // ----- Construct trigger object -----
+    $jobTriggerObject = new JobTrigger();
+    $jobTriggerObject->setTriggers([$triggerObject]);
+    $jobTriggerObject->setInspectJob($jobConfig);
+    $jobTriggerObject->setStatus(JobTrigger_Status::HEALTHY);
+    $jobTriggerObject->setDisplayName($displayName);
+    $jobTriggerObject->setDescription($description);
```
bshaffer: Take advantage of fluent methods here

dlp/src/inspect_datastore.php
```diff
@@ -68,41 +84,77 @@ function inspect_datastore(
...
-            print('No findings.' . PHP_EOL);
-        } else {
-            print('Findings:' . PHP_EOL);
-            foreach ($findings as $finding) {
-                printf('- Info type: %s' . PHP_EOL,
-                    $finding->getInfoType()->getName());
-                printf('  Likelihood: %s' . PHP_EOL,
-                    $likelihoods[$finding->getLikelihood()]);
+    // Construct the action to run when job completes
+    $fullTopicId = 'projects/' . $callingProjectId . '/topics/' . $topicId;
```
bshaffer: use `sprintf`!

```php
$fullTopicId = sprintf('projects/%s/topics%s', $projectId, $topicId);
```

dlp/src/inspect_bigquery.php
```diff
@@ -17,85 +17,138 @@
...
+use Google\Cloud\Dlp\V2\DlpServiceClient;
+use Google\Cloud\Dlp\V2\BigQueryOptions;
+use Google\Cloud\Dlp\V2\InfoType;
+use Google\Cloud\Dlp\V2\InspectConfig;
+use Google\Cloud\Dlp\V2\StorageConfig;
+use Google\Cloud\Dlp\V2\BigQueryTable;
+use Google\Cloud\Dlp\V2\Likelihood;
+use Google\Cloud\Dlp\V2\DlpJob_JobState;
+use Google\Cloud\Dlp\V2\InspectConfig_FindingLimits;
+use Google\Cloud\PubSub\PubSubClient;
```
bshaffer: If we are using pubsub, this needs to be added to `composer.json`:

```sh
    "require": {
        "google/cloud-dlp": "^0.5",
        "symfony/console": "^3.3",
        "google/cloud-pubsub": "^1.0"
    },
```

dlp/src/inspect_datastore.php
bshaffer: Alternatively the GAPIC clients have format methods you can use:

```php
use Google\Cloud\PubSub\V1\PublisherClient;

$fullTopicId = PublisherClient::topicName($projectId, $topicId);
```

dlp/src/inspect_file.php
```diff
@@ -17,56 +17,76 @@
...
     // Run request
-    $response = $dlp->inspectContent($inspectConfig, [$content]);
+    $response = $dlp->inspectContent($parent, [
+        'inspectConfig' => $inspectConfig,
+        'item' => $content
+    ]);
 
     $likelihoods = ['Unknown', 'Very unlikely', 'Unlikely', 'Possible',
-                    'Likely', 'Very likely'];
+                  'Likely', 'Very likely'];
```
bshaffer: This is the wrong whitespace and should be reverted to the way it was

dlp/src/inspect_gcs.php
```diff
@@ -0,0 +1,151 @@
...
+    $pubSubAction->setTopic($fullTopicId);
+
+    $action = new Action();
+    $action->setPubSub($pubSubAction);
+
+    // Construct inspect job config to run
+    $inspectJob = new InspectJobConfig();
+    $inspectJob->setInspectConfig($inspectConfig);
+    $inspectJob->setStorageConfig($storageConfig);
+    $inspectJob->setActions([$action]);
```
bshaffer: Consider using fluent interface

dlp/src/inspect_gcs.php
```diff
@@ -0,0 +1,151 @@
...
+    $storageConfig->setCloudStorageOptions($cloudStorageOptions);
+
+    // Construct the inspect config object
+    $inspectConfig = new InspectConfig();
+    $inspectConfig->setMinLikelihood($minLikelihood);
+    $inspectConfig->setLimits($limits);
+    $inspectConfig->setInfoTypes($infoTypes);
+
+    // Construct the action to run when job completes
+    $fullTopicId = 'projects/' . $callingProjectId . '/topics/' . $topicId;
```
bshaffer: Use the GAPIC formatting methods.

dlp/src/inspect_gcs.php
```diff
@@ -0,0 +1,151 @@
...
+    $cloudStorageOptions->setFileSet($fileSet);
+
+    $storageConfig = new StorageConfig();
+    $storageConfig->setCloudStorageOptions($cloudStorageOptions);
+
+    // Construct the inspect config object
+    $inspectConfig = new InspectConfig();
+    $inspectConfig->setMinLikelihood($minLikelihood);
+    $inspectConfig->setLimits($limits);
+    $inspectConfig->setInfoTypes($infoTypes);
```
bshaffer: Consider using fluent interface

dlp/src/inspect_gcs.php
```diff
@@ -0,0 +1,151 @@
...
+                $polling = false;
+            }
+        }
+    }
+
+    // Get the updated job
+    $job = $dlp->getDlpJob($job->getName());
+
+    // Print finding counts
+    print_r('Job ' . $job->getName() . ' status: ' . $job->getState() . PHP_EOL);
```
bshaffer: You don't want `print_r` here because that's usually for arrays. Normal print or printf:

```php
printf('Job %s status %s' . PHP_EOL, $job->getName(), $job->getState());
```

dlp/src/inspect_gcs.php
```diff
@@ -0,0 +1,151 @@
...
+              print_r('  Found ' . $infoTypeStat->getCount() . ' instance(s) of infoType ' . $infoTypeStat->getInfoType()->getName() . PHP_EOL);
+          }
+      }
+      break;
+    case DlpJob_JobState::ERROR:
+      $errors = $job->getErrors();
+      foreach ($errors as $error) {
+          var_dump($error->getDetails());
+      }
+      print_r('Job ' . $job->getName() . ' had errors:' . PHP_EOL);
```
bshaffer: replace with `printf`

dlp/src/inspect_gcs.php
```diff
@@ -0,0 +1,151 @@
...
+    // Print finding counts
+    print_r('Job ' . $job->getName() . ' status: ' . $job->getState() . PHP_EOL);
+    switch ($job->getState()) {
+    case DlpJob_JobState::DONE:
+      $infoTypeStats = $job->getInspectDetails()->getResult()->getInfoTypeStats();
+      if (count($infoTypeStats) === 0) {
+          print_r('No findings.' . PHP_EOL);
+      } else {
+          foreach ($infoTypeStats as $infoTypeStat) {
+              print_r('  Found ' . $infoTypeStat->getCount() . ' instance(s) of infoType ' . $infoTypeStat->getInfoType()->getName() . PHP_EOL);
```
bshaffer: replace with printf

dlp/src/inspect_gcs.php
```diff
@@ -0,0 +1,151 @@
...
+    // Get the updated job
+    $job = $dlp->getDlpJob($job->getName());
+
+    // Print finding counts
+    print_r('Job ' . $job->getName() . ' status: ' . $job->getState() . PHP_EOL);
+    switch ($job->getState()) {
+    case DlpJob_JobState::DONE:
+      $infoTypeStats = $job->getInspectDetails()->getResult()->getInfoTypeStats();
+      if (count($infoTypeStats) === 0) {
+          print_r('No findings.' . PHP_EOL);
```
bshaffer: replace with `print`:
```php
print('No findings.' . PHP_EOL);
```

dlp/src/inspect_string.php
```diff
@@ -17,55 +17,69 @@
...
     // Run request
-    $response = $dlp->inspectContent($inspectConfig, [$content]);
+    $response = $dlp->inspectContent($parent, [
+        'inspectConfig' => $inspectConfig,
+        'item' => $content
+    ]);
 
     $likelihoods = ['Unknown', 'Very unlikely', 'Unlikely', 'Possible',
-                    'Likely', 'Very likely'];
+                  'Likely', 'Very likely'];
```
bshaffer: why are these all like this??

dlp/quickstart.php
```diff
@@ -47,25 +48,33 @@
...
+
 // Run request
-$response = $dlp->inspectContent($inspectConfig, [$content]);
+$response = $dlp->inspectContent($parent, array(
+    'inspectConfig' => $inspectConfig,
+    'item' => $content
+));
 
 $likelihoods = ['Unknown', 'Very unlikely', 'Unlikely', 'Possible',
                 'Likely', 'Very likely'];
```
bshaffer: This is something we should eventually get into the `GAPIC` layer, e.g:

```php
Likelihood::fromEnum($finding->getLikelihood());
```

But in the meantime maybe we can get it in the Gapic Client layer, e.g.

```php
DlpServiceClient::likelihoodFromEnum($finding->getLikelihood());
```

@michaelbausor thoughts?

michaelbausor: Something like this would be nice on the protobuf class. Re adding to the gapic client - my concern would be cases when there are a lot of enums, potentially rarely used in deeply nested messages, or in shared protos. So perhaps we would need a configuration to say which enums we should generate this for?

Also, in this case it looks like the goal is a user-readable string, but we would probably want to convert to/from an underscore string, likely the name of the constant: https://github.com/GoogleCloudPlatform/google-cloud-php-private/blob/dlp-v2/src/Dlp/V2/Likelihood.php#L26

bshaffer:
We need to add the new files to `composer.json` in the autoload.files section, and remove the files which no longer exist.

dlp/dlp.php
ace-n: My goal here was to disambiguate between _the project calling the API_ and _the project containing the scanned data_, which can differ in some cases.

Should I make this change - perhaps to something like `project` and `dataProject`?

dlp/dlp.php
ace-n: Done.

dlp/dlp.php
ace-n: Done.

dlp/src/deidentify_dates.php
ace-n: Done.

dlp/src/deidentify_dates.php
ace-n: Done.

dlp/src/deidentify_dates.php
ace-n: Done.

dlp/src/deidentify_dates.php
ace-n: Removed.

dlp/src/deidentify_fpe.php
ace-n: Done.

dlp/src/create_trigger.php
ace-n: Done.

dlp/src/create_trigger.php
ace-n: Done.

dlp/src/inspect_datastore.php
ace-n: Done.

dlp/src/inspect_bigquery.php
ace-n: Done.

dlp/src/inspect_file.php
ace-n: Done.

dlp/src/inspect_gcs.php
ace-n: Done.

dlp/src/inspect_gcs.php
ace-n: Using `sprintf` - I assume that's OK?

dlp/src/inspect_gcs.php
ace-n: Done.

dlp/src/inspect_gcs.php
ace-n: Done.

dlp/src/inspect_gcs.php
ace-n: Done.

dlp/src/inspect_gcs.php
ace-n: Done.

dlp/src/inspect_string.php
ace-n: Done.

ace-n:
Closed in favor of #576 to avoid force-pushing.

PR #569, opened 2018-03-08
Change to endpoints managed rollout

simonz130:
We introduce a new feature where we can specify in YAML that service configuration for Endpoints is going to be managed. This means that the service will pull latest deployed service configuration automatically, rather than requiring customer to specify specific version of service configuration.

New flag:
**--rollout_strategy=managed**
Previously the only way to specify a version was using the following flag:
**--version=SERVICE_CONFIG_ID**


simonz130:
Yes, it will work for first deployment too. It should fetch the last deployed version of endpoints.
If you first deploy appengine flex, then "managed rollout" won't take any effect until you deploy endpoints config.

tmatsuo:
Ok, I assume the Kokoro config has not yet been configured fully.

PR #574, opened 2018-03-13
Firestore List subcollections sample + tests

ryanmats:
One more Firestore sample!


firestore/composer.json
```diff
@@ -20,6 +20,7 @@
             "src/set_requires_id.php",
             "src/add_doc_data_with_auto_id.php",
             "src/add_doc_data_after_auto_id.php",
+            "src/list_subcollections.php",
```
bshaffer: is there a reason `add_subcollection` is not a function we're adding? I assume it's just because it wasn't asked for as a sample?

PR #576, opened 2018-03-14
DLP v2 samples

ace-n:
#564, now with 100% more rebasing-on-master.


dlp/composer.json
```diff
@@ -3,17 +3,35 @@
     "type": "project",
     "require": {
         "google/cloud-dlp": "^0.5",
```
bshaffer: This will need to be updated to `^0.6` once the v2 release goes out

dlp/src/categorical_stats.php
```diff
@@ -0,0 +1,159 @@
...
+        ->setCategoricalStatsConfig($statsConfig);
+
+    // Construct items to be analyzed
+    $bigqueryTable = (new BigQueryTable())
+        ->setProjectId($dataProjectId)
+        ->setDatasetId($datasetId)
+        ->setTableId($tableId);
+
+    // Construct the action to run when job completes
+    $fullTopicId = PublisherClient::topicName($callingProjectId, $topicId);
```
bshaffer: You can get this by using the `$topic` var below, e.g.:

```php
$topic = $pubsub->topic($topicId);
$fullTopicId = $topic->name();
```

Then we can remove `PublisherClient` from the use statement and make this slightly simpler.

dlp/src/categorical_stats.php
```diff
@@ -0,0 +1,159 @@
...
+                $polling = false;
+            }
+        }
+    }
+
+    // Get the updated job
+    $job = $dlp->getDlpJob($job->getName());
+
+    // Helper function to convert Protobuf values to strings
+    $value_to_string = function ($value) {
```
bshaffer: Is this really where we're at? We don't have a better way to do this? Do other languages have a better way? Because we should fix this ASAP.

dlp/src/categorical_stats.php
```diff
@@ -0,0 +1,159 @@
...
+                printf('  Most common value occurs %s time(s)' . PHP_EOL, $histBucket->getValueFrequencyUpperBound());
+                printf('  Least common value occurs %s time(s)' . PHP_EOL, $histBucket->getValueFrequencyLowerBound());
+                printf('  %s unique value(s) total.', $histBucket->getBucketSize());
+
+                // Print bucket values
+                foreach ($histBucket->getBucketValues() as $percent => $quantile) {
+                    printf(
+                        '  Value %s occurs %s time(s).' . PHP_EOL,
+                      $value_to_string($quantile->getValue()),
+                      $quantile->getCount()
```
bshaffer: Whitespace is off on these two lines, should be aligned with `'  Value...`

dlp/src/create_trigger.php
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Create a Data Loss Prevention API job trigger.
+ *
+ * @param string $callingProjectId The project ID to run the API call under
+ * @param string $bucketName The name of the bucket to scan
+ * @param string $displayName (Optional) The human-readable name to give the trigger
+ * @param string $description (Optional) A description for the trigger to be created
+ * @param int $scanPeriod How often to wait between scans, in days (minimum = 1 day)
+ * @param int $maxFindings The maximum number of findings to report per request (0 = server maximum)
```
bshaffer: these two are also optional

dlp/src/create_inspect_template.php
```diff
@@ -0,0 +1,92 @@
...
+use Google\Cloud\Dlp\V2\InspectConfig_FindingLimits;
+
+/**
+ * Create a new DLP inspection configuration template.
+ *
+ * @param string $callingProjectId The project ID to run the API call under
+ * @param string $templateId The name of the template to be created
+ * @param string $displayName Optional The human-readable name to give the template
+ * @param string $description Optional A description for the trigger to be created
+ * @param int $maxFindings The maximum number of findings to report per request (0 = server maximum)
```
bshaffer: You say `$displayName` and `$description` are optional but they do not have a default value in the function declaration below. 

And if they are optional, then `$maxFindings` needs to be optional too (and labeled as such)

dlp/src/deidentify_dates.php
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Deidentify dates in a CSV file by pseudorandomly shifting them.
+ *
+ * @param string $callingProject The GCP Project ID to run the API call under
+ * @param string $inputCsvFile The path to the CSV file to deidentify
+ * @param string $outputCsvFile The path to save the date-shifted CSV file to
+ * @param array $dateFieldNames The list of (date) fields in the CSV file to date shift
+ * @param string $lowerBoundDays The maximum number of days to shift a date backward
+ * @param string $upperBoundDays The maximum number of days to shift a date forward
+ * @param string contextFieldName optional The column to determine date shift amount based on
```
bshaffer: Let's be consistent and capitalize `optional` here. Also, make sure they have an optional value (e.g. empty string) in the function below, otherwise they aren't really optional.

dlp/src/deidentify_dates.php
```diff
@@ -0,0 +1,181 @@
...
+                $protoString = (new Value())
+                    ->setStringValue($csvValue);
+                return $protoString;
+            }
+        }, explode(',', $csvRow));
+
+        $tableRow = (new Table_Row())
+            ->setValues($rowValues);
+        return $tableRow;
+    }, $csvRows);
```
bshaffer: looks like this brings the entire CSV file into memory? That's not great.

dlp/src/deidentify_fpe.php
```diff
@@ -0,0 +1,115 @@
+<?php
+
+/**
+ * Copyright 2016 Google Inc.
```
bshaffer: Copyright headers should be `2018` on new files!

dlp/src/categorical_stats.php
```diff
@@ -0,0 +1,159 @@
...
+ * limitations under the License.
+ */
+namespace Google\Cloud\Samples\Dlp;
+
+# [START dlp_categorical_stats]
+use Google\Cloud\Dlp\V2\DlpServiceClient;
+use Google\Cloud\Dlp\V2\RiskAnalysisJobConfig;
+use Google\Cloud\Dlp\V2\BigQueryTable;
+use Google\Cloud\Dlp\V2\DlpJob_JobState;
+use Google\Cloud\PubSub\PubSubClient;
```
bshaffer: Nit but this should be at the bottom instead of in the middle of the `/Dlp/` namespaces

dlp/src/inspect_bigquery.php
```diff
@@ -18,84 +18,142 @@
...
+use Google\Cloud\Dlp\V2\DlpServiceClient;
+use Google\Cloud\Dlp\V2\BigQueryOptions;
+use Google\Cloud\Dlp\V2\InfoType;
+use Google\Cloud\Dlp\V2\InspectConfig;
+use Google\Cloud\Dlp\V2\StorageConfig;
+use Google\Cloud\Dlp\V2\BigQueryTable;
+use Google\Cloud\Dlp\V2\Likelihood;
+use Google\Cloud\Dlp\V2\DlpJob_JobState;
+use Google\Cloud\Dlp\V2\InspectConfig_FindingLimits;
+use Google\Cloud\PubSub\PubSubClient;
```
bshaffer: move to bottom of namespace list

dlp/src/inspect_bigquery.php
```diff
@@ -18,84 +18,142 @@
...
-    $operation = $dlp->createInspectOperation($inspectConfig, $storageConfig, $outputConfig);
+    // Construct the inspect config object
+    $inspectConfig = (new InspectConfig())
+        ->setMinLikelihood($minLikelihood)
+        ->setLimits($limits)
+        ->setInfoTypes($infoTypes);
 
-    $operation->pollUntilComplete();
+    // Construct the action to run when job completes
+    $fullTopicId = PublisherClient::topicName($callingProjectId, $topicId);
```
bshaffer: again we can remove this and just use `$topic->name()` to simplify it.

dlp/src/inspect_bigquery.php
```diff
@@ -18,84 +18,142 @@
...
+    ]);
+
+    // Poll via Pub/Sub until job finishes
+    $polling = true;
+    while ($polling) {
+        foreach ($subscription->pull() as $message) {
+            if (isset($message->attributes()['DlpJobName']) &&
+                $message->attributes()['DlpJobName'] === $job->getName()) {
+                $subscription->acknowledge($message);
+                $polling = false;
```
bshaffer: why not use `break`? Because you want to finish the cycle of pulling first?

dlp/src/inspect_bigquery.php
```diff
@@ -18,84 +18,142 @@
...
+        }
+    }
+
+    // Get the updated job
+    $job = $dlp->getDlpJob($job->getName());
+
+    // Print finding counts
+    printf('Job %s status: %s' . PHP_EOL, $job->getName(), $job->getState());
+    switch ($job->getState()) {
+    case DlpJob_JobState::DONE:
```
bshaffer: whitespace indentation here is off

dlp/src/inspect_datastore.php
```diff
@@ -18,91 +18,144 @@
...
+        ->setInfoTypes($infoTypes)
+        ->setMinLikelihood($minLikelihood)
+        ->setLimits($limits);
+
+    // Construct the storage config object
+    $storageConfig = (new StorageConfig())
+        ->setDatastoreOptions($datastoreOptions);
+
+    // Construct the action to run when job completes
+    $fullTopicId = PublisherClient::topicName($callingProjectId, $topicId);
```
bshaffer: can remove this by using `$topic->name()`

dlp/src/inspect_file.php
```diff
@@ -18,55 +18,80 @@
...
+        ->setMinLikelihood($minLikelihood)
+        ->setLimits($limits)
+        ->setInfoTypes($infoTypes)
+        ->setIncludeQuote($includeQuote);
+
+    // Create the content item objects
+    $typeConstant = array_search(
+        mime_content_type($path),
+        ['image/jpeg', 'image/bmp', 'image/png', 'image/svg']
+    ) + 1;
```
bshaffer: so if it's not in the list, it defaults to `image/jpeg`?

dlp/src/inspect_file.php
```diff
@@ -18,55 +18,80 @@
...
+
+
```
bshaffer: remove extra newline

dlp/src/inspect_gcs.php
```diff
@@ -0,0 +1,152 @@
...
+use Google\Cloud\Dlp\V2\DlpServiceClient;
+use Google\Cloud\Dlp\V2\CloudStorageOptions;
+use Google\Cloud\Dlp\V2\CloudStorageOptions_FileSet;
+use Google\Cloud\Dlp\V2\InfoType;
+use Google\Cloud\Dlp\V2\InspectConfig;
+use Google\Cloud\Dlp\V2\StorageConfig;
+use Google\Cloud\Dlp\V2\Likelihood;
+use Google\Cloud\Dlp\V2\DlpJob_JobState;
+use Google\Cloud\Dlp\V2\InspectConfig_FindingLimits;
+use Google\Cloud\PubSub\PubSubClient;
```
bshaffer: move to bottom of imports

dlp/src/k_anonymity.php
```diff
@@ -0,0 +1,170 @@
...
+ * limitations under the License.
+ */
+namespace Google\Cloud\Samples\Dlp;
+
+# [START dlp_k_anomymity]
+use Google\Cloud\Dlp\V2\DlpServiceClient;
+use Google\Cloud\Dlp\V2\RiskAnalysisJobConfig;
+use Google\Cloud\Dlp\V2\BigQueryTable;
+use Google\Cloud\Dlp\V2\DlpJob_JobState;
+use Google\Cloud\PubSub\PubSubClient;
```
bshaffer: move to bottom

dlp/src/k_map.php
```diff
@@ -0,0 +1,193 @@
...
+ */
+namespace Google\Cloud\Samples\Dlp;
+
+# [START dlp_k_map]
+use Google\Cloud\Dlp\V2\DlpServiceClient;
+use Google\Cloud\Dlp\V2\InfoType;
+use Google\Cloud\Dlp\V2\RiskAnalysisJobConfig;
+use Google\Cloud\Dlp\V2\BigQueryTable;
+use Google\Cloud\Dlp\V2\DlpJob_JobState;
+use Google\Cloud\PubSub\PubSubClient;
```
bshaffer: move to bottom

dlp/src/l_diversity.php
```diff
@@ -0,0 +1,187 @@
...
+ * limitations under the License.
+ */
+namespace Google\Cloud\Samples\Dlp;
+
+# [START dlp_l_diversity]
+use Google\Cloud\Dlp\V2\DlpServiceClient;
+use Google\Cloud\Dlp\V2\RiskAnalysisJobConfig;
+use Google\Cloud\Dlp\V2\BigQueryTable;
+use Google\Cloud\Dlp\V2\DlpJob_JobState;
+use Google\Cloud\PubSub\PubSubClient;
```
bshaffer: move to bottom

dlp/src/redact_image.php
```diff
@@ -0,0 +1,103 @@
...
+    fclose($imageRef);
+
+    // Get the image's content type
+    $mime_type = mime_content_type($imagePath);
+    $mime_index = array_search($mime_type, ['image/jpeg', 'image/bmp', 'image/png', 'image/svg']);
+    if ($mime_index === false) {
+        $mime_index = 0;
+    } else {
+        $mime_index++;
+    }
```
bshaffer: different from the mime_type calculation above. We should make them the same.

dlp/src/redact_image.php
```diff
@@ -0,0 +1,103 @@
...
+    $response = $dlp->redactImage($parent, [
+        'inspectConfig' => $inspectConfig,
+        'byteItem' => $byteContent,
+        'imageRedactionConfigs' => $imageRedactionConfigs
+    ]);
+
+    // Save result to file
+    $outputRef = fopen($outputPath, 'wb');
+    fwrite($outputRef, $response->getRedactedImage());
+    fclose($outputRef);
```
bshaffer: this can just be `file_put_contents($outputPath, $response->getRedactedImage());`

ace-n:
FYI - I found a less-ugly way to do `get_value` function (though I haven't tested it in PHP):

If you convert the raw protobuf value to a string, it typically spits out JSON - a la `{ integerValue: 1 }` or `{ stringValue: "foo" }`.

We could do something like `array_values(json_decode($protobufValue))[0];` to implement this, which is **a lot** cleaner than what's there now.

dlp/src/deidentify_dates.php
```diff
@@ -78,23 +79,20 @@ function deidentify_dates(
         $rowValues = array_map(function ($csvValue) {
             if ($csvDate = strptime($csvValue, '%m/%d/%Y')) {
                 $date = (new Date())
-                    ->setYear((int) $csvDate['tm_year'])
-                    ->setMonth((int) $csvDate['tm_mon'])
+                    ->setYear((int) $csvDate['tm_year'] + 1900)
```
ace-n: Is there a(n easy) way to get the year without doing such an offset?

bshaffer: yes! added!

PR #579, opened 2018-03-20
fixes DLP tests, marks them as not flakey

bshaffer:


dlp/test/dlpTest.php
```diff
@@ -362,7 +362,7 @@ public function testKMap()
             'data-project' => getenv('GOOGLE_PROJECT_ID'),
             'topic-id' => getenv('DLP_TOPIC'),
             'subscription-id' => getenv('DLP_SUBSCRIPTION'),
-            'region-code' => 'USA',
+            'region-code' => 'US',
```
tmatsuo: Do you know why using 'USA' made the test flaky?

bshaffer: It didn't make it flakey, it just failed, and we didn't notice because the tests were marked as flakey. 

The flakeyness was marked for another reason occurring in the DLP V1 API, so I'm removing the flakey flag to see if they will pass now.

PR #584, opened 2018-03-29
fix region tag

sirtorry:


vision/test/visionTest.php
```diff
@@ -271,24 +271,6 @@ public function testDocumentTextCommandGcs()
...
-
-    public function testDetectWebCommandGcs()
-    {
-        $this->requireCloudStorage();
-
-        $path = 'gs://' . $this->bucketName . '/landmark.jpg';
-        $output = $this->runCommand('web', $path);
-        $this->assertContains('web entities found:', $output);
-        $this->assertContains('Palace of Fine Arts Theatre', $output);
-    }
```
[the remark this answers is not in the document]
sirtorry: the 'web' command is tested in testDetectWebNoGeoCommand and testDetectWebNoGeoCommandGcs. also the palace of fine arts theatre photo no longer returns palace of fine arts theatre as a web entity...

vision/README.md
```diff
@@ -70,11 +70,3 @@ If you have not set a timezone you may get an error from php. This can be resolv
...
 [google-cloud-php-source]: https://github.com/GoogleCloudPlatform/google-cloud-php
 [google-cloud-php-issues]: https://github.com/GoogleCloudPlatform/google-cloud-php/issues
-
-## Contributing changes
-
-* See [CONTRIBUTING.md](../../CONTRIBUTING.md)
-
-## Licensing
-
-* See [LICENSE](../../LICENSE)
```
bshaffer: Let's update these to the proper path instead of removing them

sirtorry: this format now mirrors that of firestore, etc.

vision/src/detect_face.php
```diff
@@ -55,5 +62,44 @@ function detect_face($path)
         print('Bounds: ' . join(', ',$bounds) . PHP_EOL);
         print(PHP_EOL);
     }
+
+    # draw box around faces
```
bshaffer: It will be a little more readable to put these variables inside the conditional (`if ($faces && $outFile)`) below, since they are not used unless that block is executed.

This is my last and final suggestion I promise!

PR #587, opened 2018-04-03
Adds IOT create_client region tag

bshaffer:


bshaffer:
Good question, I removed the region tags because I wanted the context of the rest of the `list_repositories` sample, so it's actually just the additional comment, and not a region tag :)

PR #594, opened 2018-04-11
BQ-DTS: Add quickstart

tswast:
Add quickstart for BigQuery Data Transfer API.

This sample is intended to appear in the documentation at https://cloud.google.com/bigquery/transfer/docs/reference/libraries

### DIFF
--- a/bigquerydatatransfer/quickstart/composer.json
+++ b/bigquerydatatransfer/quickstart/composer.json
@@ -1,0 +1,9 @@
+{
+    "require": {
+        "php": ">=5.4",
+        "google/cloud-bigquerydatatransfer": "^0.3.2"
+    },
+    "require-dev": {
+        "phpunit/phpunit": "~4"
+    }
+}
--- a/bigquerydatatransfer/quickstart/composer.lock
+++ b/bigquerydatatransfer/quickstart/composer.lock
@@ -1,0 +1,1730 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "This file is @generated automatically"
+    ],
+    "content-hash": "5827250435a40648aaf4f231fcae37fa",
+    "packages": [
+        {
+            "name": "firebase/php-jwt",
+            "version": "v5.0.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/firebase/php-jwt.git",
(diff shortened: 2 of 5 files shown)


tswast:
@bshaffer Done! I'll keep an eye on Kokoro before we merge.

bshaffer:
@dpebot merge when green

bigquerydatatransfer/quickstart.php
```diff
@@ -0,0 +1,45 @@
...
+        echo 'Data source: ', $dataSource->getDisplayName(), PHP_EOL;
+        echo 'ID: ', $dataSource->getDataSourceId(), PHP_EOL;
+        echo 'Full path: ', $dataSource->getName(), PHP_EOL;
+        echo 'Description: ', $dataSource->getDescription(), PHP_EOL;
+    }
+} finally {
+    $bqdtsClient->close();
+}
+# [END bigquerydatatransfer_quickstart]
+return true;
```
bshaffer: @tswast if you look at the travis output, none of the tests actually echo anything other than `Supported Data Sources: `, which means the inside of the `for` loop is not tested.

I would suggest removing this return statement, and instead use output buffering to verify the output contains `Data source: ` or something similar (see below)

bigquerydatatransfer/test/quickstartTest.php
```diff
@@ -0,0 +1,43 @@
...
+        $contents = file_get_contents(__DIR__ . '/../quickstart.php');
+        $contents = str_replace(
+            ['YOUR_PROJECT_ID', '__DIR__'],
+            [$projectId, sprintf('"%s/.."', __DIR__)],
+            $contents
+        );
+        file_put_contents($file, $contents);
+
+        // Invoke quickstart.php
+        $this->result = include $file;
```
bshaffer: Instead, try this:

```php
ob_start();
include $file;
$result = ob_get_clean();

$this->assertContains('Data source:', $result);

bigquerydatatransfer/quickstart.php
tswast: Yeah, I actually expected empty output because there is more than one step to enable DTS on a project. https://cloud.google.com/bigquery/docs/enable-transfer-service I'll follow-up to update the sample once it's been whitelisted.

bigquerydatatransfer/test/quickstartTest.php
tswast: Thanks. Done in https://github.com/GoogleCloudPlatform/php-docs-samples/pull/597

PR #596, opened 2018-04-12
Add available db regions

miya0001:
If we run `php wordpress-helper.php setup` with the option like `--db_region=asia-northeast1`, the value of the region in `app.yml` and `wp-config.php` will be empty, because some regions are not defined.

This PR adds those regions to `wordpress-helper.php`.


tmatsuo:
@dpebot Merge when green

PR #597, opened 2018-04-12
BQ-DTS: test for actual API output

tswast:
Follow-up to https://github.com/GoogleCloudPlatform/php-docs-samples/pull/594#pullrequestreview-111801409

### DIFF
--- a/bigquerydatatransfer/test/quickstartTest.php
+++ b/bigquerydatatransfer/test/quickstartTest.php
@@ -35,6 +35,7 @@ class quickstartTest extends PHPUnit_Framework_TestCase
         file_put_contents($file, $contents);
 
         // Invoke quickstart.php and capture output
+        ob_start();
         include $file;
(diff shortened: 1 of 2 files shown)


bigquerydatatransfer/test/quickstartTest.php
```diff
@@ -34,10 +34,11 @@ public function testQuickstart()
         );
         file_put_contents($file, $contents);
 
-        // Invoke quickstart.php
-        $this->result = include $file;
+        // Invoke quickstart.php and capture output
```
bshaffer: You need to add 

```php
ob_start();
```

to start the output buffering

tswast: Ah, that must be why phpunit said the test is "risky"

tswast:
@bshaffer Okay, I think this one is ready now.

PR #598, opened 2018-04-12
Add test suite to kokoro for PHP 7.2

bshaffer:
Summary of changes:

1. Added `.kokoro` and relevant build / Docker files
2. Removed video "detect faces" sample, which is no longer supported
3. Removed BigQuery list projects sample, which is not supported
4. Consolidated Environment variables
  - Renamed `GOOGLE_BUCKET_NAME` to `GOOGLE_CLOUD_BUCKET`
  - Renamed `GOOGLE_CLOUD_PROJECT` to `GOOGLE_PROJECT_ID`
  - Removed or standardized other misc variables
5. Organized all GCS test files into subdirectories based on product / sample names.


bshaffer:
These are passing with the exception of video tests, which are waiting for the new project to be whitelisted

bshaffer:
@tmatsuo @ryanmats kokoro and travis tests now passing. PTAL!

.kokoro/system_tests.sh
```diff
@@ -1,4 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 set -ex
 
-echo "hello world"
+cd github/php-docs-samples
+
+# Unencrypt and extract secrets
+source ${KOKORO_GFILE_DIR}/secrets.sh
```
tmatsuo: I don't see the `secrets.sh`. From the comment in `secrets-example.sh`,  I assume you need to manually populate it?

How about to use KMS? something like

1. Manually populate `secrets.sh` locally
2. Use KMS to encrypt that file `secrets.sh.enc` and check that file in
3. Have decrypt command in the test runner

.kokoro/secrets-example.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# General
+export GOOGLE_APPLICATION_CREDENTIALS=$KOKORO_GFILE_DIR/service-account.json
+export GOOGLE_PROJECT_ID=
+export GOOGLE_API_KEY=
+export GOOGLE_STORAGE_BUCKET=$GOOGLE_PROJECT_ID
+export GOOGLE_CLIENT_ID=
+export GOOGLE_CLIENT_SECRET=
+export GCLOUD_PROJECT=$GOOGLE_PROJECT_ID
+export FIRESTORE_PROJECT_ID=$GOOGLE_PROJECT_ID
```
ryanmats: I'd recommend leaving it as 'FIRESTORE_PROJECT_ID=' because a user will want a separate Firestore project (because you currently cannot use both Datastore and Firestore in one GCP project).

vision/src/detect_web_with_geo_metadata_gcs.php
```diff
@@ -42,14 +42,16 @@ function detect_web_with_geo_metadata_gcs($path)
     $response = $imageAnnotator->webDetection($path, ['imageContext' => $imageContext]);
     $web = $response->getWebDetection();
 
-    if ($web->getWebEntities()) {
+    if ($web) {
```
ryanmats: Why does the  'vision/src/detect_web_gcs.php' file check 'if ($web->getWebEntities())' but this one doesn't?

bshaffer: I changed it because if `$web` is null it throws a fatal error. This may also be the case with `detect_web_gcs`, in which case it should also be changed.

bshaffer: I looked into it, and `detect_web_gcs` prints every result, regardless of if there are any. So the `if` checks in `detect_web` and `detect_web_gcs` are completely unnecessary.

.kokoro/system_tests.sh
bshaffer: Yes, that is correct. Locally you would need to populate it. I think this is a great idea, but I still want to keep `secrets-example.sh` up-to-date to keep track of all the environment variables we use in our tests!

PR #599, opened 2018-04-14
Add wp-cli.yml into the WordPress project

miya0001:
`wp-cli.yml` is a configuration file for the WP-CLI command.
If there is the file, we can run WP-CLI command without `--path` parameter under the `wordpress-project` directory.
And also, we can launch local development server with `wp server` command to browse `http://localhost:8080`.
The combination of `cloud_sql_proxy` and `wp server` is very nice for me.

Thanks!


bshaffer:
Thank you for your contribution, @miya0001. Looks good to me! Let's see what @tmatsuo has to say.

PR #604, opened 2018-04-25
[Storage] Fix comment in PHP list objects samples

frankyn:
Update comments for list object and list object with prefix samples.


frankyn:
Thanks @bshaffer!

PR #608, opened 2018-05-16
Updating all the region tags to have a prefix of datastore

billyjacobson:
[DO NOT MERGE] Updating all the region tags to have a prefix of datastore


jmdobry:
Thanks! Looks like we'll need to ping you to merge it?

PR #611, opened 2018-05-18
adds multiple php versions to Kokoro

bshaffer:


bshaffer:
Tests are failing solely because they are taking too long on travis (~50 minutes), so I am merging this anyway. I believe otherwise they're all good.

PR #612, opened 2018-05-22
Spanner Batch Query sampler2

Mdsfqx93:


Mdsfqx93:
Mdsfqx@gmail.com

bshaffer:
We already have a batch query sample here: https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/spanner/src/batch_query_data.php

PR #619, opened 2018-06-01
refactors and moves wordpress samples

bshaffer:
Depends on https://github.com/GoogleCloudPlatform/php-tools/pull/35

Moves common functions into `google/cloud-tools` and then utilizes these functions in `appengine/standard/wordpress` and `appengine/flexible/wordpress`.


tmatsuo:
I'm not sure if this is a good chnage...

The original script asks you which environment you want to use (standard/flex). Now it doesn't.
Also, this change seems to lack the ability to create an environment for standard/php55. Some users may need that to create a test environment etc.

bshaffer:
Hey @tmatsuo !
This change is so the script works for all three environments, `php55`, `php72`, and `flexible`. The script is contained inside the directory for those samples (e.g. `appengine/php55/wordpress` and `appengine/flexible/wordpress`.

It still allows you to create one for php55!

This is important because we want the `appengine/` directory to only contain the versions of appengine supported. As "wordpress" is not a supported appengine version, We are moving these samples inside the directories for each version of appengine, and will remove the `wordpress` directory (after docs are updated)

bshaffer:
Changes for appengine standard start [here](https://github.com/GoogleCloudPlatform/php-docs-samples/pull/619/files#diff-8dd868573e97e85fbb9385a71d598fae)

appengine/flexible/wordpress/files/app.yaml
```diff
@@ -2,7 +2,7 @@ runtime: php
 env: flex
 
 beta_settings:
-  cloud_sql_instances: {{db_connection}}
+  cloud_sql_instances: "YOUR_DB_CONNECTION"
```
tmatsuo: Originally, it was a twig template and the script replaces those placeholders. How does it work now?

tmatsuo: Ok, I see it's using strtr

bshaffer:
thanks @tmatsuo !

PR #624, opened 2018-06-12
appengine/php72/cloudsql added region tags

angelafunk:


bshaffer:
But since these are in [`appengine/php72/cloudsql/app.yaml`](https://github.com/angelafunk/php-docs-samples/blob/49bed639e1fe6cacbd10e0bcb567bb72f79c97a9/appengine/php72/cloudsql/app.yaml#L19) already do they really need to be in `app-postgres.yaml` as well?

angelafunk:
Yes! So that users can use the github widget and find the entire postgres app.yaml file if needed from this code snippet.

bshaffer:
That makes sense, but then we should make them distinct. So this one should be `gae_cloudsql_posgres_settings` and the other should be `gae_cloudsql_mysql_settings`

bshaffer:
@angelafunk please update the one in `app.yaml` to be `gae_cloudsql_mysql_settings`

PR #625, opened 2018-06-12
appengine/php72/cloudsql/index.php changed region tags

angelafunk:


bshaffer:
I don't think we want "example" as the region tag, as this will conflict. If the intention is to have the entire sample in one include, we should make it `gae_cloudsql_example`

angelafunk:
updated!

bshaffer:
Looks like the cs check is failing (not your fault), but can you change the final region tag to fix [this error](https://travis-ci.org/GoogleCloudPlatform/php-docs-samples/jobs/391968242#L3641)? It needs to be:

```
<?php # [END gae_cloudsql_example]?>
```

bshaffer:
Or actually since it goes to the end of the file I believe we can just leave out the region tag.

PR #626, opened 2018-06-13
Add Spanner struct parameters sample.

jsimonweb:


bshaffer:
I think you need to run `composer update` on PHP 5.6 to fix those failing tests

PR #627, opened 2018-06-15
Add DLP code samples for custom info types

mwdaub:
The DLP team is in the process of expanding code samples to include features that we have launched since the initial samples were written.

In this pull request, I updated inspect_*.php to include dictionary and regex based custom info types, along with some tests.


bshaffer:
Thank you for your PR!

We want a separate sample for this, instead of adding these to all the existing samples.

mwdaub:
Apologies for my slow responses. This was done as part of our team's fixit, so I'm slower to respond now that it is over and normal responsibilities have resumed. Please reopen this pull request, and I will address your comments as soon as I can.

PR #628, opened 2018-06-18
Adds Trace sample

bshaffer:
(and also cleans up a few other samples)

### DIFF
--- a/appengine/php72/trace/test/DeployTest.php
+++ b/appengine/php72/trace/test/DeployTest.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * Copyright 2018 Google Inc.
+ *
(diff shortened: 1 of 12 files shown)


appengine/php72/trace/test/DeployTest.php
```diff
@@ -0,0 +1,68 @@
...
+        ]);
+        $stack = HandlerStack::create();
+        $stack->push($middleware);
+
+        // create the HTTP client
+        $trace = new Client([
+            'handler' => $stack,
+            'auth' => 'google_auth',
+            'base_uri' => sprintf(
+                'https://cloudtrace.googleapis.com/v1/projects/%s/',
```
bshaffer: We have to use an authed HTTP request because the v2 API does not have a way to get traces.

PR #629, opened 2018-06-18
Update app.yaml

angelafunk:
appengine/php72/cloudsql/app.yaml changed region tag name


bshaffer:
Please add this to #624

PR #632, opened 2018-06-20
Adds KMS integration samples for Google Cloud Storage

bshaffer:


storage/src/upload_with_kms_key.php
```diff
@@ -0,0 +1,57 @@
...
+use Google\Cloud\Storage\StorageClient;
+
+/**
+ * Upload a file using KMS encryption.
+ *
+ * @param string $projectId Your Google Cloud project ID.
+ * @param string $bucketName the name of your Google Cloud bucket.
+ * @param string $objectName the name of the object.
+ * @param string $source the path to the file to upload.
+ * @param string $kmsKeyName KMS key ID to use as the default KMS key.
```
frankyn: KMS key ID to use to encrypt object server side.

storage/storage.php
```diff
@@ -336,6 +336,48 @@
...
+
+<info>php %command.full_name% --help</info>
+
+EOF
+    )
+    ->addArgument('project', InputArgument::REQUIRED, 'Your billable Google Cloud Project ID')
+    ->addArgument('bucket', InputArgument::REQUIRED, 'The Cloud Storage bucket name')
+    ->addArgument('object', InputArgument::REQUIRED, 'The Cloud Storage bucket name')
+    ->addArgument('upload-from', InputArgument::REQUIRED, 'Path to the file to upload')
+    ->addArgument('kms-key-name', InputArgument::REQUIRED, 'KMS key ID to use as the default KMS key.')
```
frankyn: not default kms key.

storage/src/upload_with_kms_key.php
bshaffer: done!

storage/storage.php
bshaffer: done!

storage/src/enable_default_kms_key.php
```diff
@@ -0,0 +1,53 @@
...
+
+# [START enable_default_kms_key]
+use Google\Cloud\Storage\StorageClient;
+
+/**
+ * Enable a bucket's requesterpays metadata.
+ *
+ * @param string $projectId Your Google Cloud project ID.
+ * @param string $bucketName Name of your Google Cloud Storage bucket.
+ * @param string $kmsKeyName KMS key ID to use to encrypt objects server side.
```
frankyn: Keep `KMS key ID to use as the default KMS key.` for bucket default kms key name

PR #636, opened 2018-07-12
Check in Cloud Job Discovery sample code

michael54:
Cloud Job Discovery got Public beta launch so we would like to have our samples checked in as other projects.


xinyunh0929:
Hi reviewers,

It seems your credential file isn't available. May you check it?

jobs/cjd_sample/composer.json
```diff
@@ -0,0 +1,29 @@
...
+        "google/apiclient": "^2.1",
+        "google/apiclient-services": "^0.64",
+        "symfony/console": " ^3.0"
+    },
+    "require-dev": {
+        "phpunit/phpunit": "~4.0"
+    },
+    "autoload": {
+        "psr-4": {
+            "Google\\Cloud\\Samples\\Language\\": "src/"
```
bshaffer: You can remove this section, as no classes are being autoloaded using PSR-4. All of this is handled by "files" below.

jobs/cjd_sample/src/auto_complete_sample.php
```diff
@@ -0,0 +1,115 @@
...
+            'languageCode' => 'en-US',
+            'pageSize' => 10);
+        if (isset($companyName)) {
+            $optParams['companyName'] = $companyName;
+        }
+
+        $jobService = JobServiceConnector::get_job_service();
+        $results = $jobService->v2->complete($optParams);
+
+        var_export($results);
```
bshaffer: why `var_dump`?

jobs/cjd_sample/composer.json
```diff
@@ -0,0 +1,29 @@
+{
+    "require": {
+        "google/apiclient": "^2.1",
+        "google/apiclient-services": "^0.64",
```
bshaffer: This line can be removed, as the `apiclient-services` dependency should be handled implicitly in `google/apiclient`. Also, this line will KEEP the dependency at `0.64` and not allow it to go to `0.65` (because karets don't upgrade minor versions for pre-1.0 releases). This isn't good, since this API and other APIs are constantly getting updated.

jobs/cjd_sample/composer.json
michael54: Thanks for review! I removed this.

jobs/cjd_sample/composer.json
michael54: Done.

jobs/cjd_sample/src/auto_complete_sample.php
michael54: I'm not sure what are you asking? Are you asking why I'm using var_export instead of var_dump? Or are you asking why I'm using var_export here? I just want to print out the response here, that's it.

michael54:
Thanks for reviewing this. I have moved all code under jobs/ directly.

jobs/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Build and Run
+1.  **Enable APIs** - [Enable the Job Discovery API](https://console.cloud.google.com/flows/enableapi?apiid=jobs.googleapis.com)
+    and create a new project or select an existing project.
+2.  **Activate your Credentials** - If you do not already have an active set of credentials, create and download a [JSON Service Account key](https://pantheon.corp.google.com/apis/credentials/serviceaccountkey). Set the environment variable `GOOGLE_APPLICATION_CREDENTIALS` as the path to the downloaded JSON file.
+4.  **Clone the repo** and cd into this directory
+
+    ```
+    $ git clone https://github.com/GoogleCloudPlatform/php-docs-samples
+    $ cd php-docs-samples/jobs/cjd_sample
```
bshaffer: update to new path

jobs/cjd_sample.php
```diff
@@ -0,0 +1,43 @@
+<?php
```
bshaffer: rename this file to `jobs.php`

jobs/test/AutoCompleteSampleTest.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    private $commandTester;
+
+    public function setUp()
+    {
+        if (!getenv('GOOGLE_APPLICATION_CREDENTIALS')) {
+            return $this->markTestSkipped("Set the GOOGLE_APPLICATION_CREDENTIALS environment variable.");
+        }
+
+        $application = require __DIR__ . '/../cjd_sample.php';
```
bshaffer: update all these to `'/../jobs.php'`

jobs/src/job_service_connector.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * Gets Google_Service_JobService.
+     *
+     * @return Google_Service_JobService
+     */
+    public static function get_job_service()
+    {
+        if (!isset($jobService)) {
```
bshaffer: should be `!isset(self::$jobService)`

jobs/src/basic_company_sample.php
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * Updates a company.
+     *
+     * @param string $companyName
+     * @param Google_Service_JobService_Company $companyToBeUpdated
+     * @return Google_Service_JobService_Company
+     */
+    public static function update_company(string $companyName, Google_Service_JobService_Company $companyToBeUpdated)
+    {
+        $jobService = JobServiceConnector::get_job_service();
```
bshaffer: I understand the concept of DRY, but since these are samples, I think we should explicitly create the job service or at least have it be `self::get_job_service` so that the users don't have to travel too far to find how the job service is created.

michael54: Done. It's repeated it in each sample now.

jobs/src/job_service_connector.php
michael54: Done.

jobs/test/AutoCompleteSampleTest.php
michael54: Done.

jobs/cjd_sample.php
michael54: Done.

jobs/README.md
michael54: Done.

bshaffer:
All my tests passed locally for this, but I had to make this compatible for PHP 5.6 (see ac154c54c87b2c77dcf8ac728ae821822ec2505b). If these pass in #644 I will merge that PR.

PR #646, opened 2018-07-30
Changed region tag syntax

willgoldby:


bshaffer:
I think these comments are invalid and will throw an error in twig or will be output as HTML

willgoldby:
Okay, that makes sense. Was hoping to use them as region tags. My bad.

PR #651, opened 2018-07-31
Changed GOOGLE_BUCKET_NAME to GOOGLE_STORAGE_BUCKET

willgoldby:
I'm having users pick their own database names so thought removing the comment made sense.

### DIFF
--- a/appengine/php72/getting-started/app.yaml
+++ b/appengine/php72/getting-started/app.yaml
@@ -2,10 +2,9 @@ runtime: php72
 instance_class: F2
 
 env_variables:
-    GOOGLE_BUCKET_NAME: ""
+    GOOGLE_STORAGE_BUCKET: ""
     # populate these to use the "mysql" or "postres" backends
     CLOUDSQL_CONNECTION_NAME: ""
     CLOUDSQL_USER: ""
     CLOUDSQL_PASSWORD: ""
-    ## Uncomment to give your database a name other than "bookshelf"
-    # CLOUDSQL_DATABASE_NAME: ""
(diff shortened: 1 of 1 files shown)


appengine/php72/getting-started/app.yaml
```diff
@@ -2,10 +2,9 @@ runtime: php72
 instance_class: F2
 
 env_variables:
-    GOOGLE_BUCKET_NAME: ""
+    GOOGLE_STORAGE_BUCKET: ""
     # populate these to use the "mysql" or "postres" backends
     CLOUDSQL_CONNECTION_NAME: ""
     CLOUDSQL_USER: ""
     CLOUDSQL_PASSWORD: ""
-    ## Uncomment to give your database a name other than "bookshelf"
```
bshaffer: Why did this comment get removed? And also the `CLOUDSQL_DATABASE_NAME` should remain commented out, as it defaults to `bookshelf` so it isn't required.

willgoldby: In the tutorial, users create a new database and name it, so I was thinking all users would need to uncomment it out... What are your thoughts on that?

bshaffer: Well, I wanted to keep it simple and have it default to "bookshelf". The tutorial could suggest this, or another name like `getting_started_php` if you prefer, but I think it's nice to keep it optional.

willgoldby: Sounds good to me. I'll just make a note in the tutorial. :)

PR #652, opened 2018-07-31
Added region tags to show how to write to the database.

willgoldby:


appengine/php72/getting-started/src/CloudSqlDataModel.php
```diff
@@ -113,14 +113,15 @@ public function create($book, $id = null)
         $placeHolders = array_map(function ($key) {
             return ":$key";
         }, $names);
+        [START sql_write]
```
bshaffer: These both need to be comments!  :)

Also, let's make them `cloudsql_write` for consistency.

PR #656, opened 2018-08-02
Correct upload url for createUploadUrl method

levidurfee:
It later refers to the URL being `/upload_handler.php`, not `/upload/handler`

> in the example above, this is /upload_handler.php


bshaffer:
Fixed in 8b903414e7b017f5d79f2111ffa83c44f8b4fcd0

PR #659, opened 2018-08-08
adds slim framework example

bshaffer:


appengine/php72/slim-framework/.gcloudignore
```diff
@@ -0,0 +1,17 @@
+# This file specifies files that are *not* uploaded to Google Cloud Platform
+# using gcloud. It follows the same syntax as .gitignore, with the addition of
+# "#!include" directives (which insert the entries of the given .gitignore-style
```
grayside: Need to read this parenthetical a couple times and know what "include" tends to mean to follow. Read the topic and while the line in place pulls directly from it, I think it lacks context. Maybe something like:

> [...] "#include:<filename>" directives (which insert the ignore entries from a separate file into this one)

Also, is .gcloudignore recursively additive like .gitignore as new files are found in sub-directories? That's not clear from the topic help :)

appengine/php72/slim-framework/README.md
```diff
@@ -0,0 +1,43 @@
+# Slim Example for App Engine for PHP 7.2
+
+This sample demonstrates how to deploy a *very* basic [Slim][slim] application to
+[Google App Engine for PHP 7.2][appengine-php].
```
grayside: In cases where the README is truncated (such as the project landing on GH mobile) will these links be broken?

appengine/php72/slim-framework/app.yaml
```diff
@@ -0,0 +1,2 @@
+runtime: php72
+
```
grayside: Per previous discussion, do we need to pre-populate favicons and some file routing fallbacks?

appengine/php72/slim-framework/index.php
```diff
@@ -0,0 +1,31 @@
...
+ * This front controller is called by the App Engine web server to handle all
+ * incoming requests. To change this, you will need to modify the "entrypoint"
+ * directive in `app.yaml`.
+ * @see https://cloud.google.com/appengine/docs/standard/php/config/appref
+ */
+require 'vendor/autoload.php';
+
+$app = new Slim\App();
+$app->get('/', function ($request, $response) {
+    // Use the Null Coalesce Operator in PHP7
```
grayside: Nice callout! Can we add a link to http://php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce?

appengine/php72/slim-framework/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+The last command will open `https://{YOUR_PROJECT_ID}.appspot.com/`
+in your browser.
+
+## Application Components
+
+The application consists of three components:
+
+ 1. An [`app.yaml`](app.yaml) which sets your application runtime to be `php72`.
+ 2. A [`composer.json`](composer.json) which declares your application's dependencies
```
grayside: Nit: TerminatingPunctuationInconsistencyWarning

appengine/php72/slim-framework/composer.json
```diff
@@ -0,0 +1,9 @@
+{
+    "require": {
+        "slim/slim": " ^3.0"
+    },
+    "require-dev": {
+        "google/cloud-tools": "^0.8.1",
+        "phpunit/phpunit": "^5"
```
grayside: Noting the current release of phpunit is 7.3.1

appengine/php72/slim-framework/.gcloudignore
bshaffer: Good questions. This file is autogenerated though, so I do not want to change it until it's changed at the source!

I agree it's confusing, I'd file an internal bug (or submit a CL) concerning it.

appengine/php72/slim-framework/composer.json
bshaffer: yes, unfortunately our samples run against PHP 5.6 (since that's still supported, through 2018). I would really like to rewrite our tests so that we can add version variance (e.g. if composer marks them with a required PHP version, the test is skipped). Until then, or until we deprecate the samples for PHP 5.6, we're stuck with phpunit 5 for all samples.

appengine/php72/slim-framework/README.md
bshaffer: I believe the README is still parsed in its entirety and truncated by the browser?

appengine/php72/slim-framework/app.yaml
bshaffer: probably a good idea, although I really want this sample to focus on simplicity

appengine/php72/slim-framework/README.md
bshaffer: woops

appengine/php72/slim-framework/app.yaml
grayside: It seems like a gotcha that will need to be ubiquitous or have really good SEO. Okay with dropping it shy of a more generalized decision on how to handle.

PR #668, opened 2018-08-10
Since F1 works, I was thinking it'd be simpler not to use F2? Thoughts?

willgoldby:


bshaffer:
I already removed them :D 

https://github.com/GoogleCloudPlatform/php-docs-samples/pull/663/files

PR #669, opened 2018-08-10
[KMS] Remove global location from kms example to generalize for GCS

frankyn:
Hi @bshaffer,

I'm removing global location mention to generalize sample for GCS CMEK documentation. The sample is used to show developers how to set an IAM policy on a KMS key, but global location aren't supported so I want to generalize that part of the sample.

PTAL, thank you!


frankyn:
Failures don't seem related, @bshaffer can you take a look when you have a moment?

PR #670, opened 2018-08-13
update region tags

averikitsch:
Updating and removing region tags only


bshaffer:
Please remove the tag when it's ready to merge!

appengine/flexible/sendgrid/app.yaml
```diff
@@ -5,7 +5,7 @@ runtime_config:
   document_root: .
 
 # [START gae_flex_sendgrid_yaml]
-gae_flex_sendgrid_yaml:
+env_variables:
```
bshaffer: that was a close one :D

PR #673, opened 2018-08-14
Update region tags for consistency

nnegrey:


nnegrey:
Updated some more region tags. PTAL
Should be good to go now.

nnegrey:
@bshaffer Any ideas why the tests might be failing?

bshaffer:
The error states: `"Errors in 'fake_display_name_for_testing_...' entity: Name length is greater than 30 chars.",`

That seems pretty clear. Although I thought we fixed this already. But we just need to shorten the entity name length in the tests. Pretty awesome that Dialogflow decided to break BC like that.

nnegrey:
@bshaffer Got it working

PR #674, opened 2018-08-14
adds front controller samples

bshaffer:


appengine/php72/front-controller/index.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+// Static list provides security against URL injection by default.
+# [START gae_simple_front_controller]
```
grayside: It may be worth a comment cross-linking github or docs for https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/php72/grpc/index.php to see a regex example.

appengine/php72/front-controller/homepage.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+/**
+ * A list of routes for our homepage. This is really only here to demonstrate
+ * that PHP is being executed.
+ */
+$routes = ['contact.php'];
```
grayside: It's a bit weird to see a hard-wired routes array here in a separate file from front controller itself. It's a nit-pick, but I might call it $links or $menu to clarify that it's not used for routing.

appengine/php72/front-controller/homepage.php
```diff
@@ -0,0 +1,17 @@
...
+?>
+
+<h1>Welcome to the Homepage!</h1>
+
+<ul>
+    <?php foreach ($routes as $route): ?>
```
grayside: I haven't seen the colon/endforeach syntax since PHP4... is it still typically used?

bshaffer: for rendering in HTML, yes. 
But is rendering PHP in HTML used? ...yes 😄

appengine/php72/front-controller/homepage.php
bshaffer: ahh that's a good call, will change it.

appengine/php72/front-controller/index.php
bshaffer: Agreed. I'll add the links above the region tags, since the docs themselves call out some other options.

bshaffer:
@grayside ready for another look!

angeloko23:
php72-frontcontroller

PR #676, opened 2018-08-14
Update README.md

sirtorry:


bshaffer:
Hmm, looks like dialogflow tests are legitimately broken.
```
'fake_display_name_for_testing_...' entity: Name length is greater than 30 chars.",
```
We can fix in a different PR

PR #677, opened 2018-08-15
Vision region tag update

alixhami:
Updates region tags to be consistent with those in other languages. Region tag update on cloud.google.com is also in progress and will be submitted after all language PRs are merged.

### DIFF
--- a/vision/src/detect_face.php
+++ b/vision/src/detect_face.php
@@ -17,10 +17,9 @@
 // [START vision_face_detection]
 namespace Google\Cloud\Samples\Vision;
 
-// [START vision_face_detection_tutorial_client]
+// [START vision_face_detection_tutorial_imports]
 use Google\Cloud\Vision\V1\ImageAnnotatorClient;
-
-// [END vision_face_detection_tutorial_client]
+// [END vision_face_detection_tutorial_imports]
 
 function detect_face($path, $outFile = null)
 {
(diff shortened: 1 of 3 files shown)


vision/src/detect_face.php
```diff
@@ -97,9 +96,12 @@ function detect_face($path, $outFile = null)
...
             }
         }
-        # [END highlight_image]
+        # [END vision_face_detection_tutorial_process_response]
+        # [START vision_face_detection_tutorial_run_application]
         call_user_func($imageWriteFunc[$ext], $outputImage, $outFile);
         printf('Output image written to %s' . PHP_EOL, $outFile);
+        # [END vision_face_detection_tutorial_run_application]
     }
+    // [START vision_face_detection]
```
bshaffer: I believe this one should be deleted, as the `vision_face_detection` region is closed at line 62.

vision/src/detect_face.php
```diff
@@ -97,9 +96,12 @@ function detect_face($path, $outFile = null)
...
 }
 // [END vision_face_detection]
```
bshaffer: Same with this one, should be removed

vision/src/detect_face.php
alixhami: It looks a little weird, but I did this so the closing curly braces would render with the sample. If you start and end a region tag multiple times on the same file, it will add all the pieces together.

bshaffer:
Woot!

PR #691, opened 2018-08-23
Update Region Tags

averikitsch:
Ongoing GAE region tag updates
*Docs are frozen at last commit*

### DIFF
--- a/appengine/php55/memcache/app.php
+++ b/appengine/php55/memcache/app.php
@@ -38,7 +38,7 @@ $app->get('/', function (Application $app, Request $request) {
 $app->post('/', function (Application $app, Request $request) {
     /** @var Twig_Environment $twig */
     $twig = $app['twig'];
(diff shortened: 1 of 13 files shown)


appengine/php55/memcache/app.php
```diff
@@ -38,46 +38,31 @@
...
-    # [END memcached_get]
-});
-
-$app->put('/memcached/{key}', function ($key, Request $request) {
-    # [START memcached_put]
-    $memcache = new Memcached;
-    $value = $request->getContent();
-    return $memcache->set($key, $value);
-    # [END memcached_put]
-});
```
bshaffer: Why does this PR remove `memcached` samples? Was this intentional?

averikitsch: This looked like a copy/paste of the exact code above - will revert

PR #707, opened 2018-09-14
TTS GA

sirtorry:


texttospeech/README.md
```diff
@@ -0,0 +1,86 @@
+# Cloud Text-to-Speech: PHP Samples
+
+[![Open in Cloud Shell][shell_img]][shell_link]
+
+[shell_img]: http://gstatic.com/cloudssh/images/open-btn.png
```
bshaffer: I'm so tired of this damn blurry image... use this one instead: `http://gstatic.com/cloudssh/images/open-btn.svg`

texttospeech/README.md
```diff
@@ -0,0 +1,86 @@
...
+### Authentication
+
+This sample requires you to have authentication setup. Refer to the [Authentication Getting Started Guide](https://cloud.google.com/docs/authentication/getting-started) for instructions on setting up credentials for applications.
+
+## Install Dependencies
+
+1. [Enable the Cloud Text-to-Speech API](https://console.cloud.google.com/flows/enableapi?apiid=texttospeech.googleapis.com).
+
+1. **Install dependencies** via [Composer](http://getcomposer.org/doc/00-intro.md).
+    Run `php composer.phar install` (if composer is installed locally) or `composer install`
```
bshaffer: Just leave `php composer.phar install` out and say to run `composer install` for two reasons: 1. `composer` is installed globally on Cloud Shell and 2. It's consistent with (most of) our other samples.

texttospeech/README.md
```diff
@@ -0,0 +1,86 @@
+# Cloud Text-to-Speech: PHP Samples
+
+[![Open in Cloud Shell][shell_img]][shell_link]
+
+[shell_img]: http://gstatic.com/cloudssh/images/open-btn.png
+[shell_link]: https://console.cloud.google.com/cloudshell/open?git_repo=https://github.com/googlecloudplatform/php-docs-samples&page=editor&working_dir=texttospeech
```
bshaffer: Not crazy about having this sample open in cloudshell since there's no way to play the resulting audio. Thoughts?

texttospeech/composer.json
```diff
@@ -0,0 +1,21 @@
+{
+    "require": {
+        "google/cloud-text-to-speech": "^0.1",
+        "symfony/console": "^3.0"
+    },
+    "require-dev": {
+        "phpunit/phpunit": "~4.8"
+    },
+    "autoload": {
+        "psr-4": {
```
bshaffer: You can get rid of this block... no classes are being loaded using `psr-4`

texttospeech/test/quickstartTest.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+class quickstartTest extends PHPUnit_Framework_TestCase
+{
+    public function testQuickstart()
+    {
+        $file = sys_get_temp_dir() . '/tts_quickstart.php';
+        $contents = file_get_contents(__DIR__ . '/../quickstart.php');
+        $contents = str_replace(
+            ['__DIR__'],
```
bshaffer: nit but no need to use arrays here since there's only one item

```php
$contents = str_replace('__DIR__', sprintf('"%s/.."', __DIR__), $contents)
```

bshaffer:
One thing I forgot to mention - delete that `texttospeech/src/audio_profile.php` file

texttospeech/README.md
sirtorry: lol wheeze

texttospeech/README.md
sirtorry: I agree running on cloud shell isn't the best idea. open to anything

texttospeech/README.md
```diff
@@ -21,7 +21,7 @@ This sample requires you to have authentication setup. Refer to the [Authenticat
 1. [Enable the Cloud Text-to-Speech API](https://console.cloud.google.com/flows/enableapi?apiid=texttospeech.googleapis.com).
 
 1. **Install dependencies** via [Composer](http://getcomposer.org/doc/00-intro.md).
-    Run `php composer.phar install` (if composer is installed locally) or `composer install`
+    Run `composer install` (if composer is installed locally) or `composer install`
```
bshaffer: This line doesn't make sense, you have `composer install` there twice. Just say ``Run `composer install` (this assumes composer is installed [globally](https://getcomposer.org/doc/00-intro.md#globally))``

PR #708, opened 2018-09-14
Refactor BigQuery sample tests

alixhami:
Tests can now be run without relying on the test runner having certain files in GCS and tables in BigQuery. The temporary dataset and tables are now generated during the test run and the GCS files are in the cloud samples bucket, which is consistent with canonical.

### DIFF
--- a/bigquery/api/test/bigqueryTest.php
+++ b/bigquery/api/test/bigqueryTest.php
@@ -242,14 +242,14 @@ class FunctionsTest extends TestCase
     public function testListDatasets()
     {
         $output = $this->runSnippet('list_datasets');
-        $this->assertContains('test_dataset', $output);
+        $this->assertContains(self::$datasetId, $output);
     }
 
     public function testListTables()
     {
-        $this->createTempEmptyTable();
+        $tempTableId = $this->createTempEmptyTable();
         $output = $this->runSnippet('list_tables', [self::$datasetId]);
-        $this->assertContains('test_table', $output);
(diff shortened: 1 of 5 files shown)


bigquery/api/snippets/import_from_file.php
```diff
@@ -38,7 +38,7 @@
 // $projectId  = 'The Google project ID';
 // $datasetId  = 'The BigQuery dataset ID';
 // $tableId    = 'The BigQuery table ID';
-// $source     = 'The path to the source file to import';
+// $source     = 'The path to the CSV source file to import';
```
bshaffer: Should we rename this to `import_from_csv_file.php`?

bigquery/api/test/bigqueryTest.php
```diff
@@ -249,13 +247,14 @@ public function testListDatasets()
 
     public function testListTables()
     {
+        $this->createTempEmptyTable();
```
bshaffer: Shouldn't this return `$tableId` and that should be used below in `assertContains` instead of `test_table`?

bigquery/api/snippets/import_from_file.php
alixhami: Alternatively, I could rename it to `import_from_local_csv.php` to distinguish it from `import_from_storage_csv.php` and keep the same naming convention. Let me know which you prefer.

bshaffer:
AMAZING

PR #709, opened 2018-09-17
Adds php72 logging sample

bshaffer:
Oh, what the heck, why not a logging sample?

### DIFF
--- a/appengine/php72/logging/index.php
+++ b/appengine/php72/logging/index.php
@@ -22,7 +22,7 @@ use Google\Cloud\Logging\LoggingClient;
 $logging = new LoggingClient();
 
 // Create a PSR-3-Compatible logger
-$logger = $logging->psrLogger('app');
+$logger = $logging->psrLogger('app', ['batchEnabled' => true]);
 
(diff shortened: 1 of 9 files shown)


appengine/php72/logging/index.php
```diff
@@ -0,0 +1,34 @@
...
+// Create the logging client
+$logging = new LoggingClient();
+
+// Create a PSR-3-Compatible logger
+$logger = $logging->psrLogger('app');
+
+// Log messages with varying log levels.
+$logger->info('This will show up as log level INFO');
+$logger->warning('This will show up as log level WARNING');
+$logger->error('This will show up as log level ERROR');
```
tmatsuo: This sample issues 3 API calls and those latency adds up. You can use `batchEnabled` option to bundle 3 logs into 1 API call. I think we should do that.

bshaffer: Great idea! I could also call `$logging->psrBatchLogger`

bshaffer: I went with the original suggestion because it's more explicit

PR #715, opened 2018-09-22
Copied config file from endpoints-samples/k8s

navinger:
Copied the k8s config file used in the Endpoints Kubernetes tutorial from the endpoints-samples repo. This way the sample code and the config files are all in the same repo.


bshaffer:
Hey this is great, but I think it would be better to put a README in `endpoints/k8s` with a link to the `endpoints-samples/k8s` config instead of having the config file duplicated.

bshaffer:
closing this as I do not believe it is relevant any longer

PR #717, opened 2018-09-24
use same region tag as others

sirtorry:
please merge. thanks.


sirtorry:
test fail not caused by PR.

PR #730, opened 2018-10-05
brace variable

flashadvocate:
This may be a matter of preference, and I hate to utter the words "good practice", but this may fall into a grey area. :)


bshaffer:
I do think this is a matter of practice, unless there's some source that says otherwise.

PR #732, opened 2018-10-08
Fix typo in cloudsql cli argument

nathang21:
Switch "host" from a flag to an argument as expected by gcloud SK 195.0.0


bshaffer:
Try upgrading your gcloud version. The current version (226.0.0) expects the `--host` flag.

```sh
$ gcloud sql users set-password root foo
ERROR: (gcloud.sql.users.set-password) Positional argument deprecated_host has been removed. Use --host instead.
```

PR #738, opened 2018-10-15
Fix Google Storage broken link

badah:
Fix the Google Storage Console link. The "www" part does not work, it's broken.


bshaffer:
Thank you!

PR #739, opened 2018-10-15
[DO NOT MERGE] Propose new sample format

alixhami:
Proposed template for auto generated Python samples.
See more details in the [Node.js PR](https://github.com/googleapis/nodejs-dlp/pull/154) description.
This PHP-specific details are based on the layout of the current [BigQuery samples](https://github.com/GoogleCloudPlatform/php-docs-samples/tree/dec932a96652fc365c6a4b5ba756427890d9ac89/bigquery/api), which were recently refactored by Brent Shaffer.


dlp/src/dlp_inspect_image_file.php
```diff
@@ -0,0 +1,84 @@
...
+/** Uncomment and populate these variables in your code */
+// $projectId = 'YOUR_PROJECT_ID';
+// $filepath = 'path/to/image.png';
+
+// Instantiate a client.
+$dlp = new DlpServiceClient();
+
+// Get the bytes of the file
+$fileBytes = (new ByteContentItem())
+    ->setType(3)
```
grayside: What's `3`? Could this be a commented constant?

dlp/src/dlp_inspect_image_file.php
```diff
@@ -0,0 +1,84 @@
...
+    ->setData(file_get_contents($filepath));
+
+// Construct request
+$parent = $dlp->projectName($projectId);
+$item = (new ContentItem())
+    ->setByteItem($fileBytes);
+$inspectConfig = (new InspectConfig())
+    // The infoTypes of information to match
```
grayside: If this comment is worth having, it seems worth expanding to explain what "matching" means.

dlp/src/dlp_inspect_image_file.php
```diff
@@ -0,0 +1,84 @@
...
+// $projectId = 'YOUR_PROJECT_ID';
+// $filepath = 'path/to/image.png';
+
+// Instantiate a client.
+$dlp = new DlpServiceClient();
+
+// Get the bytes of the file
+$fileBytes = (new ByteContentItem())
+    ->setType(3)
+    ->setData(file_get_contents($filepath));
```
grayside: file_get_contents() returns explicit `false` if it could not load data, and will also toss a warning-level error under various circumstances. Will setData() choke in that case?

dlp/test/data/test.txt
```diff
@@ -1 +1 @@
-Robert Jordan wrote the Wheel of Time.
\ No newline at end of file
+My phone number is (223) 456-7890 and my email address is gary@somedomain.com.
```
grayside: example.com please

dlp/src/dlp_inspect_image_file.php
alixhami: Is there a better way to do this other than explaining what the number is? Here are the type definitions in the [library source code](https://github.com/googleapis/google-cloud-php/blob/master/Dlp/src/V2/ByteContentItem/BytesType.php)
The other language samples are able to use the named constants instead of the number.

dlp/composer.json
```diff
@@ -11,8 +11,6 @@
             "src/inspect_datastore.php",
```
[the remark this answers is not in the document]
grayside: If we aren't building a CLI application, seems like we can drop one of the dependencies. Run `composer remove symfony/console` and there should be changes to both composer.json and composer.lock which should be committed.

dlp/src/dlp_inspect_image_file.php
grayside: You already have `use Google\Cloud\Dlp\V2\ByteContentItem;` so it should be something like `ByteContentItem::IMAGE_PNG`.

Note that this seems a bit off to me, normally use statements need to target classes, while namespaces target directories, that could be a bug I missed before.

Matching the order between the client library and the reference docs seems like a bug to file in https://github.com/googleapis/google-cloud-php?

dlp/composer.json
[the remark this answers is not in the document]
alixhami: Ran the command and committed - https://github.com/GoogleCloudPlatform/php-docs-samples/pull/739/commits/be91de204136ddee1b3980952d0624a8d9d5b6c8
Changes to composer.lock don't show up because they're in the gitignore

dlp/src/dlp_inspect_image_file.php
alixhami: Updated to use the named constant so the sample is more clear

dlp/src/dlp_inspect_image_file.php
alixhami: Here's the error I get for these cases:
+ Setting filepath to an empty file with a fake format
+ Replacing the line with `->setData(false);`
```
Fatal error: Uncaught Google\ApiCore\ApiException: {
    "message": "ContentItem has no data to inspect.",
    "code": 3,
    "status": "INVALID_ARGUMENT",
    "details": [
        {
            "@type": 0,
            "data": "type.googleapis.com\/google.rpc.DebugInfo"
        },
        {
            "@type": 0,
            "data": "[ORIGINAL ERROR] generic::invalid_argument: com.google.cloud.privacy.dlp.common.exceptions.InvalidInputException: ContentItem has no data to inspect. [google.rpc.error_details_ext] { message: \"ContentItem has no data to inspect.\" }"
        }
    ]
}

  thrown in /Users/ajhamilton/php-docs-samples/dlp/vendor/google/gax/src/ApiCore/ApiException.php on line 138
```
It appears `setData()` will allow this and then let you get an API error. This error seems fine to me, but let me know if you think this needs more handling.

bshaffer:
Hi @alixhami ! 
I know we talked about this already, but what are the next steps here? I could write a doc up on what the canonicals should look like, and then we can begin refactoring. Does that sound right?

alixhami:
This project wasn't meant to require a full refactoring of existing samples. I've been using the canonical examples if samples need to be edited anyway, so I apply the new changes. I'll send you a link to the internal page where the per-language guidance lives.

bshaffer:
merged in #880

PR #745, opened 2018-10-30
[detect_document_text] Update to new Vision client

gogasca:
Move to new VisionClient library.


gogasca:
Use Google\Cloud\Vision\VisionClient

sirtorry:
let's hold on merging this.
would be preferable if @GoogleCloudPlatform/yoshi fixes the 0/null/empty array coordinate issue.

PR #750, opened 2018-11-12
[detect_face] Update to new Vision Client

gogasca:
Update to new Vision Client
Implemented workaround. (Handle issue from Library where X/Y are not present are Null instead of 0)

### DIFF
--- a/vision/src/detect_face.php
+++ b/vision/src/detect_face.php
@@ -53,14 +53,10 @@ function detect_face($path, $outFile = null)
             $bounds = [];
             foreach ($vertices as $vertex) {
                 # get (x, y) coordinates if available.
-                $x = $y = 0;
-                if (isset($vertex['x'])) {
-                    $x = $vertex['x'];
-                }
-                if (isset($vertex['y'])) {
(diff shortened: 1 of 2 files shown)


sirtorry:
let's hold on merging this until workaround isn't required

vision/src/detect_face.php
```diff
@@ -18,48 +18,58 @@
...
+            foreach ($vertices as $vertex) {
+                # get (x, y) coordinates if available.
+                $x = $y = 0;
+                if (isset($vertex['x'])) {
+                    $x = $vertex['x'];
+                }
+                if (isset($vertex['y'])) {
+                    $y = $vertex['y'];
+                }
+                $bounds[] = sprintf('(%d,%d)', $x, $y);
```
bshaffer: This would be more concise and elegant as:

```php
$bounds[] = sprintf('(%d,%d)',
    isset($vertex['x']) ? $vertex['x'] : 0,
    isset($vertex['y']) ? $vertex['y'] : 0
);
```

PR #771, opened 2018-12-04
BigQuery: Fix extract table sample

alixhami:
This PR fixes the extract table sample to properly set the file format. Previously, the format was being improperly set, which resulted in defaulting to 'CSV' every time. This was identified by an internal bug report. The sample now also follows the [python canonical sample](https://github.com/googleapis/google-cloud-python/blob/9908ed55bd9940852ca410a24a568c2c5c2d4f71/bigquery/docs/snippets.py#L1984) more closely, though the PHP library [cannot yet extract a public table](https://github.com/googleapis/google-cloud-php/issues/1512).

### DIFF
--- a/bigquery/api/src/extract_table.php
+++ b/bigquery/api/src/extract_table.php
@@ -28,7 +28,7 @@ if (count($argv) != 5) {
     return printf("Usage: php %s PROJECT_ID DATASET_ID TABLE_ID BUCKET_NAME\n", __FILE__);
 }
 
-list($_, $projectId, $datasetId, $tableId, $bucketName, $objectName) = $argv;
+list($_, $projectId, $datasetId, $tableId, $bucketName) = $argv;
 
 # [START bigquery_extract_table]
 use Google\Cloud\BigQuery\BigQueryClient;
--- a/bigquery/api/src/extract_table.php
+++ b/bigquery/api/src/extract_table.php
@@ -24,54 +24,32 @@
 // Include Google Cloud dependendencies using Composer
 require_once __DIR__ . '/../vendor/autoload.php';
 
-if (count($argv) < 6 || count($argv) > 7) {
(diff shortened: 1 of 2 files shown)


bigquery/api/src/extract_table.php
```diff
@@ -24,54 +24,32 @@
 // Include Google Cloud dependendencies using Composer
 require_once __DIR__ . '/../vendor/autoload.php';
 
-if (count($argv) < 6 || count($argv) > 7) {
-    return printf("Usage: php %s PROJECT_ID DATASET_ID TABLE_ID BUCKET_NAME OBJECT_NAME [FORMAT]\n", __FILE__);
+if (count($argv) != 5) {
+    return printf("Usage: php %s PROJECT_ID DATASET_ID TABLE_ID BUCKET_NAME\n", __FILE__);
 }
 
 list($_, $projectId, $datasetId, $tableId, $bucketName, $objectName) = $argv;
```
bshaffer: `$objectName` should be removed

bshaffer: Also this line is causing the kokoro tests to fail.

alixhami: Thanks for the catch!

alixhami:
I'm not sure why all that logic was in there to poll the job; maybe a library update? The current reference docs for [runJob](https://googleapis.github.io/google-cloud-php/#/docs/google-cloud/v0.87.0/bigquery/table?method=runJob) specify that it waits for the job to complete.

bshaffer:
Cool, then a side-note is we should remove all the job logic on the other samples, except maybe one to illustrate how to use `startJob`

alixhami:
I'll update them to follow canonical and the sample rubric as they require updates. If I have time, though, I might do a big update again. The async job stuff should probably have its own page in the docs if it's going to be used in a sample. I wouldn't want to arbitrarily set the copy job sample, for example, to be the async sample because I'd be worried (a) people would assume async is necessary for copy jobs specifically or (b) they won't know to look for the copy job sample when they want to do async job processing.

PR #772, opened 2018-12-04
Adds file to run all tests in PR

bshaffer:
It's a bit annoying when a sample modifies `.kokoro`, root, or `testing` files and then we have to wait for 3 hours for all tests to pass. This makes it so PRs *never* run the entire test suite unless the `testing/RUN_ALL_TESTS` file is modified. Otherwise they only run tests for samples in the directories of the files modified.

It's a bit of a hack but it's the best I've come up with. Thoughts?

### DIFF
--- a/testing/RUN_ALL_TESTS
+++ b/testing/RUN_ALL_TESTS
@@ -1,0 +1,2 @@
+# Modify this file with whitespace in a PR in order to run all tests for the PR.
+# See "testing/run_test_suite.sh" for more details.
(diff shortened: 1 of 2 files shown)


SurferJeffAtGoogle:
What's the goal?  To run all the tests less often?  To give code submitters more control over which tests run?

bshaffer:
The goal is to prevent the full test suite from running when we only need to run for the sample that's been added

bshaffer:
Put another way, rarely do we want to run every sample in a pull request. This happens every night anyway, so do we want to do it in a pull request? Probably not. 

I could remove the file entirely, which would mean there's no way to trigger the entire suite from a PR. I thought it would be nice to keep SOME way to do it however...

PR #776, opened 2018-12-06
Update texttospeech.php

nnegrey:


sirtorry:
differing to brent on whether we want this.

bshaffer:
These region tags already exist in `texttospeech/src/*.php` for each sample.

PR #784, opened 2018-12-12
Add php55 and php73 docker images for tests

tmatsuo:
cc @dwsupplee

### DIFF
--- a/.kokoro/docker/php55/Dockerfile
+++ b/.kokoro/docker/php55/Dockerfile
@@ -1,0 +1,98 @@
+FROM gcr.io/gcp-runtimes/ubuntu_16_0_4
+ENV PHP_DIR=/opt/php55
+ENV PHP_SRC_DIR=/usr/local/src/php55-build
+ENV PATH=${PATH}:/usr/local/bin:${PHP_DIR}/bin
+
+RUN apt-get update && \
+    apt-get -y install \
+            autoconf \
+            build-essential \
+            git-core \
+            libbz2-dev \
+            libcurl4-openssl-dev \
+            libc-client2007e \
+            libc-client2007e-dev \
+            libfcgi-dev \
+            libfcgi0ldbl \
+            libfreetype6-dev \
+            libicu-dev \
+            libjpeg62-dbg \
+            libjpeg-dev \
+            libkrb5-dev \
+            libmcrypt-dev \
+            libpng12-dev \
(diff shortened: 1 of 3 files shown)


bshaffer:
This is great! Super great to be adding 7.3 so quickly! A few questions here:

1. Is there a reason we are adding PHP 5.5 when it's deprecated? I can see this being justified by the fact that `google/cloud` supports it, but this will add a lot of overhead (and will almost certainly require additional work to get the systems tests to pass). I would prefer to not add it if possible.

2. Is there a reason you aren't using the existing `Dockerfile.template` to keep these Dockerfiles consistent with each other?

tmatsuo:
1. google/cloud supports it, so we need to have test with php 5.5.

2. Dockerfile.template depends on gcr.io/google-appengine/phpx there are no images for PHP 5.5 and 7.3

tmatsuo:
You don't have to run the test against PHP 5.5. Only google-cloud-php will use that image.

tmatsuo:
If you really want to maintain the docker images for 1) php-docs-samples and 2) google-cloud-php, that's fine, but considering the requirements will be very similar, I suggest we use the same images.

bshaffer:
It doesn't really make sense to keep Dockerfiles here that are used in `google-cloud-php`. If `google-cloud-php` wants to use Dockerfiles in this repository, that's fine, but if they're only being used by `google-cloud-php`, then I think they belong in that repository.

Also, I believe we are moving the dockerfiles to a dedicated repository for container files anyway?

tmatsuo:
> Also, I believe we are moving the dockerfiles to a dedicated repository for container files anyway?
Yes. Eventually the `.kokoro/docker` directory will be moved to the new repo. When moving, it is easier to have all the Dockerfiles in one place, so that's my motivation to add them here.

If you don't like this, feel free to close this PR. I'll add it to somewhere else.

.kokoro/docker/php55/Dockerfile
```diff
@@ -0,0 +1,98 @@
+FROM gcr.io/gcp-runtimes/ubuntu_16_0_4
```
bshaffer: Just one small request - let's add a comment up here explaining this image is built for `GoogleCloudPlatform/google-cloud-php` testing.

tmatsuo: Done

PR #786, opened 2018-12-13
[entities] Migrate to new client LanguageServiceClient in 0.16.2

gogasca:
Migrate analyze entities and analyze entities to read from GCS
Modify test cases to read single URI

### DIFF
--- a/language/src/analyze_entities.php
+++ b/language/src/analyze_entities.php
@@ -25,7 +25,6 @@
 namespace Google\Cloud\Samples\Language;
 
 use Google\Cloud\Language\V1beta2\Document;
-use Google\Cloud\Language\V1beta2\Entity;
 use Google\Cloud\Language\V1beta2\LanguageServiceClient;
 /**
  * Find the entities in text.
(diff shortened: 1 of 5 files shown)


language/src/analyze_entities.php
```diff
@@ -39,26 +39,41 @@
...
-    $entities = $annotation->entities();
-    foreach ($entities as $entity) {
-        printf('Name: %s' . PHP_EOL, $entity['name']);
-        printf('Type: %s' . PHP_EOL, $entity['type']);
-        printf('Salience: %s' . PHP_EOL, $entity['salience']);
-        if (array_key_exists('wikipedia_url', $entity['metadata'])) {
-            printf('Wikipedia URL: %s' . PHP_EOL, $entity['metadata']['wikipedia_url']);
+    $languageServiceClient = new LanguageServiceClient(['projectId' => $projectId]);
+    try {
+        $entity_types = [
```
beccasaurus: 😭

You should be able to use `Enum_Class::name($entity->getType())` 😄

beccasaurus: e.g. as seen here: https://github.com/googleapis/gapic-generator/pull/2482/files#diff-e0f22c3a62995f9e35bf5c1caa4b5f8b

gogasca: Hi Rebbeca, I tried to see if I could implement the suggestion but the class Type does not have a name method not sure If I'm missing something obvious?, I look for alternatives using ReflectionClass but needs additional library: http://php.net/manual/en/reflectionclass.getconstants.php, which I prefer not to add additional dependency

beccasaurus: D'ohhhhh... I just saw that this isn't available until an upcoming version of protobuf is released for PHP (protobuf 3.7) – Hopefully _"~early January"_

/fyi @michaelbausor

beccasaurus: Bleech, these are so gross. Copy/pasting enums from the website isn't ideal. But these will be able to be updated soon!!!

👍 Carry On.

gogasca: Sounds good, I'' keep an eye once is updated I will create a new PR. Thanks

PR #790, opened 2018-12-14
[classify] Migrate to new client LanguageServiceClient in 0.16.2

gogasca:


language/language.php
```diff
@@ -290,7 +290,7 @@
         // Regex to match a Cloud Storage path as the first argument
         // e.g "gs://my-bucket/file_with_text.txt"
         if (preg_match('/^gs:\/\/([a-z0-9\._\-]+)\/(\S+)$/', $content, $matches)) {
-            classify_text_from_file($content, $projectId);
+            classify_text_from_file($matches[0], $projectId);
```
bshaffer: In this case, `$content` will always be equal to `$matches[0]`. Why update this?

language/src/classify_text.php
```diff
@@ -47,21 +47,23 @@ function classify_text($text, $projectId = null)
...
+        // Pass GCS URI and set document type to PLAIN_TEXT
+        $document->setContent($text)->setType(1);
+        // Call the analyzeSentiment function
+        $response = $languageServiceClient->classifyText($document);
+        $categories = $response->getCategories();
+        // Print document information
+        foreach ($categories as $category) {
+            printf('Category Name: %s' . PHP_EOL, $category->getName());
+            printf('Confidence: %s' . PHP_EOL, $category->getConfidence());
+            printf(PHP_EOL);
```
bshaffer: nit: just use `print` here

language/src/classify_text_from_file.php
```diff
@@ -36,22 +37,25 @@
...
-    foreach ($categories as $category) {
-        printf('Category Name: %s' . PHP_EOL, $category['name']);
-        printf('Confidence: %s' . PHP_EOL, $category['confidence']);
-        printf(PHP_EOL);
+    $languageServiceClient = new LanguageServiceClient(['projectId' => $projectId]);
+    try {
+        // Create a new Document
+        $document = new Document();
+        // Pass GCS URI and set document type to PLAIN_TEXT
+        $document->setGcsContentUri($gcsUri)->setType(1);
```
bshaffer: let's use an actual call to the Enum class here

language/src/classify_text.php
```diff
@@ -47,21 +47,23 @@ function classify_text($text, $projectId = null)
...
-    foreach ($categories as $category) {
-        printf('Category Name: %s' . PHP_EOL, $category['name']);
-        printf('Confidence: %s' . PHP_EOL, $category['confidence']);
-        printf(PHP_EOL);
+    $languageServiceClient = new LanguageServiceClient(['projectId' => $projectId]);
+    try {
+        // Create a new Document
+        $document = new Document();
+        // Pass GCS URI and set document type to PLAIN_TEXT
+        $document->setContent($text)->setType(1);
```
bshaffer: let's use an actual call to the Enum class here

gogasca:
Changes submitted, thanks for the review

PR #793, opened 2018-12-15
[analyze_all] Migrate to new client LanguageServiceClient in 0.16.2

gogasca:


gogasca:
[all] Migrate to new client LanguageServiceClient in 0.16.2

gogasca:
@bshaffer @beccasaurus is there a workaround for converting class Type constants to string?

bshaffer:
@gogasca There will be in the next version of protobuf for PHP, but unfortunately that won't be released until next month.

PR #798, opened 2018-12-21
add comment for clarification of authentication sample in docs

hongalex:
add the comment about default credentials/scope for adding to this sample:
https://cloud.google.com/iot/docs/how-tos/credentials/authenticating-applications#iot-core-service-accounts-php

Currently it's pointing to list_registries.php, but to keep it consistent with other samples, we will point it to create_registry.php instead and add this comment.

### DIFF
--- a/iot/src/create_registry.php
+++ b/iot/src/create_registry.php
@@ -38,7 +38,12 @@ function create_registry(
 ) {
     print('Creating Registry' . PHP_EOL);
 
-    // Instantiate a client.
+    // The Google Cloud Client Library automatically checks the environment
+    // variable GOOGLE_APPLICATION_CREDENTIALS for the Service Account
+    // credentials, and defaults scopes to [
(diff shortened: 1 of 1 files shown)


bshaffer:
But it looks to me like the other samples are using a `getClient` method, and not a `create_registry` method, for those samples.

hongalex:
@bshaffer Go, Java, and Ruby don't have explicit `getClient` methods and so the examples given are within the context of the `create_registry` methods. For the other languages (such as Node, Python, and C#) the samples were written with an explicit `getClient` method defined. PHP falls under the former and we wanted to point the docs from `list_registries` to `create_registry`

After thinking about this some more, I'm not entirely sure what the original reasoning was behind having explicit/implicit client retrieval for different languages and will follow up. In the meantime, we're trying to at least minimize the inconsistency by switching the PHP docs to point at `create_registry.php` instead.

bshaffer:
Fine with me! Thank you.

PR #799, opened 2018-12-27
Local fallback for compiled views

tyloo:
If global server variable 'GAE_SERVICE' is not set, then we are in local environment and fallback to default Laravel behavior.


bshaffer:
Thank you for your submission!
As of #823 we set the `VIEW_COMPILED_PATH` environment variable in `app.yaml`, which means the behavior works as intended in the local environment. So this change is not necessary.

PR #800, opened 2018-12-28
Updates KMS samples to use google/cloud-kms client library

bshaffer:


kms/quickstart.php
```diff
@@ -19,40 +19,27 @@
...
 // Your Google Cloud Platform project ID
 $projectId = 'YOUR_PROJECT_ID';
 
-// Lists keys in the "global" location.
-$location = 'global';
-
-// The resource name of the location associated with the KeyRings
-$parent = sprintf('projects/%s/locations/%s', $projectId, $location);
+// Lists keys in the "global" location. Could also be "us-west1", etc.
+$locationId = 'global';
```
grayside: Why you would ever choose a specific region when global is available?

bshaffer: Maybe you want to lock down permissions, or maybe it's for organization? I don't know TBH.

bdhess: You might choose a region like 'us-east1' if you have specific latency and/or data residency requirements:
https://cloud.google.com/kms/docs/locations#location_types

kms/src/remove_member_from_keyring_policy.php
```diff
@@ -0,0 +1,70 @@
...
+
+// The resource name of the CryptoKey.
+$keyRingName = $kms->keyRingName($projectId, $locationId, $keyRingId);
+
+// Get the current IAM policy and remove the account to it.
+$policy = $kms->getIamPolicy($keyRingName);
+foreach ($policy->getBindings() as $binding) {
+    if ($binding->getRole() == $role) {
+        $members = $binding->getMembers();
+        foreach ($members as $i => $existingMember) {
```
grayside: Could shrink this a couple lines with `array_search()`. It's looking like one of the more complicated operations. Maybe the client library should have some helper functions on a couple of these operations that require fairly boilerplate modifications?

In practice, I would wrap this in a helper library rather than inline the code anywhere.

kms/src/add_member_to_cryptokey_policy.php
```diff
@@ -0,0 +1,64 @@
...
+$policy = $kms->getIamPolicy($cryptoKeyName);
+$bindings = $policy->getBindings();
+$bindings[] = new Binding([
+    'members' => [$member],
+    'role' => $role,
+]);
+$policy->setBindings($bindings);
+
+// Set the new IAM Policy.
+$kms->setIamPolicy($cryptoKeyName, $policy);
```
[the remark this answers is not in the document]
bshaffer: I believe the API will throw an error if this is not successful

PR #801, opened 2018-12-28
speech gapic migration

sirtorry:


speech/src/transcribe_sync.php
```diff
@@ -34,30 +37,46 @@
...
+    $handle = fopen($audioFile, 'r');
+    $content = fread($handle, filesize($path));
+    fclose($handle);
+
+    // set string as audio content
+    $audio = (new RecognitionAudio())
+        ->setContent($content);
+
+    // set config
+    $config = (new RecognitionConfig())
```
bshaffer: It's now possible to do this:

```php
$config = new RecognitionConfig([
    'encoding' => $encoding,
    'sampleRateHertz' => $sampleRateHertz,
    'languageCode' => $languageCode,
]);
```

I don't really have strong opinions on using this here or not, the way you have it looks great. Just wanted to give you a heads up :D

speech/src/streaming_recognize.php
```diff
@@ -46,8 +46,13 @@
...
  */
-function streaming_recognize($audioFile, $languageCode, $encoding, $sampleRateHertz)
+function streaming_recognize($audioFile)
 {
+    // change these variables
+    $encoding = AudioEncoding::LINEAR16;
+    $sampleRateHertz = 32000;
+    $languageCode = 'en-US';
+
     if (!defined('Grpc\STATUS_OK')) {
```
bshaffer: I don't know why we did it this way... would be much better as

```php
if (!extension_loaded('grpc')) {
   //...
}
```

speech/src/transcribe_async.php
```diff
@@ -36,43 +37,55 @@
...
 
-    // Create the asyncronous recognize operation
-    $operation = $speech->beginRecognizeOperation(
-        fopen($audioFile, 'r'),
-        $options
-    );
+    // get contents of a file into a string
+    $handle = fopen($audioFile, 'r');
+    $content = fread($handle, filesize($audioFile));
+    fclose($handle);
```
bshaffer: I am not sure if this is intentional, but you could simply simplify this by just calling:

```php
$content = file_get_contents($audioFile);
```

speech/src/transcribe_auto_punctuation.php
```diff
@@ -32,11 +32,11 @@
...
  * Transcribe the given audio file with auto punctuation enabled
  */
-function transcribe_auto_punctuation($path)
+function transcribe_auto_punctuation($audioFile)
 {
     // get contents of a file into a string
-    $handle = fopen($path, 'r');
-    $content = fread($handle, filesize($path));
+    $handle = fopen($audioFile, 'r');
+    $content = fread($handle, filesize($audioFile));
```
bshaffer: same here, there's no advantage to using this method over calling `file_get_contents`, since we're loading the entire contents of the file into memory either way.

speech/src/transcribe_async_words.php
```diff
@@ -42,46 +43,61 @@
...
-    $speech = new SpeechClient([
-        'languageCode' => $languageCode,
-    ]);
+    // change these variables
+    $encoding = AudioEncoding::LINEAR16;
+    $sampleRateHertz = 32000;
+    $languageCode = 'en-US';
 
-    // When true, time offsets for every word will be included in the response.
-    $options['enableWordTimeOffsets'] = true;
```
bshaffer: We need this to be included, otherwise the individual words do not come back, and the tests fail.

sirtorry: this would explain a lot.

PR #803, opened 2019-01-02
Removes script: auto from php72 samples where it's no longer needed

bshaffer:


appengine/php72/grpc/app.yaml
```diff
@@ -1,8 +1,5 @@
 runtime: php72
 
 handlers:
-- url: /
-  script: auto
-
-- url: /(monitoring|spanner|speech)\.php
+- url: /(monitoring|spanner|speech)\.php$
```
grayside: Why is this needed? Seems like it should be more deleted.

PR #804, opened 2019-01-03
Update detect_intent_stream.php

edocabhi:
The query result block was missing reference to the response. Hence it was throwing error "Fatal error: Uncaught Error: Call to a member function getQueryText() on null in ...". Moving it inside the loop fixes the issue.


edocabhi:
I signed it!

PR #820, opened 2019-01-17
add additional details re cloud sql proxy auth

flashadvocate:
Ran into some authentication issues when using Cloud SQL Proxy. Discovered some additional requirements through the following resources:

- https://github.com/GoogleCloudPlatform/cloudsql-proxy
- https://cloud.google.com/sql/docs/mysql/sql-proxy

The documentation mentions several ways to authenticate, but keeping consistency with the other steps provided in this guide, application default auth seems to be the most appropriate

### DIFF
--- a/appengine/php72/laravel-framework/README.md
+++ b/appengine/php72/laravel-framework/README.md
@@ -8,8 +8,9 @@ overview of PHP and learn ways to run PHP apps on GCP.
 ## Prerequisites
 
 1. Create a project in the [Google Cloud Platform Console](https://console.cloud.google.com/project).
-1. Enable billing for your project.
-1. Install the [Google Cloud SDK][cloud_sdk].
+2. Enable billing for your project.
+3. Install the [Google Cloud SDK][cloud_sdk].
+4. Authenticate with the Cloud SDK - `gcloud auth application-default login`
 
 ## Prepare
 
@@ -22,7 +23,7 @@ laravel.com. This version was tested to work with `laravel/laravel-framework:^5.
 
         php artisan serve
 
-1. Visit [http://localhost:8000](http://localhost:8000) to see the Laravel
+2. Visit [http://localhost:8000](http://localhost:8000) to see the Laravel
    Welcome page.
 
 ## Deploy
@@ -37,7 +38,7 @@ laravel.com. This version was tested to work with `laravel/laravel-framework:^5.
           APP_KEY: YOUR_APP_KEY
           APP_STORAGE: /tmp
 
-1. Copy the [`bootstrap/app.php`](bootstrap/app.php) and
+2. Copy the [`bootstrap/app.php`](bootstrap/app.php) and
   [`config/view.php`](config/view.php) files included in this sample into the
   corresponding directories of your Laravel application. These two files ensure
   your Laravel application writes to `/tmp` for caching in production.
@@ -45,7 +46,7 @@ laravel.com. This version was tested to work with `laravel/laravel-framework:^5.
   > If you are using an existing Laravel application, just copy the
   `google-app-engine-deployment` blocks from these files.
 
-1. Replace `YOUR_APP_KEY` in `app.yaml` with an application key you generate
+3. Replace `YOUR_APP_KEY` in `app.yaml` with an application key you generate
   with the following command:
 
         php artisan key:generate --show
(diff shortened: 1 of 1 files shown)


appengine/php72/laravel-framework/README.md
```diff
@@ -91,6 +91,10 @@ Laravel, you need to manually add the `DB_SOCKET` value to
     * Use `gcloud` to create a database for the application.
 
             gcloud sql databases create laravel --instance=YOUR_INSTANCE_NAME
+            
+    * Authenticate with the Google Cloud SDK. Note that the default service account needs to have the `Cloud SQL Client` role. The Cloud SQL Admin API must also be enabled under `APIs and Services` in order to use the Cloud SQL Proxy Client.
+                
+            gcloud auth application-default login
```
bshaffer: This is great! It doesn't really belong here. In the `Requirements` section we cover installing  and authenticating with the Cloud SDK. We could move this bit above, and add another bullet point for enabling the API.

flashadvocate: Do you mean as part of "prerequisites", or somewhere else @bshaffer

bshaffer: As part of prerequisites, yes

flashadvocate: Made some more updates. Also tweaked the bulleting since it was getting broken under "Deploy". Let me know if this is more what you were thinking.

appengine/php72/laravel-framework/README.md
```diff
@@ -8,8 +8,9 @@ overview of PHP and learn ways to run PHP apps on GCP.
 ## Prerequisites
 
 1. Create a project in the [Google Cloud Platform Console](https://console.cloud.google.com/project).
-1. Enable billing for your project.
-1. Install the [Google Cloud SDK][cloud_sdk].
+2. Enable billing for your project.
```
bshaffer: FYI changing the number here is not necessary, as markdown will increment all the numbers as expected regardless of the number chosen. So using `1.` for everything is preferable so that when new steps are added, you don't have to redo all the list numbers.

appengine/php72/laravel-framework/README.md
```diff
@@ -91,8 +92,10 @@ Laravel, you need to manually add the `DB_SOCKET` value to
     * Use `gcloud` to create a database for the application.
 
             gcloud sql databases create laravel --instance=YOUR_INSTANCE_NAME
+            
+    * The default service account needs to have the `Cloud SQL Client` role. The Cloud SQL Admin API must also be enabled under `APIs and Services` in order to use the Cloud SQL Proxy Client.
```
bshaffer: How about a link to enable the API? The following link should do it: `https://console.cloud.google.com/flows/enableapi?apiid=sqladmin`

appengine/php72/laravel-framework/README.md
```diff
@@ -10,6 +10,7 @@ overview of PHP and learn ways to run PHP apps on GCP.
 1. Create a project in the [Google Cloud Platform Console](https://console.cloud.google.com/project).
 1. Enable billing for your project.
 1. Install the [Google Cloud SDK][cloud_sdk].
+1. Authenticate with the Cloud SDK - `gcloud auth application-default login`
```
bshaffer: Under [Initialize the SDK](https://cloud.google.com/sdk/docs/quickstart-macos#initialize_the_sdk), the user is directed to login with `gcloud`. So having this step here as well is redundant and also incorrect, since authenticating as a user instead of a service account is also acceptable. So I think we need to remove this line. We could change the above line to "Install and initialize Google Cloud SDK" if you think that is clearer.

flashadvocate: I mentioned this in my original post; the context being provided by the rest of the steps lends much more to simply using the application-default login. For someone walking through the tutorial, and having to take the additional steps to leverage a service account, it's going to be a much more complicated experience. I think it's certainly worth mentioning the several ways you can authenticate, however, the cloud sql proxy page mentions these. Authing as application-default isn't as clear, and thought it made sense to point it out more directly here.

flashadvocate: I disagree, personally. This step in particular was what made it possible for me to actually use the proxy client. The [documentation ](https://cloud.google.com/sql/docs/mysql/sql-proxy)for the SQL proxy connection doesn't mention this anywhere, and so I thought it made sense to more clearly point this out here, instead of being vague, and sending users off on another adventure through verbose documentation. This would have made my first go-around much easier.

It may be worth noting that this process is **purely** for getting the cloud sql client proxy to work correctly.

flashadvocate: I've made additional changes. I think it's worth pointing users to the repo, which includes information not provided in the documentation regarding their authentication options. Hopefully this is a worthwhile compromise.

bshaffer:
Closing this because we removed the README tutorials in favor of the community tutorial (see https://cloud.google.com/community/tutorials/run-laravel-on-appengine-standard). Feel free to submit your PR over there (the code is [here](https://github.com/GoogleCloudPlatform/community/blob/master/tutorials/run-laravel-on-appengine-standard/index.md)). I am sorry for the goose chase. I can port your code if you'd like, but then you won't get credit :-/

bshaffer:
Added these changes in https://github.com/GoogleCloudPlatform/community/pull/749

Thanks!

PR #833, opened 2019-02-14
Laravel 5.7.* Google App Engine Standard Support

Snaver:
Setup Laravel correctly to run in a production environment on Google App Engine Standard.

**Issues:**
When deploying using `gcloud app deploy` PHP composer dependencies should be ignored via `.gcloudignore` and instead installed on each deploy/build automatically via Google Cloud. 

Part of this composer process is the running of the callback `post-autoload-dump`, which triggers a nifty Laravel feature to autoload dependencies into the App namespace - however this causes issues as the framework expects bootstrap/cache to be writable. In App Engine Standard the only writable directory the app has access to is the /tmp directory.

Laravel actually uses this directory for a few other things (config and route cache), so we need to configure the framework to use a different directory.

**Changes:**
- Change cached framework files directory from bootstrap/cache to /tmp. 
- Exclude local bootstrap/cache files from deployment
- Tweak useStoragePath() call to only be used when ENV var is present


Snaver:
Hmm during my testing I hard coded the cache path to `/tmp` - which worked, however this unfortunately cached the results in Cloud Build upon subsequent builds. I changed the code in Laravel to use an Environment var that doesn't appear to exist in the Build process, however my build continue to pass as it was using the cache.. I've just pushed again and the build broke.

I've hacked it temporarily to work in the short term (at the expense of breaking the magic of my local environment, but would be good to figure out a way to set Environment vars dynamically for the Build process???

```
"scripts": {
    "post-autoload-dump": [
        "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
        "php -r 'var_dump(@$_ENV);'",
        "php -r 'var_dump(@$_SERVER);'",
        "APP_BOOTSTRAP_PATH=/tmp php artisan package:discover --ansi"
    ],
    "post-create-project-cmd": [
        "php artisan key:generate --ansi"
    ]
}
```

appengine/php72/laravel-framework/app/Application.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+namespace App;
+
+class Application extends \Illuminate\Foundation\Application
```
bshaffer: When I follow [the tutorial](https://cloud.google.com/community/tutorials/run-laravel-on-appengine-standard) and run `laravel new blog`,  this file is not generated. So I am not sure how this is supposed to fit in the tutorial

appengine/php72/laravel-framework/app.yaml
```diff
@@ -5,8 +5,14 @@ env_variables:
...
 
   ## To use Stackdriver logging in your Laravel application, copy
   ## "app/Logging/CreateStackdriverLogger.php" and "config/logging.php"
   ## into your Laravel application. Then uncomment the following line:
   # LOG_CHANNEL: stackdriver
+
+  APP_SERVICES_CACHE: /tmp/services.php
+  APP_PACKAGES_CACHE: /tmp/packages.php
+  APP_CONFIG_CACHE: /tmp/config.php
+  APP_ROUTES_CACHE: /tmp/routes.php
```
bshaffer: These environment variables are never used

bshaffer:
I am closing this because it contains fixes that are not accurate. Try following the tutorial on the community site and open an issue if you encounter errors: https://cloud.google.com/community/tutorials/run-laravel-on-appengine-standard

sertaconay:
This actually worked. I was getting "/srv/bootstrap/cache directory not writable" error even I copied and pasted everything from the documentation. Thanks!
I use GAE standard and Laravel 5.8.24

VoordeMensen:
+1 - after reading https://medium.com/meetech/deploying-a-laravel-app-to-google-app-engine-with-gitlab-ci-2f9dbde5bef4 and adding *all* the temp env vars to app.yaml I got Bitbucket CI working on GAE.

uziel-bueno:
Setting the following environment variables pointing the writable `/tmp` path allow _Laravel_ to override the default `bootstrap/cache` path:

```
env_variables:
  APP_STORAGE: /tmp
  VIEW_COMPILED_PATH: /tmp
  APP_SERVICES_CACHE: /tmp/services.php
  APP_PACKAGES_CACHE: /tmp/packages.php
  APP_CONFIG_CACHE: /tmp/config.php
  APP_ROUTES_CACHE: /tmp/routes.php
  APP_EVENTS_CACHE: /tmp/events.php
```

As mentioned in the tutorial, an additional line must be added to the `bootstrap/app.php` file in order to override
the default storage directory path which can be done better in this way instead:

```
// boostrap/app.php

$app->useStoragePath(env('APP_STORAGE', $app->storagePath()));
```

I think this is better because there is no hardcoded `/storage` string. This will be compatible with new versions of
the framework in case the authors decide to internally change the default storage path from `/storage` to something
else.

PR #858, opened 2019-03-18
storage: add bucket policy only samples

frankyn:
Add Bucket Policy Only samples.
https://cloud.google.com/storage/docs/using-bucket-policy-only


storage/storage.php
```diff
@@ -402,6 +402,30 @@
...
+    ->addOption('get', null, InputOption::VALUE_NONE, 'Get Bucket Policy Only on a Cloud Storage bucekt')
+    ->setCode(function ($input, $output) {
+        $bucketName = $input->getArgument('bucket');
+        if ($input->getOption('enable')) {
+            enable_bucket_policy_only($bucketName);
+        } elseif ($input->getOption('disable')) {
+            disable_bucket_policy_only($bucketName);
+        } elseif ($input->getOption('get')) {
+            get_bucket_policy_only($bucketName);
+        }
```
bshaffer: throw an exception if no option is supplied

storage/test/BucketPolicyOnlyCommandTest.php
```diff
@@ -0,0 +1,112 @@
...
+              '--disable' => true,
+          ],
+          ['interactive' => false]
+        );
+
+        $outputString = <<<EOF
+Bucket Policy Only was disabled for {$this->bucket->name()}
+
+EOF;
+        $this->expectOutputString($outputString);
```
bshaffer: Would like to see some verification that this was actually enabled/disabled

frankyn:
Thanks @bshaffer, PTAL.

PR #861, opened 2019-03-25
Add Cloud Spanner Batch DML sample

jsimonweb:


spanner/spanner.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Copyright 2016 Google Inc.
+ * Copyright 2019 Google LLC
```
bshaffer: Year shouldn't be updated, FYI

spanner/quickstart.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Copyright 2016 Google Inc.
+ * Copyright 2019 Google LLC
```
bshaffer: No reason to update the year here

PR #864, opened 2019-03-28
Add Object Tracking support for Video

gogasca:
Add Video Tracking support via local file and GCS URI

### DIFF
--- a/video/src/analyze_object_tracking.php
+++ b/video/src/analyze_object_tracking.php
@@ -1,0 +1,83 @@
+<?php
+
+/**
+ * Copyright 2019 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


sirtorry:
looks fine. I'll let @bshaffer comment on it.

video/src/analyze_object_tracking.php
```diff
@@ -0,0 +1,83 @@
...
+
+    # Wait for the request to complete.
+    $operation->pollUntilComplete($options);
+
+    # Print the results.
+    if ($operation->operationSucceeded()) {
+        $results = $operation->getResult()->getAnnotationResults()[0];
+        # Process video/segment level label annotations
+        $objectAnnotation = $results->getObjectAnnotations()[0];
+
```
sirtorry: I'd probably make an $objectEntity

PR #866, opened 2019-04-09
[Cloud Tasks] Add HTTP task sample

averikitsch:
Test projects will have to be granted permission for HTTP targets or wait until whitelisting is off.


tasks/phpunit.xml.dist
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright 2018 Google LLC.
```
bshaffer: nit: 2019! And should be for all new files!

tasks/README.md
```diff
@@ -0,0 +1,69 @@
...
+5.  Create a Queue
+    To create a queue using the Cloud SDK, use the following gcloud command:
+    ```sh
+    gcloud beta tasks queues create-app-engine-queue my-appengine-queue
+    ```
+6. Set environment variables:
+
+    First, your project ID:
+
+        export PROJECT_ID=my-project-id
```
bshaffer: Setting environment  variables for PROJECT_ID, LOCATION_ID, and QUEUE_ID are not required, as we pass them into the script directly. The only environment variable required *should be* `GOOGLE_APPLICATION_CREDENTIALS`

averikitsch:
Thanks for the review. Can I go ahead and merge?

bshaffer:
Looks great!! Thank you!

PR #873, opened 2019-04-22
Updates dlp, error, video, speech, language, and tts samples to new format

bshaffer:
Updates some samples to use most current version and uses enum `::name` methods instead of hardcoded enum arrays.


bshaffer:
Tests are failing because of https://github.com/googleapis/gax-php/pull/237

PR #1, opened 2015-09-25
Move the existing sample into a compute/ subdirectory and update README

SurferJeffAtGoogle:
and CONTRIBUTING docs.

PR #880, opened 2019-05-01
Update DLP samples to new format

bshaffer:
Updates all DLP samples to new samples format and fixes a few bugs

Supersedes #739


bshaffer:
Considering this format as well, which I like better. The difference here is the one-sentence summary of what the sample does is moved above the `use` statements:

```php
# [START dlp_categorical_stats]
/**
 * Computes risk metrics of a column of data in a Google BigQuery table.
 */
use Google\Cloud\Dlp\V2\DlpServiceClient;
use Google\Cloud\Dlp\V2\RiskAnalysisJobConfig;
use Google\Cloud\Dlp\V2\BigQueryTable;
use Google\Cloud\Dlp\V2\DlpJob\JobState;
use Google\Cloud\Dlp\V2\Action;
use Google\Cloud\Dlp\V2\Action\PublishToPubSub;
use Google\Cloud\Dlp\V2\PrivacyMetric\CategoricalStatsConfig;
use Google\Cloud\Dlp\V2\PrivacyMetric;
use Google\Cloud\Dlp\V2\FieldId;
use Google\Cloud\PubSub\PubSubClient;

// $callingProjectId = 'The project ID to run the API call under';
// $dataProjectId = 'The project ID containing the target Datastore';
// $topicId = 'The name of the Pub/Sub topic to notify once the job completes';
// $subscriptionId = 'The name of the Pub/Sub subscription to use when listening for job';
// $datasetId = 'The ID of the dataset to inspect';
// $tableId = 'The ID of the table to inspect';
// $columnName = 'The name of the column to compute risk metrics for, e.g. "age"';
```

dlp/src/categorical_stats.php
```diff
@@ -29,129 +45,120 @@
...
+    ->setTopic($topic->name());
+
+$action = (new Action())
+    ->setPubSub($pubSubAction);
+
+// Construct risk analysis job config to run
+$riskJob = (new RiskAnalysisJobConfig())
+    ->setPrivacyMetric($privacyMetric)
+    ->setSourceTable($bigqueryTable)
+    ->setActions([$action]);
```
grayside: While it's not in scope of this PR, wanted to flag that this list of instantiations is difficult to follow. A couple things that would help me:

* Group by product a bit more, e.g., move Pub/Sub instantiation down before Pub/Sub is used.
* More contextual comments, the current ones are very concise but don't explain "why" very much.

dlp/src/categorical_stats.php
```diff
@@ -29,129 +45,120 @@
...
+// Listen for job notifications via an existing topic/subscription.
+$subscription = $topic->subscription($subscriptionId);
+
+// Poll via Pub/Sub until job finishes
+while (true) {
+    foreach ($subscription->pull() as $message) {
+        if (isset($message->attributes()['DlpJobName']) &&
+            $message->attributes()['DlpJobName'] === $job->getName()) {
+            $subscription->acknowledge($message);
+            break 2;
```
grayside: 1. Pounding the Pub/Sub API like this seems overly aggressive even for a sample, maybe it should have pauses/exponential backoff?
2. Seems like this should note the assumption the Pub/Sub topic might be used for other work than DLP jobs?
3. Why acknowledge the message before doing any work? This means we don't get Pub/Sub retries if later code chokes.

dlp/src/categorical_stats.php
```diff
@@ -29,129 +45,120 @@
...
-    // Poll via Pub/Sub until job finishes
-    while (true) {
-        foreach ($subscription->pull() as $message) {
-            if (isset($message->attributes()['DlpJobName']) &&
-                $message->attributes()['DlpJobName'] === $job->getName()) {
-                $subscription->acknowledge($message);
-                break 2;
-            }
+/** Uncomment and populate these variables in your code */
+// $callingProjectId = 'The project ID to run the API call under';
```
alixhami: In our sample rubric meetings, we talked about removing parameters whenever possible. In a sample, it's much easier to understand if you give actual example values whenever possible. The purpose of the commented out variables isn't intuitive for everyone, and if you do have commented out variables, it's easier to understand if they have an actual example value rather than a string description of what the value should be (`'your-project-id'` rather than `'Your project ID.'`).

For example (example is BQ python of course 😃 ) this:
`table = client.create_table('your_new_table')`
is easier to understand than:
```
# Uncomment and populate these variables in your code
# table_id = 'The name of the table to create'
# other_stuff = 'here'
...

table = client.create_table(table_id)
```

In the first example, the user can see what kind of value a `table_id` is (a string), and what it would look like. The sample value is also in-context rather than separated at the top of the sample. There's also no opportunity for miscommunication about why variables would need to be uncommented.

Even project ID can be set to an example string if you immediately reset it:
```
project_id = 'your-project-id'  # replace with your project ID
# [END region_tag]
project_id = 'actual-project-id-for-testing'
# [START region_tag]
```

tldr: samples aren't supposed to be reusable functions, so I'd remove as many of these parameters as possible and use example values instead.

dlp/src/categorical_stats.php
bshaffer: Added Exponential backoff to fix #1.

dlp/README.md
```diff
@@ -27,32 +27,20 @@ DLP API from PHP.
...
+    Usage: php src/inspect_string.php PROJECT_ID STRING
+
+    $ php src/inspect_string.php your-project-id 'bob@example.com'
     Findings:
-      Quote: Robert
-      Info type: US_MALE_NAME
-      Likelihood: Very likely
+      Quote: bob@example.com
+      Info type: EMAIL_ADDRESS
+      Likelihood: 4
```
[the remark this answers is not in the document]
bshaffer: Added a link to the official docs in the README

dlp/src/categorical_stats.php
bshaffer: I think it's a great idea to use `project_id = 'your-project-id'` instead of `project_id = 'A project ID for testing`. However, I do think having all placeholder variables in one place at the top of the sample is okay, because otherwise they can be lost in the context of the sample, e.g. a user may not notice the placeholder `your_new_table` if it's part of the function call `table = client.create_table('your_new_table')`. In that case, I do think the following is better:

```
tableName = 'your_new_table'; 
table = client.create_table(tableName);
```

PR #882, opened 2019-05-03
Add GAE Flex websockets sample

ace-n:
@grayside @andrewsg FYI


ace-n:
**Note** (mostly to myself): some of the packages used here aren't compatible with older versions of PHP (and are causing CI failures).

appengine/flexible/websockets/composer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "require": {
+        "cboden/ratchet": "^0.4.1",
+        "guzzlehttp/psr7": "^1.5",
+        "silex/silex": "^2.3",
```
grayside: Silex is end-of-life and should not be used.
Newer samples might use raw PHP or Slim. @bshaffer has also been looking at Lumen for bookshelf.

appengine/flexible/websockets/index.html.twig
```diff
@@ -0,0 +1,88 @@
+{#
```
grayside: This file doesn't seem to be using Twig? Maybe just use pure HTML and drop the Twig dependency.

appengine/flexible/websockets/index.html.twig
```diff
@@ -0,0 +1,88 @@
...
+# Copyright 2019 Google LLC
```
grayside: I overhauled this HTML for the Go sample (fusing the ideas in Python and Nodejs and adding some additional styling), feel free to copy [golang-samples index.html](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/appengine_flexible/websockets/static/index.html), screenshot in [golang-samples PR #837](https://github.com/GoogleCloudPlatform/golang-samples/pull/837)

appengine/flexible/websockets/composer.json
ace-n: Chatted offline - @bshaffer is fine with this, so I'm marking as resolved for now.

appengine/flexible/websockets/index.html.twig
ace-n: Done.

appengine/flexible/websockets/index.html.twig
ace-n: Done.

appengine/flexible/websockets/socket-server.php
```diff
@@ -0,0 +1,46 @@
...
+
+$port = 8000;
+
+$server = IoServer::factory(
+    new HttpServer(
+        new WsServer(
+                    $socketDemo
+                )
+            ),
+    $port
```
grayside: trivial bit of whitespace here.

ace-n: Fixed - surprised that `php-cs-fixer` didn't handle this.

ace-n:
> **Note** (mostly to myself): some of the packages used here aren't compatible with older versions of PHP (and are causing CI failures).

This is now a **do not merge** until @bshaffer signs off.

PR #887, opened 2019-05-09
Storage: v2 signed url sample

jkwlui:
first line of PHP since many, many years.. Feedback appreciated :)

### DIFF
--- a/storage/src/get_object_v2_signed_url.php
+++ b/storage/src/get_object_v2_signed_url.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


storage/test/ObjectSignedUrlTest.php
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * Copyright 2017 Google Inc.
```
bshaffer: Let's make this 2019

storage/src/get_object_v2_signed_url.php
```diff
@@ -0,0 +1,52 @@
...
+function get_object_v2_signed_url($bucketName, $objectName)
+{
+    $storage = new StorageClient();
+    $bucket = $storage->bucket($bucketName);
+    $object = $bucket->object($objectName);
+    # This URL is valid for 1 hour
+    $url = $object->signedUrl(
+        new \DateTime('next hour'),
+        [
+            'version' => 'v2'
```
frankyn: Not necessary and existing V2 samples don't use `version` => 'v2' in other languages.

PR #892, opened 2019-05-16
Updates DLP comments and variables to be better canonical samples

bshaffer:
Addresses @alixhami's comments about canonical DLP samples

### DIFF
--- a/dlp/src/categorical_stats.php
+++ b/dlp/src/categorical_stats.php
@@ -26,9 +26,9 @@
 require_once __DIR__ . '/../vendor/autoload.php';
 
 if (count($argv) != 8) {
-    return print("Usage: php categorical_stats.php CALLING_PROJECT DATA_PROJECT TOPIC SUBSCRIPTION DATASET TABLE COLUMN\n");
+    return print("Usage: php categorical_stats.php PROJECT_ID BIGQUERY_PROJECT TOPIC SUBSCRIPTION DATASET TABLE COLUMN\n");
 }
-list($_, $callingProjectId, $dataProjectId, $topicId, $subscriptionId, $datasetId, $tableId, $columnName) = $argv;
+list($_, $projectId, $bigqueryProjectId, $topicId, $subscriptionId, $datasetId, $tableId, $columnName) = $argv;
 
 # [START dlp_categorical_stats]
 /**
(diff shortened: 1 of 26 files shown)


ace-n:
Leaving this up to @alixhami since he's more familiar with sample standards.

alixhami:
To be honest, I don't think this is a quality improvement overall. These are my concerns:
1. **Prefer to keep documentation text out of samples** - There is a lot of explanation in comments about the parameters that is probably already explained in the docs page that will show this sample. It's not a best practice to have redundant information in sample comments because it takes up extra space, is prone to typos, and is less likely to be updated than the documentation copy. Example of information that should be in the documentation copy instead:
    ```
    // (Optional) The name of the surrogate custom info type to use
    // Only necessary if you want to reverse the deidentification process
    // Can be essentially any arbitrary string, as long as it doesn't appear
    // in your dataset otherwise.
    ```
2. **Prefer samples to be as short and clear as possible** - The samples are noticeably longer now. It's harder to grasp what's going on in a sample if you have to scroll to see the whole thing. This would be fine if the operation was complex enough to need to be long, but I think these are longer than necessary.
3. **Prefer in-context example values to parameters** - In our rubric meetings, we decided that it was a best practice to remove as many parameters as possible, which was not done here. It's easier to tell what's going on if you have literal example values in-context rather than parameters. The commented out parameters do show some literal example values, but these are separated from where they are used and the commented out values are not subject to testing, so they are prone to typos. As the user is reading the sample, they have to scroll back and forth between where the parameters are defined and where they are used. This hinders comprehension. (More explanation in my comment on the [previous PR](https://github.com/GoogleCloudPlatform/php-docs-samples/pull/880#discussion_r281286319)). An example of something that does not need to be parameterized is the string to deidentify:
    Current (out of context parameter):
    ```php
    // The string to deidentify
    // $string = 'My SSN is 372819127';
    //...
    // <lots of code here, requiring user scrolling>
    //...
    $item = (new ContentItem())
        ->setValue($string);
   ```
    Proposed (in-context literal):
    ```php
    $item = (new ContentItem())
        ->setValue('My SSN is 372819127');
    ```

bshaffer:
Makes sense. One of these days I'll get around to making these changes.

PR #895, opened 2019-05-22
Fix Cloud Spanner budget-transfer samples

hegemonic:
The `write_data_with_dml_transaction` function moved $200,000 from Album1 to Album2. However, Album1 is initialized with a budget of $100,000. As a result, you couldn't work through the [getting started doc](https://cloud.google.com/spanner/docs/getting-started/php/) from start to finish. This function also checked whether Album1's budget exceeded a minimum, which isn't consistent with the equivalent snippets for other languages.

In addition, the `read_write_transaction` function moved $200,000, but in the opposite direction, which was somewhat confusing.

Finally, neither function confirmed that the source album's budget was at least as large as the transfer amount.

I made both functions transfer $100,000 from Album1 to Album2. I also fixed the logic errors and updated the tests as needed.

### DIFF
--- a/spanner/src/read_write_transaction.php
+++ b/spanner/src/read_write_transaction.php
@@ -32,7 +32,7 @@ use UnexpectedValueException;
  * Performs a read-write transaction to update two sample records in the
  * database.
  *
- * This will transfer 200,000 from the `MarketingBudget` field for the second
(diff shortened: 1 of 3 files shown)


hegemonic:
@jsimonweb FYI

PR #925, opened 2019-06-24
Run a test suite with just REST tests

bshaffer:


bshaffer:
The tests that are failing are because of https://github.com/googleapis/google-cloud-php/issues/2064 (REST-only).

PR #937, opened 2019-08-01
[Storage] Add bucket_metadata sample

frankyn:
cc: @JesseLovelace

### DIFF
--- a/storage/storage.php
+++ b/storage/storage.php
@@ -125,7 +125,7 @@ EOF
 $application->add(new Command('buckets'))
     ->setDescription('Manage Cloud Storage buckets')
     ->setHelp(<<<EOF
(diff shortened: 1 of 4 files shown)


storage/storage.php
```diff
@@ -506,7 +524,8 @@
 
 $application->add(new Command('get-object-v4-upload-signed-url'))
     ->setDescription('Generate a v4 signed URL for uploading an object.')
-    ->setHelp(<<<EOF
+    ->setHelp(
```
bshaffer: seems fine to me, but why did you push all these EOFs down?

frankyn: php linter made me do it. :sweat_smile:

frankyn: reverting based on offline discussion

PR #938, opened 2019-08-02
Moves Flex helloworld to this repo

bshaffer:
See https://github.com/GoogleCloudPlatform/getting-started-php/pull/205

### DIFF
--- a/appengine/flexible/helloworld/test/ControllersTest.php
+++ b/appengine/flexible/helloworld/test/ControllersTest.php
@@ -1,7 +1,7 @@
 <?php
 
 /*
- * Copyright 2015 Google Inc. All Rights Reserved.
+ * Copyright 2019 Google LLC.
  *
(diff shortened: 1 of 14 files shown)


appengine/flexible/helloworld/test/ControllersTest.php
```diff
@@ -0,0 +1,54 @@
...
+    public function testTopPage()
+    {
+        $client = $this->createClient();
+        $crawlerexport = $client->request('GET', '/');
+        $resp = $client->getResponse();
+        $this->assertTrue($resp->isOk());
+        $this->assertContains('Hello World', $resp->getContent());
+    }
+
+    public function testBlog()
```
tmatsuo: Nit: Why this is called testBlog?

bshaffer:
Agreed about updating from Silex. I have a PR which updates the Bookshelf to be Lumen. We should upgrade all Silex as well with a future PR

PR #943, opened 2019-08-13
Update README.md

apacheborys:
For my opinion describing about Monolog bundle it's important, because Monolog is absent in standart symfony collection.


bshaffer:
This is incorrect. The tutorial specifies to install symfony like this:

```
$ PROJECT_DIR='symfony-on-appengine'
$ composer create-project symfony/symfony-demo:^1.2 $PROJECT_DIR
```

After you do this, you can see monolog is installed:

```
$ cd $PROJECT_DIR
$ ls -la vendor/
.
..
monolog/
```

PR #948, opened 2019-08-21
Update Firestore increment snippet name

jlara310:
fs_update_doc_array is already in use in https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/firestore/src/update_doc_array.php


jlara310:
@BenWhitehead

BenWhitehead:
Thanks for catching this @jlara310!

PR #952, opened 2019-09-06
run PHP fixer

gogasca:
Run PHP fixers in various files.


bshaffer:
Added some of these rules in https://github.com/GoogleCloudPlatform/php-docs-samples/pull/970

PR #961, opened 2019-09-17
Update composer.json

sirtorry:


language/composer.json
```diff
@@ -1,6 +1,7 @@
 {
     "require": {
         "google/cloud-language": "~0.18",
-        "google/cloud-storage": "^1.3"
+        "google/cloud-storage": "^1.3",
```
dwsupplee: This is unrelated to this change, but it doesn't look like the storage component is being used here anymore. It should be safe to drop :).

sirtorry:
@gabidavila pointed out that bcmath is required for the natural language api. unclear if this is also true for others.

gabidavila:
I noticed this, @dwsupplee, on a fresh PHP install that didn’t have `bcmath`. Not sure about other languages though

dwsupplee:
I believe it will be an issue with any GAPIC that uses HTTP/1.1 instead of gRPC until we have some type of fallback in place in protobuf. Some related issues:

https://github.com/protocolbuffers/protobuf/issues/4284
https://github.com/protocolbuffers/protobuf/issues/4465

bshaffer:
If that's the case, then `ext-bcmath` isn't actually required. One of `ext-bcmath` OR `ext-grpc` is required. So I don't think we should put this in there. 

Does GAPIC throw a good clear error message regarding this? Something like `The "bcmath" PHP extension is required when using HTTP 1.1 for Google Cloud API Calls`?

bshaffer:
I also don't believe `bcmath` is required, I believe it's either `bcmath` OR `gmp`. I am not seeing any hard requirement in the client library, but maybe @dwsupplee has more context.

bshaffer:
Either way, the requirement would be in the client library and not here, so I'm closing this PR.

dwsupplee:
> One of ext-bcmath OR ext-grpc is required. 

I should have also added that there isn't an issue when the protobuf C extension is installed, so installing that is an option. It may be a decent compromise here instead of bcmath as we recommend its usage elsewhere.

> I also don't believe bcmath is required, I believe it's either bcmath OR gmp.

Hmm, I believe it is just bcmath, there are [bc* calls](https://github.com/protocolbuffers/protobuf/blob/master/php/src/Google/Protobuf/Internal/Message.php#L900
) in methods utilized by [Message::serializeToJsonString()](https://github.com/protocolbuffers/protobuf/blob/master/php/src/Google/Protobuf/Internal/Message.php#L1489). Is bcmath a subset of gmp?
> Does GAPIC throw a good clear error message regarding this? Something like The "bcmath" PHP extension is required when using HTTP 1.1 for Google Cloud API Calls?

That's a decent idea 👍 . So far my hope has been to have this be resolved in protobuf itself. It would be nice to take some time to see if creating a pure PHP fallback for the bc* calls is viable.

> Either way, the requirement would be in the client library and not here, so I'm closing this PR.

I'm apprehensive to add this requirement to the client library as that would be breaking. Users who are currently using the gRPC/protobuf extensions successfully and do not have bcmath installed would be affected. This library has no semantic versioning constraints, so I was looking at this as a nice compromise to help guide users to the correct usage until the issue is resolved in protobuf. Once the issue is resolved, we could just drop the requirement with no problems. We could also look at adding bcmath as a suggestion in the protobuf library, but there's a good chance people might miss that.

bshaffer:
Added https://github.com/googleapis/google-cloud-php/issues/2400 to track this.

PR #963, opened 2019-09-19
adds execute permission to composer.sh

bshaffer:


CONTRIBUTING.md
```diff
@@ -42,11 +42,15 @@ accept your pull requests.
    export GOOGLE_STORAGE_BUCKET=YOUR_BUCKET
    ```
 
-   To run the tests in a samples directory,
+   To run the tests in a samples directory, first install the global Composer
+   dependencies by running `bash test/composer.sh`. Then, run `composer install`
+   and `phpunit` in any directory containing a `phpunit.xml.dist` file to run
+   the tests.
    ```
+   bash test/composer.sh
```
billyjacobson: if you are adding the other line to the .sh file, do you need to do "bash script.sh" or can you just do "script.sh"

bshaffer: With the shebang added, you can just do `./testing/composer.sh`, but either will work!

PR #964, opened 2019-09-20
Putting Hello World into one file for more clarity on docs page.

billyjacobson:


bigtable/src/hello_world.php
```diff
@@ -23,38 +23,48 @@
...
 
-if (count($argv) != 4) {
+if (count($argv) < 3 || count($argv) > 5) {
     return printf("Usage: php %s PROJECT_ID INSTANCE_ID TABLE_ID" . PHP_EOL, __FILE__);
 }
 list($_, $project_id, $instance_id, $table_id) = $argv;
 
-// [START bigtable_hw_create_table]
-
+// [START bigtable_hw]
```
billyjacobson: Is there a way to include the entire file without the above part on the cloud doc site? If so, I can remove this extra region tag

bigtable/src/hello_world.php
```diff
@@ -23,42 +23,53 @@
  */
 
 // Include Google Cloud dependencies using Composer
-require_once __DIR__ . '/../../vendor/autoload.php';
+require_once __DIR__ . '/../vendor/autoload.php';
 
-if (count($argv) != 4) {
+if (count($argv) < 3 || count($argv) > 5) {
```
bshaffer: This should stay `!=`, the two part conditional is only for when there are optional parameters!

bigtable/src/hello_world.php
bshaffer: You could include all the regions together and set `noheader=True`, and that would look essentially the same, but I'm personally okay with the nested tags!

PR #971, opened 2019-10-10
DO NOT MERGE: Talent v4beta1 samples

bshaffer:
Submitting this PR so we don't lose track of this branch, which *should* be merged into master.

These samples lack tests and contain sample inconsistencies with the rest of this repository, but are currently published in production docs.


fhinkel:
FYI, all PHP includes of this branch got deactivated in our docs, see e.g., cl/316717698. We should either add tests and active them in the docs again or delete this branch. For Node.js, we merged the branch without tests and are in the process of moving the docs to the samples in nodejs-talent instead of nodejs-docs-samples, which do have tests.

bshaffer:
These samples have been removed

PR #984, opened 2019-11-01
feat: add pubsub batch publishing sample

### LINKED ISSUE #978: pubsub: add sample to demonstrate batch publishing
In https://cloud.google.com/pubsub/docs/publisher#batching, there is no sample demonstrating batch publishing for PHP.

Create a sample for showing the use of [`batchPublisher`](https://github.com/googleapis/google-cloud-php/blob/master/PubSub/src/Topic.php#L469-L530). We will add a comment about running the batch daemon as a note in the docs alongside the sample.

jdpedrie:
Closes #978.

cc @hongalex

This is my first time working with the samples, so please forgive any nonsense contained below.

### DIFF
--- a/pubsub/api/src/publish_message_batch.php
+++ b/pubsub/api/src/publish_message_batch.php
@@ -1,0 +1,62 @@
+<?php
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/pubsub/api/README.md
+ */
+
+namespace Google\Cloud\Samples\PubSub;
+
+# [START pubsub_publisher_batch_settings]
+use Google\Cloud\PubSub\PubSubClient;
+
+/**
+ * Publishes a message for a Pub/Sub topic.
+ *
+ * The publisher should be used in conjunction with the `google-cloud-batch`
+ * daemon, which should be running in the background.
+ *
+ * To start the daemon, from your project root call `vendor/bin/google-cloud-batch daemon`.
+ *
+ * @param string $projectId  The Google project ID.
+ * @param string $topicName  The Pub/Sub topic name.
+ * @param string $message    The message to publish.
+ */
+function publish_message_batch($projectId, $topicName, $message)
+{
+    $batchOptions = [
+        'batchSize' => 10, // Max messages for each batch.
+        'callPeriod' => 2.0, // Max time in seconds between each batch publish.
+    ];
+
+    $pubsub = new PubSubClient([
+        'projectId' => $projectId,
+    ]);
+    $topic = $pubsub->topic($topicName);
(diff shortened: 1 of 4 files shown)


pubsub/api/src/publish_message_batch.php
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @param string $projectId  The Google project ID.
+ * @param string $topicName  The Pub/Sub topic name.
+ * @param string $message    The message to publish.
+ */
+function publish_message_batch($projectId, $topicName, $message)
+{
+    $batchOptions = [
+        'batchSize' => 10, // Max messages for each batch.
+        'callPeriod' => 2.0, // Max time in seconds between each batch publish.
```
hongalex: For these, can we change this to 100 messages for `batchSize` and 0.01 (10 ms) for `callPeriod`? This is based on the recent changes we want to make to client library defaults.

jdpedrie: The sample uses the publisher without the daemon running. This adds some constraints which I'm finding a bit difficult to work around.

The batch publisher is filled with all ten messages immediately, but as neither the batch size nor the interval have been reached, it doesn't flush the messages to Cloud Pub/Sub. As there is no process watching for messages (i.e. the daemon), nothing will happen until another call is made to the batch publisher. I set the max messages to 10 in order to cause the final publish call to flush the queue.

I see a couple options.
1. I could pass the values needed by the unit test in as function arguments and document the defaults you mention in the comments on lines 44 and 45.
2. I could run the daemon in the unit test. I'm not sure this would work. It would have to run in the background, and I'd need to see whether PHP would kill it once the test run finishes.

Option 1 is definitely easier, but if you prefer the 2nd option, which would allow for the defaults you desire, I can investigate whether it's workable.

hongalex: My main concern is that that we want our samples to be copy/pastable with pretty good settings. The intention of lowering the `callPeriod` is to reduce the perception of lag/latency, especially if the user is publishing < `batchSize` every 2 seconds. I'm ok with option 1 with leaving `batchSize` at 10 (and documenting the default). However, I still would prefer if `callPeriod` was reduced to something lower (and also documenting the default).

jdpedrie: @hongalex I was able to find a way to run the batch daemon in the background of the unit test, so I updated the test to use the desired defaults.

jdpedrie: @hongalex I was able to find a way to run the daemon in the background of the unit test, so I've updated the sample to use the desired defaults.

hongalex:
Looks good, but I'll leave David to comment on PHP specific things I might have missed.

pubsub/api/src/publish_message_batch.php
```diff
@@ -0,0 +1,62 @@
...
+ * The publisher should be used in conjunction with the `google-cloud-batch`
+ * daemon, which should be running in the background.
+ *
+ * To start the daemon, from your project root call `vendor/bin/google-cloud-batch daemon`.
+ *
+ * @param string $projectId  The Google project ID.
+ * @param string $topicName  The Pub/Sub topic name.
+ * @param string $message    The message to publish.
+ */
+function publish_message_batch($projectId, $topicName, $message)
```
bshaffer: This method should check if the batch publisher is running, and print a warning method with instructions on how to run it if it isn't.

jdpedrie: @bshaffer done. I used a notice because you can still use the batch publisher without the daemon, with the various caveats I detailed in my convo with Alex.

pubsub/api/src/publish_message_batch.php
```diff
@@ -0,0 +1,72 @@
...
+{
+    // Check if the batch daemon is running.
+    if (getenv('IS_BATCH_DAEMON_RUNNING') !== "true") {
```
dwsupplee: Just a minor nit, but would we be able to be consistent and use single quotes please?

jdpedrie: duh duh duh done

PR #988, opened 2019-11-12
Update Cloud Tasks to new library

averikitsch:


appengine/php72/tasks/snippets/composer.json
```diff
@@ -1,5 +1,9 @@
 {
     "require": {
-        "google/cloud-tasks": "^1.0.0"
+        "google/cloud-tasks": "^1.4.0"
+    },
+    "require-dev": {
```
bshaffer: We do not want these deps here, they are installed globally

PR #1002, opened 2019-12-19
Exclude vendor dir from code coverage

ace-n:
✨This PR will greatly simplify + speed up [automated test labeling efforts](https://github.com/GoogleCloudPlatform/repo-automation-playground/pull/11).

### DIFF
--- a/appengine/flexible/analytics/phpunit.xml.dist
+++ b/appengine/flexible/analytics/phpunit.xml.dist
@@ -29,6 +29,9 @@
     <filter>
         <whitelist>
             <file>app.php</file>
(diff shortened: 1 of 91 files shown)


appengine/flexible/analytics/phpunit.xml.dist
```diff
@@ -29,6 +29,9 @@
     <filter>
         <whitelist>
             <file>app.php</file>
+            <exclude>
+                <directory suffix=".php">./vendor</directory>
```
bshaffer: Do we need `suffix=".php"` in there?

PR #1003, opened 2019-12-22
Added php7.4 flexible environment sample

thiagocordeiro:
Adds a sample how to setup an AppEngine application with PHP 7.4, even though it uses a specific php version it's possible to change to any other as official PHP image well as any custom docker image.

The sample uses Alpine, changing from alpine to any debina base image is also quite similar.

Dockerfile installs opcache extension with a basic setup (which is enough for having requests below 20ms on a symfony app), nginx as webserver and supervisor to make sure nginx and php-fpm keep running.

This setup approach follows the standard/supported PHP versions (5.6, 7.0, 7.1 and 7.2) for AppEngine.

https://cloud.google.com/appengine/docs/php/

### DIFF
--- a/appengine/php74/.alpine/etc/nginx/conf.d/default.conf
+++ b/appengine/php74/.alpine/etc/nginx/conf.d/default.conf
@@ -1,0 +1,20 @@
+server {
+  server_name localhost;
+  listen      8080;
+  root        /var/www/html/public;
+  index       index.php;
+
+  error_log  /var/log/nginx/error.log;
+  access_log /var/log/nginx/access.log;
+
+  location / {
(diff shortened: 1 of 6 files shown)


bshaffer:
@thiagocordeiro Thank you for your contribution! This would make a much better [Google Cloud Community Tutorial](https://cloud.google.com/community/tutorials/write), as samples here are specific to documentation published on our official website. Are you interested in writing a community tutorial?

thiagocordeiro:
Sure @bshaffer :)

bshaffer:
Thank you! Please let me know if you have any other questions!

PR #1025, opened 2020-02-18
Update JobSearchHistogramSearch.php to actually pass in the queries

hilts-vaughan:
.. it wasn't before. I think the Ruby one might need it as well.

https://cloud.google.com/talent-solution/job-search/docs/histogram#job_search_histogram_search-ruby page as well never uses https://github.com/googleapis/google-cloud-php-talent/blob/master/src/V4beta1/SearchJobsResponse.php#L167 to show how to actually get the data off from the response -- should it? 

I can update the docs / file a bug if needed.

### DIFF
--- a/jobs/v4/JobSearchHistogramSearch.php
+++ b/jobs/v4/JobSearchHistogramSearch.php
@@ -56,7 +56,8 @@ function sampleSearchJobs($projectId, $tenantId, $query)
 
     try {
         // Iterate through all elements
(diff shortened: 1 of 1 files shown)


hilts-vaughan:
Just saw `job_search_histogram_search`... will need to correct upstream.

hilts-vaughan:
https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1027 for all of this

PR #1027, opened 2020-02-18
cloud talent: Updated V4 Beta 1 docs

hilts-vaughan:
Regenerated the docs via synthtool

These are currently not right on the production site. Is there any plans to get these into master?

### DIFF
--- a/jobs/v4/JobSearchAutocompleteJobTitle.php
+++ b/jobs/v4/JobSearchAutocompleteJobTitle.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2019 Google LLC
+ * Copyright 2020 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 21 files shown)


bshaffer:
See #971, essentially these samples were pushed to the `talent-v4better1-samples` branch and have not been reviewed. The branch was published to our documentation and that's all I know. .

The issue is some of the sample generation needs to be cleaned up. The `$opts` array for example, is not the way we want to handle this. So until we can find someone to work with to make the required changes, I do not think we can merge it into `master`. However, I am more than happy to work with someone to make the required changes, I do not think it will be a large effort.

bshaffer:
I'd like to see these samples [more closely match our other samples](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/bigquery/api/src/browse_table.php#L25) before merging into master.

hilts-vaughan:
Could you summarize the changes you want to see? I imagine a lot of it comes down to the generation tool being used.

bshaffer:
@hilts-vaughan I've been told the generation tool is not going to be maintained going forward, so I think it would make more sense to just manually change the files here.

bshaffer:
closing these because PHP samples have been removed for this API

PR #1022, opened 2020-02-13
Add documentation for sending emails with Symfony on GAE

Sethbass:
Hi @bshaffer,

As discussed in the issue #1021 , I have enriched the documentation about the emails with Symfony and GAE. 

Cheers,

Sebastien

PR #1030, opened 2020-02-24
Add documentation for sending emails with Symfony on GAE

bshaffer:
Resolves committer issues in #1022


Sethbass:
@googlebot I consent

Sethbass:
@bshaffer On the first commit I was using the profile of another project instead of github... I found the trick a bit too late but everything seems ok now thanks to you ;)

PR #1041, opened 2020-03-12
feat: add query optimizer version samples

larkee:


spanner/src/create_client_with_query_options.php
```diff
@@ -0,0 +1,60 @@
+<?php
+/**
+ * Copyright 2019 Google LLC.
```
hengfengli: 2020?

spanner/src/query_data_with_query_options.php
```diff
@@ -0,0 +1,61 @@
+<?php
+/**
+ * Copyright 2019 Google LLC.
```
hengfengli: Same here.

PR #1046, opened 2020-03-24
feat: spanner backup and restore code samples

ava12:


spanner/src/list_backup_operations.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * Copyright 2019 Google Inc.
```
skuruppu:
```suggestion
 * Copyright 2020 Google Inc.
```

spanner/src/list_backups.php
```diff
@@ -0,0 +1,99 @@
+<?php
+/**
+ * Copyright 2019 Google Inc.
```
skuruppu:
```suggestion
 * Copyright 2020 Google Inc.
```

spanner/src/list_backups.php
```diff
@@ -0,0 +1,99 @@
...
+    // List all backups with a size greater than some bytes.
+    $size = 500;
+    print("All backups with size greater than $size bytes:" . PHP_EOL);
+    $filter = "size_bytes > $size";
+    foreach ($instance->backups(['filter' => $filter]) as $backup) {
+        print('  ' . basename($backup->name()) . PHP_EOL);
+    }
+
+    // List backups that were created after a timestamp that are also ready.
+    $createTime = $spanner->timestamp(new \DateTime('-1 hour'));
```
skuruppu: Might need to make this -1 day just to be safe for the tests in case the backups take more than an hour to create.

spanner/src/list_database_operations.php
```diff
@@ -0,0 +1,57 @@
+<?php
+/**
+ * Copyright 2019 Google Inc.
```
skuruppu:
```suggestion
 * Copyright 2020 Google Inc.
```

spanner/src/list_database_operations.php
```diff
@@ -0,0 +1,57 @@
...
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/spanner/README.md
+ */
+
+namespace Google\Cloud\Samples\Spanner;
+
+// [START spanner_list_database_operations]
+use Google\Cloud\Spanner\SpannerClient;
+
+/**
+ * List all database operations in an instance.
```
skuruppu:
```suggestion
 * List all optimize restored database operations in an instance.
```

spanner/src/list_backup_operations.php
```diff
@@ -0,0 +1,58 @@
...
+    $filter = "(metadata.database:$databaseId) AND " .
+              "(metadata.@type:type.googleapis.com/" .
+              "google.spanner.admin.database.v1.CreateBackupMetadata)";
+
+    $operations = $instance->backupOperations(['filter' => $filter]);
+
+    foreach ($operations as $operation) {
+        if (!$operation->done()) {
+            $operationName = basename($operation->info()['name']);
+            print("Backup operation $operationName pending" . PHP_EOL);
```
skuruppu:
```suggestion
            print("Backup $metadataName on database $metadataDatabase is $progressPercent% pending" . PHP_EOL);
```
I'm not sure how to get the metadata out of the operation in PHP but all the necessary info should be in there like the name, database and progress_percent.

spanner/src/list_database_operations.php
```diff
@@ -0,0 +1,57 @@
...
+    $filter = "(metadata.@type:type.googleapis.com/" .
+              "google.spanner.admin.database.v1.OptimizeRestoredDatabaseMetadata)";
+
+    $operations = $instance->databaseOperations(['filter' => $filter]);
+
+    foreach ($operations as $operation) {
+        if (!$operation->done()) {
+            $operationName = basename($operation->name());
+            var_dump($operation->info());
+            print("Running optimize operation: $operationName" . PHP_EOL);
```
skuruppu:
```suggestion
            print("Database $metadataName restored from backup is $progressPercent% optimized." . PHP_EOL);
```

Same here, I don't actually know how to get the metadata out of the operation in PHP but it should all be in there.

spanner/src/restore_backup.php
```diff
@@ -0,0 +1,58 @@
+<?php
+/**
+ * Copyright 2019 Google Inc.
```
skuruppu:
```suggestion
 * Copyright 2020 Google Inc.
```

spanner/src/update_backup.php
```diff
@@ -0,0 +1,50 @@
+<?php
+/**
+ * Copyright 2019 Google Inc.
```
skuruppu:
```suggestion
 * Copyright 2020 Google Inc.
```

spanner/test/spannerTest.php
```diff
@@ -606,6 +614,130 @@ public function testCreateClientWithQueryOptions()
...
+        self::$instance->database($db2)->create();
+        $backup = self::$instance->backup(self::$backupId . '-pro');
+        $lro = $backup->create($db2, new \DateTime('+7 hours'));
+        $output = $this->runCommand('list-backup-operations');
+        $this->assertContains(basename($lro->name()), $output);
+        $lro->pollUntilComplete();
+    }
+
+    /**
+     * @depends testListBackupOperations
```
skuruppu: I think this can just depend on testCreateBackup.

spanner/test/spannerTest.php
```diff
@@ -606,6 +614,130 @@ public function testCreateClientWithQueryOptions()
...
+    {
+        $output = $this->traitRunCommand('list-backups', [
+            'instance_id' => self::$instanceId,
+        ]);
+        $this->assertContains(self::$backupId, $output);
+        $this->assertContains(self::$backupId . '-pro', $output);
+    }
+
+    /**
+     * @depends testListBackups
```
skuruppu: Same here, can depend on testCreateBackup.

spanner/src/list_backup_operations.php
ava12: Current implementation returns raw, binary-encoded metadata as part of operation info. I don't know how to decode it. Looks like an oversight to me, but this implementation is approved.

skuruppu: @AVaksman would you be able to help with this? The reason why we wanted to do this was to show users through the sample how to decode the metadata because it's not obvious. I don't know how it would work in PHP but it would be good to make this consistent with that of the other languages. E.g. in [Node.js](https://github.com/googleapis/nodejs-spanner/blob/e3e454dd9484d70acada78a1ab1feac68346ebac/samples/backups-get-database-operations.js#L45).

skuruppu: Friendly ping on this comment. We need the samples to be consistent across the languages before they can be merged. We have to get this submitted within the next couple of days.

ava12: Couple of days? That will be hard.
@AVaksman proposes ([#2536](https://github.com/googleapis/google-cloud-php/pull/2536)) some changes inside Core to solve this problem, but that solution has problems itself.

skuruppu: Wow ok, I understand why it's hard then. Thanks for the background info. I'm happy to approve this as is but I also filed #1050 to fix the samples once you're able to decode the metadata.

skuruppu:
@ava12 the tests seem to be flaky possibly because the tests for the different versions of PHP are running in parallel but reusing the same instance. We can only do one restore at a time per instance so this failure happens:

```
1) Google\Cloud\Samples\Spanner\spannerTest::testRestoreBackup
Google\Cloud\Core\Exception\FailedPreconditionException: {
    "message": "Cannot restore database (projects\/php-docs-samples-kokoro\/instances\/kokoro-testing\/databases\/test-15852602531984820283-res) because the maximum number of pending restores (1) for the instance (projects\/php-docs-samples-kokoro\/instances\/kokoro-testing) has been reached. Please retry the operation once the pending restores complete.",
    "code": 9,
    "status": "FAILED_PRECONDITION",
    "details": [
        {
            "@type": "grpc-server-stats-bin",
            "data": "<Unknown Binary Data>"
        }
    ]
}
```

ava12:
Yes, the same problem with backup creation. I'll add a wrapper that catches this exception and reruns the sample.

spanner/test/spannerTest.php
```diff
@@ -737,6 +740,33 @@ private static function waitForDatabaseOperations($instance) {
...
+     */
+    private function retry($call, $args)
+    {
+        $cutoffTime = time() + 20 * 60;
+        $sleepDuration = 10;
+
+        while (true) {
+            try {
+                return call_user_func_array($call, $args);
+
```
skuruppu, line 759:
```suggestion
```

skuruppu:
> @skuruppu You merged in changes which broke master because you merged this PR before the tests pass. Please do NOT merge in changes without getting permissions from the owners of this repo. And it goes without saying, but **especially** if the tests are failing.

Thanks for reaching out @bshaffer. According to the CI build results on the merge commit, the tests were [passing](https://source.cloud.google.com/results/invocations/b831c0fa-6a38-4d14-93c9-a6ddb1d29842/targets). So I'm not sure what you're referring to. Would you be able to share a link?

Also, I'm not sure why but this repo doesn't seem to assign any repo owners as reviewers automatically. Who should we be adding in the future? Can we automate this?

PR #1047, opened 2020-03-24
Update detect_object.php

teg1:
Use float specifier so that normalized vertices and confidence score are printed out.
Otherwise we get all zeroes for the vertices and confidence score.


teg1:
@googlebot I signed it!

bshaffer:
@teg1 Thank you for finding this and submitting the fix!

PR #1048, opened 2020-03-24
Cleanup unused samples and affected code

Strykrol:
Removing PHP samples that aren't being surfaced to cloud.google.com. I made some best guesses on the corresponding test and configuration file adjustments to align with these deletions.


bshaffer:
It seems unnecessary to remove working well-tested samples just because they're not in the docs. What is the justification for this?

Strykrol:
It's part of a clean-up to reconcile samples to a state more catered around UX, whether that is by finding a sensible spot in docs or deleting the clutter from a samples repository where sample context isn't necessarily prevalent. I can provide the internal project link if you want to reach out beyond this PR.

FWIW the dialogflow-specific deletion decisions were signed off by @nnegrey who may have additional insight. Dialogflow sample deletions are occurring across all language/repos.

bshaffer:
I'd rather just remove the region tags (to indicate these are not in docs) and leave the working samples alone. We can even flag them in the header as not being used in the docs so that we can remove them down the road if they cause issues, but because they're not requiring any maintenance cost and they're working and fully tested, I'd rather keep them around in case someone stumbles upon them and finds them useful

Strykrol:
If you want to keep the code, then we might as well keep the region tags for _potential_ future use. I can close this PR with no changes performed.

PR #1053, opened 2020-04-06
feat: add sqlserver sample

jdpedrie:
Closes #992.

PR #1057, opened 2020-04-10
feat: add mysql sample

### LINKED ISSUE #991: Create Cloud SQL connectivity sample for MySQL
Create a copy of the [Postgres connectivity sample](https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/cloud_sql/postgres/pdo) and update it to work with MySQL.

Requirements:

- PR the sample into this repo under `cloud_sql/mysql/pdo`
- Update the following region tag(s): 
    * `cloud_sql_postgres_pdo_create` -> `cloud_sql_mysql_pdo_create`
- Update the README for MySQL and verify the instructions work

Please add myself as a reviewer.

Assigned: @jdpedrie

FYI: @dmahugh, @jsimonweb

jdpedrie:
Closes #991.

cc @kurtisvg.


cloud_sql/mysql/pdo/src/DB.php
```diff
@@ -0,0 +1,54 @@
...
+            $dsn = sprintf(
+                'mysql:dbname=%s;unix_socket=/cloudsql/%s',
+                $schema,
+                $cloud_sql_connection_name
+            );
+        } else {
+            // Connect using TCP
+            // $hostname = '127.0.0.1';
+            $dsn = sprintf('mysql:dbname=%s;host=%s', $schema, $hostname);
+        }
```
kurtisvg, line 49: @shubha-rajan - I'll defer to you if we want to figure out how to handle split TCP/Unix socket connections snippets in this PR or fix all the PHP samples together later.

cloud_sql/mysql/pdo/index.php
```diff
@@ -0,0 +1,42 @@
...
+    include_once("./template.php");
+} elseif ($_SERVER['REQUEST_URI'] == '/' && $_SERVER['REQUEST_METHOD'] == 'POST') {
+    $message = 'Invalid vote. Choose Between TABS and SPACES';
+
+    if (!empty($_POST['team']) && in_array($_POST['team'], ['SPACES', 'TABS'])) {
+        $message = $votes->save($_POST['team']);
+    }
+
+    echo $message;
+}
```
grayside, line 42: This is missing 404/405 route error handling, so any other requests sent to the service will receive a 200 OK.

Something like the following:

```php
elseif ($_SERVER['REQUEST_URI'] == '/') {
  return http_response_code(405);
}
else {
  return http_response_code(404);
}
```

jdpedrie:
@grayside @kurtisvg would you like me to apply all these suggestions to the [postgres](https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/cloud_sql/postgres/pdo) samples as well, which provided the template for this and the sqlserver PRs?

kurtisvg:
> @grayside @kurtisvg would you like me to apply all these suggestions to the [postgres](https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/cloud_sql/postgres/pdo) samples as well, which provided the template for this and the sqlserver PRs?

Yes please!

cloud_sql/mysql/pdo/README.md
```diff
@@ -0,0 +1,66 @@
...
+```bash
+export GOOGLE_APPLICATION_CREDENTIALS=/path/to/service/account/key.json
+export CLOUD_SQL_CONNECTION_NAME='<MY-PROJECT>:<INSTANCE-REGION>:<MY-DATABASE>'
+export DB_USER='my-db-user'
+export DB_PASS='my-db-pass'
+export DB_NAME='my-db-name'
+export DB_HOSTNAME='localhost'
+```
+
+Note: Saving credentials in environment variables is convenient, but not secure - consider a more secure solution such as [Cloud KMS](https://cloud.google.com/kms/) to help keep secrets safe.
```
shubha-rajan, line 20: Nit: We're moving towards suggesting [Secret Manager](https://cloud.google.com/secret-manager/docs/overview) instead of KMS.

cloud_sql/mysql/pdo/src/DB.php
shubha-rajan: I think we should figure out how to handle TCP/Unix socket connections and update the postgres sample for consistency as soon as this is merged.  One note I have right now is that if we're checking for the existence of `CLOUD_SQL_CONNECTION_NAME` to decide whether to use TCP or Unix Sockets, we should add a note in the readme to not set that env variable if they want to use TCP.

jdpedrie:
@grayside @shubha-rajan @kurtisvg @bshaffer I refactored the app to use the Slim 4 framework used [elsewhere](https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/appengine/php72/cloudsql) in the PHP docs samples. I believe I've addressed all the comments, but I haven't written any tests yet. I'd like to make sure you are all in favor of the new direction before I wrap it up. Once everyone is happy, I'll also refactor the sqlserver and postgres samples in a similar direction.

cloud_sql/mysql/pdo/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Execute the following:
+
+```bash
+$ php -S localhost:8080
+```
+
+Navigate towards http://localhost:8080 to verify your application is running correctly.
+
+## Google App Engine Standard
```
jdpedrie, line 62: (todo https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1053#discussion_r408338349 )

cloud_sql/mysql/pdo/src/app.php
```diff
@@ -0,0 +1,94 @@
...
+                $schema,
+                $cloud_sql_connection_name
+            );
+        } else {
+            // Connect using TCP
+            // $hostname = '127.0.0.1';
+            $dsn = sprintf('mysql:dbname=%s;host=%s', $schema, $hostname);
+        }
+
+        $conn = new PDO($dsn, $username, $password);
```
grayside, line 64: In the event this fails on something other than a PDOException, this is pretty messy:

Screen Shot 2020-04-23 at 9 03 27 PM

The problem in this case is a TypeError because I didn't configure the environment values. We could catch a TypeError and say "wake up and pay attention to the env values", but maybe a more generic alternative to `PDOException` in `Exception` would be appropriate as a catch-all? (The risk of this is more pronounced in local development or in a future adaptation of this sample for Cloud Run.)

Configuring with non-working values was successfully caught by the exception handler.

cloud_sql/mysql/pdo/index.php
```diff
@@ -0,0 +1,50 @@
...
+declare(strict_types=1);
+
+use GuzzleHttp\Psr7;
+
+include __DIR__ . '/vendor/autoload.php';
+
+$app = include __DIR__ . '/src/app.php';
+
+$app->get('/', function ($request, $response) {
+    return $this->get('view')->render($response, 'template.twig', [
```
grayside, line 27: I like this pattern of thin controller and nicely separated model and view :)

What happens if the twig template fails to compile (e.g., syntax error) or a query fails?

cloud_sql/mysql/pdo/views/template.twig
```diff
@@ -0,0 +1,99 @@
...
+            <div class="card-panel {{ tabCount < spaceCount ?: 'blue lighten-3' }}">
+                <i class="material-icons large">space_bar</i>
+                <h3>{{ spaceCount }} votes</h3>
+                <button id="voteSpaces" class="btn blue">Vote for SPACES</button>
+            </div>
+        </div>
+    </div>
+    <h4 class="header center">Recent Votes</h4>
+    <ul class="container collection center">
+        {% for vote in votes %}
```
grayside, line 62: nit: If there are no votes yet is there something different we want to do?

cloud_sql/mysql/pdo/index.php
jdpedrie: If it raises an exception, a similar error handling screen to the Slim one elsewhere.

cloud_sql/mysql/pdo/src/app.php
jdpedrie: I've added a TypeError catch to improve the error message users will get.

cloud_sql/mysql/pdo/views/template.twig
jdpedrie: Good idea! I added an if/else and a "No votes have been cast" message.

cloud_sql/mysql/pdo/src/app.php
```diff
@@ -0,0 +1,94 @@
...
+        if ($cloud_sql_connection_name) {
+            // Connect using UNIX sockets
+            $dsn = sprintf(
+                'mysql:dbname=%s;unix_socket=/cloudsql/%s',
+                $schema,
+                $cloud_sql_connection_name
+            );
+        } else {
+            // Connect using TCP
+            // $hostname = '127.0.0.1';
```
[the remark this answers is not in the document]
jdpedrie: I removed it here and added it to the list of commented-out variables above with a note advising users to only set it in TCP mode.

cloud_sql/mysql/pdo/views/template.twig
```diff
@@ -0,0 +1,99 @@
...
+                    A vote for <b>{{ vote.vote_value }}</b>
+                </span>
+            <p>was cast at {{ vote.time_cast }}</p>
+        </li>
+        {% endfor %}
+    </ul>
+</div>
+<script>
+    function vote(voteValue) {
+        var xhr = new XMLHttpRequest();
```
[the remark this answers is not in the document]
kurtisvg: /cc @shubha-rajan - might be good to add to our list of updates

cloud_sql/mysql/pdo/src/app.php
```diff
@@ -0,0 +1,111 @@
...
+                'mysql:dbname=%s;unix_socket=/cloudsql/%s',
+                $dbName,
+                $cloud_sql_connection_name
+            );
+        } else {
+            // Connect using TCP
+            $dsn = sprintf('mysql:dbname=%s;host=%s', $dbName, $hostname);
+        }
+
+        $conn = new PDO($dsn, $username, $password);
```
shubha-rajan, line 64: Could this be configured to set a connection timeout? It might be a good idea to show an example of that in the sample.

cloud_sql/mysql/pdo/src/app.php
```diff
@@ -0,0 +1,111 @@
...
+// Setup the database connection in the container.
+$container['db'] = function () {
+    $username = getenv('DB_USER');
+    $password = getenv('DB_PASS');
+    $dbName = getenv('DB_NAME');
+    $hostname = getenv('DB_HOSTNAME') ?: '127.0.0.1';
+    $cloud_sql_connection_name = getenv('CLOUD_SQL_CONNECTION_NAME');
+
+    try {
+        // # [START cloud_sql_mysql_pdo_create]
```
shubha-rajan, line 45: The other languages' samples on the Managing Connections page where this sample will be published implement connection pools, but from quickly researching (I'm not too familiar with PHP), there isn't really a way to use pooling with PHP/PDO. Can you include a comment explaining how PDO handles connections and why we're not using pooling here, to make it less confusing in the context of the page? (https://cloud.google.com/sql/docs/mysql/manage-connections)

cloud_sql/mysql/pdo/src/Votes.php
```diff
@@ -0,0 +1,103 @@
...
+     * @return boolean
+     */
+    public function insertVote(string $value) : bool
+    {
+        $sql = "INSERT INTO votes (time_cast, vote_value) VALUES (NOW(), :voteValue)";
+        $statement = $this->connection->prepare($sql);
+        $statement->bindParam('voteValue', $value);
+
+        return $statement->execute();
+    }
```
shubha-rajan, line 102: This might need to be wrapped in `cloud_sql_mysql_pdo_connection` region tags with a comment about how PDO handles opening and closing connections. 
See the python sample here for an example: https://cloud.google.com/sql/docs/mysql/manage-connections#open-close-python

cloud_sql/mysql/pdo/src/app.php
jdpedrie: Sure! I added connection timeout and also demonstrated how to set PDO to throw exceptions by default if any errors occur.

cloud_sql/mysql/pdo/src/app.php
jdpedrie: Some PDO drivers support connection pooling ([sqlserver](https://docs.microsoft.com/en-us/sql/connect/php/connection-pooling-microsoft-drivers-for-php-for-sql-server?view=sql-server-ver15) for instance) via driver configuration (not configured at the application level).

In general usage it's a bit less of a concern for PHP due to the manner in which PHP lifecycles work. PHP applications are usually not long running, they start at the beginning of request processing and more or less shut down at the end (this is a bit tricky because certain runtimes like php-fpm don't fully shut down). Even if they don't fully shut down, all application state is lost, including database connections.

Event loops are becoming more common in PHP, and that's a case where connection pooling begins to be more of a requirement. In the case of PDO drivers which don't support pooling, I've seen cases where the application will simply manage a pool of `PDO` instances. `new PDO()` connects immediately, at least for mysql. If anything is mentioned, it should be this option, I believe.

The other alternative is persistent connections, but the pitfalls there are rather dangerous and it's not something I think we should recommend.

cloud_sql/mysql/pdo/src/Votes.php
shubha-rajan: That looks great. Thank you.

jdpedrie:
@grayside could you please take another look?

cloud_sql/mysql/pdo/src/app.php
```diff
@@ -0,0 +1,116 @@
...
+
+        // Connect to the database.
+        // Here we set the connection timeout to five seconds and ask PDO to
+        // throw an exception if any errors occur.
+        $conn = new PDO($dsn, $username, $password, [
+            PDO::ATTR_TIMEOUT => 5,
+            PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
+        ]);
+        # [END cloud_sql_mysql_pdo_create]
+    } catch (TypeError $e) {
```
grayside, line 72: I'm loving the troubleshooting guidance in this error handling.

shubha-rajan:
@jdpedrie looks like a checkstyle issue is causing the build to fail.  I can merge this as soon as that's fixed and the tests pass.

image

PR #1082, opened 2020-04-15
fix: fixes storage samples

bshaffer:


storage/src/remove_bucket_conditional_iam_binding.php
```diff
@@ -40,7 +40,7 @@
  *
  * @return void
  */
-function remove_bucket_conditional_iam_binding($bucketName, $role, $members, $title, $description, $expression)
+function remove_bucket_conditional_iam_binding($bucketName, $role, $title, $description, $expression)
```
bshaffer, line 43: `$members` is unused, and should not be a function argument. Especially since the invocation in `storage.php` does not pass it in.

storage/src/remove_bucket_conditional_iam_binding.php
```diff
@@ -51,11 +51,11 @@ function remove_bucket_conditional_iam_binding($bucketName, $role, $members, $ti
 
     $key_of_conditional_binding = null;
     foreach ($policy['bindings'] as $key => $binding) {
-        if ($binding['role'] == $role && $binding['condition'] != null) {
+        if ($binding['role'] == $role && isset($binding['condition'])) {
```
bshaffer, line 54: `isset` needs to be used. With the previous implementation, a PHP warning is issued when the index `condition` does not exist.

storage/src/remove_bucket_conditional_iam_binding.php
```diff
@@ -51,11 +51,11 @@ function remove_bucket_conditional_iam_binding($bucketName, $role, $members, $ti
...
             $condition = $binding['condition'];
             if ($condition['title'] == $title
-                 && $condition['description'] == description
-                 && $condition['expression'] == expression) {
+                 && $condition['description'] == $description
+                 && $condition['expression'] == $expression) {
```
bshaffer, line 58: the variables `$description` and `$expression` were missing the `$` variable declaration, causing them to be interpreted as undefined constants

storage/src/remove_bucket_iam_member.php
```diff
@@ -50,10 +50,14 @@ function remove_bucket_iam_member($bucketName, $role, $member)
             if ($key !== false) {
                 unset($binding['members'][$key]);
 
-                // If the last member is removed from the binding, clean up
-                // the binding.
+                // If the last member is removed from the binding, clean up the
+                // binding.
                 if (count($binding['members']) == 0) {
                     unset($policy['bindings'][$i]);
+                    $policy['bindings'] = array_values($policy['bindings']);
```
bshaffer, line 57: Without this block, `json_encode` interprets `$policy['bindings']` as an object instead of an array when the keys get out of order, thus sending an empty array to the API which erases the policy's bindings

storage/src/remove_bucket_iam_member.php
```diff
@@ -50,10 +50,14 @@ function remove_bucket_iam_member($bucketName, $role, $member)
...
+                } else {
+                    $binding['members'] = array_values($binding['members']);
+                    $policy['bindings'][$i] = $binding;
```
bshaffer, line 60: Without this block, `json_encode` interprets `$binding['members']` as an object instead of an array when the keys get out of order, thus sending an empty array to the API which erases the binding's members

storage/storage.php
```diff
@@ -318,7 +318,7 @@
             if (!$expression) {
                 throw new InvalidArgumentException('Must provide expression as an option.');
             }
-            remove_bucket_conditional_iam_binding($bucket_name, $role, $title, $description, $expression);
+            remove_bucket_conditional_iam_binding($bucketName, $role, $title, $description, $expression);
```
bshaffer, line 321: the variable `$bucket_name` doesn't exist

storage/test/IamCommandTest.php
```diff
@@ -59,11 +60,10 @@ private static function cleanUpIam()
         $iam = $bucket->iam();
 
         $policy = $iam->policy(['requestedPolicyVersion' => 3]);
-        $roles = ['roles/storage.objectViewer', 'roles/storage.objectCreator'];
```
bshaffer, line 62: The tests were conflating `roles/storage.objectViewer` and `roles/storage.objectCreator`

storage/test/IamCommandTest.php
```diff
@@ -133,14 +131,16 @@ public function testAddBucketConditionalIamBinding()
...
+                if (in_array(self::$user, $binding['members']) &&
                     isset($binding['condition']) &&
                     $binding['condition']['title'] == $title &&
                     $binding['condition']['description'] == $description &&
-                    $binding['condition']['expression'] == $expression;
-                break;
+                    $binding['condition']['expression'] == $expression
+                ) {
+                    $foundBinding = true;
+                    break;
```
bshaffer, line 142: Because there can be multiple roles of the same name per binding, this logic was breaking after the first match, making `$foundBinding` false even if another binding would have made it true.

storage/test/IamCommandTest.php
```diff
@@ -186,28 +186,30 @@ public function testListIamMembers()
...
 
         $foundRoleMember = false;
         $policy = self::$storage->bucket(self::$bucket)->iam()->policy([
             'requestedPolicyVersion' => 3
         ]);
         foreach ($policy['bindings'] as $binding) {
-            if ($binding['role'] == $role) {
+            if (
+                $binding['role'] == self::$role
+                && empty($binding['condition'])
```
bshaffer, line 211: Without checking that the conditional is null, this was resulting in false positives

storage/test/IamCommandTest.php
```diff
@@ -221,29 +223,32 @@ public function testRemoveBucketIamMember()
...
 
         $foundBinding = false;
         $policy = self::$storage->bucket(self::$bucket)->iam()->policy([
             'requestedPolicyVersion' => 3
         ]);
         foreach ($policy['bindings'] as $binding) {
-            if ($binding['role'] == $role && $binding['condition'] != null) {
+            if (
+                $binding['role'] == self::$role
+                && isset($binding['condition'])
```
bshaffer, line 250: again, `isset` must be used instead of accessing an index that may not exist.

bshaffer:
@frankyn agreed it's super subtle and easy to miss. Hopefully this explains it:

```php
// this results in a JSON array, which is recognized by the API, and so policies are updated as expected
var_dump(json_encode([0 => 'foo', 1 => 'bar'], true));
// OUTPUT: string(13) "["foo","bar"]"

// this results in a JSON object, which is not recognized by the API, and so all policies are removed
var_dump(json_encode([0 => 'foo', 2 => 'bar'], true));
// OUTPUT: string(21) "{"0":"foo","2":"bar"}"
```

I'll add a comment in the sample as well

frankyn:
Oh wow, is that why you'd like to get PolicyBuilder back into these samples? Seems like a easy thing to miss if the library doesn't do it for users.

PR #1087, opened 2020-04-21
updates for slim v4

bshaffer:
Supersedes #1075 and #1076

cc @jdpedrie


bshaffer:
Consider using these deps instead of `slim/psr` and `slim/http`:

```
        "guzzlehttp/psr7": "^1.6",
        "http-interop/http-factory-guzzle": "^1.0"
```

PR #1096, opened 2020-04-28
Fix vision_classification_predict.php

santoshbaggam:
At line 48, $content has to be passed, instead of $image which is undefined.


santoshbaggam:
@googlebot I signed it!

bshaffer:
Thank you for finding and fixing this @santoshbaggam !!

PR #1099, opened 2020-04-30
feat: refactor postgres sample to use slim app

jdpedrie:


cloud_sql/postgres/pdo/README.md
```diff
@@ -39,15 +66,16 @@ $ php -S localhost:8080
 
 Navigate towards http://localhost:8080 to verify your application is running correctly.
 
-## Google App Engine Standard
+## Google App Engine Flex
```
grayside, line 69: I'm a bit confused here, shouldn't we cover both Standard and Flex?

cloud_sql/postgres/pdo/tests/VotesTest.php
```diff
@@ -0,0 +1,150 @@
+<?php
```
grayside, line 1: Oh yeah, got some tests!

Is there a plan to add integration/end-to-end tests in the future? Mocking the database for a database usage sample only gets you so far.

cloud_sql/postgres/pdo/README.md
jdpedrie: I'm not sure. Perhaps I extrapolated from sql server (where standard isn't an option because you can't install the extension) to the other products where there's no compatibility issue?

@shubha-rajan @kurtisvg

cloud_sql/postgres/pdo/tests/VotesTest.php
jdpedrie: @bshaffer what's the standard course here?

cloud_sql/postgres/pdo/README.md
shubha-rajan: The compatibility issue is only with SQL server, because the drivers don't support Unix sockets and GAE standard doesn't support TCP connections. If you can connect through unix sockets, then Standard should work.

cloud_sql/postgres/pdo/tests/VotesTest.php
kurtisvg: The Cloud SQL DPEs have an OKR around e2e testing for every sample next quarter.

cloud_sql/postgres/pdo/README.md
kurtisvg: We should include instructions for both.

cloud_sql/postgres/pdo/README.md
```diff
@@ -2,35 +2,62 @@
...
+
+### Set Configuration Values
+
+Set the required environment variables for your connection to Cloud SQL. If you are using TCP mode as described above, do not set the `CLOUD_SQL_CONNECTION_NAME` variable.
+
+```bash
+export DB_USER='my-db-user'
+export DB_PASS='my-db-pass'
+export DB_NAME='my-db-name'
+export DB_HOSTNAME='localhost'
```
kurtisvg, line 56:
```suggestion
export DB_HOST='127.0.0.1'
```

cloud_sql/postgres/pdo/README.md
```diff
@@ -39,15 +66,16 @@ $ php -S localhost:8080
...
 
 First, update `app.yaml` with the correct values to pass the environment variables into the runtime.
 
+Then, make sure that the service account `service-{PROJECT_NUMBER}>@gae-api-prod.google.com.iam.gserviceaccount.com` has the IAM role `Cloud SQL Client`.
+
 Next, the following command will deploy the application to your Google Cloud project:
 
 ```bash
-$ gcloud app deploy
+$ gcloud beta app deploy
```
kurtisvg, line 80: Does this really need to be beta?

jdpedrie: I got that from [here](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/pull/1745/files#diff-7c3739336b12f6b0f8a40a9e3f87dbfaR76) and given that the app.yaml has a `beta_settings` section in it, assumed that `beta` was required in the command.

shubha-rajan:
@jdpedrie it looks like there's a merge conflict in the README. do you mind fixing that and updating the PR?

cloud_sql/postgres/pdo/src/app.php
```diff
@@ -0,0 +1,116 @@
...
+                $cloud_sql_connection_name
+            );
+        }
+
+        // Connect to the database.
+        // Here we set the connection timeout to five seconds and ask PDO to
+        // throw an exception if any errors occur.
+        $conn = new PDO($dsn, $username, $password, [
+            PDO::ATTR_TIMEOUT => 5,
+            PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
```
shubha-rajan, line 69: Question: Would it make sense to move lines 65 and 66 to after 67, and wrap that section in the `cloud_sql_postgres_pdo_timeout` region tag? For reference, the snippet would be included here: https://cloud.google.com/sql/docs/postgres/manage-connections#timeout
Is this functionality similar to the samples for other languages in this section?

```suggestion
        $conn = new PDO($dsn, $username, $password, [
            // [START cloud_sql_postgres_pdo_timeout]
            // Here we set the connection timeout to five seconds and ask PDO to
            // throw an exception if any errors occur.
            PDO::ATTR_TIMEOUT => 5,
            PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
            // [END cloud_sql_postgres_pdo_timeout]
```

jdpedrie: That block is already a part of the `cloud_sql_postgres_pdo_create` region. Is it possible to have nested regions?

Elsewhere I think we wanted to have the timeout and error mode as part of the other snippet, but those settings can also be set separately and duplicating them won't cause any trouble. We could do something like this:

```php
        // # [START cloud_sql_postgres_pdo_create]
        // // $username = 'your_db_user';
        // // $password = 'yoursupersecretpassword';
        // // $dbName = 'your_db_name';
        // // $cloud_sql_connection_name = getenv("CLOUD_SQL_CONNECTION_NAME");
        // // $hostname = "127.0.0.1"; // Only used in TCP mode.

        if ($hostname) {
            // Connect using TCP
            $dsn = sprintf('pgsql:dbname=%s;host=%s', $dbName, $hostname);
        } else {
            // Connect using UNIX sockets
            $dsn = sprintf(
                'pgsql:dbname=%s;host=/cloudsql/%s',
                $dbName,
                $cloud_sql_connection_name
            );
        }

        // Connect to the database.
        // Here we set the connection timeout to five seconds and ask PDO to
        // throw an exception if any errors occur.
        $conn = new PDO($dsn, $username, $password, [
            PDO::ATTR_TIMEOUT => 5,
            PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
        ]);
        # [END cloud_sql_postgres_pdo_create]

        # [START cloud_sql_postgres_pdo_timeout]
        $conn->setAttribute(PDO::ATTR_TIMEOUT, 5);
        $conn->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
        # [END cloud_sql_postgres_pdo_timeout]
```

It's a bit odd in the application to set those values twice, but for the purpose of the docs site it may be alright.

shubha-rajan: It is possible to nest region tags, so we won't need to duplicate them. In other samples, we've been using a combination of nesting and using `EXCLUDE` tags, but I think I actually like the approach of setting them outside of the connection creation statement better (as opposed to setting them twice). It looks cleaner. 
```
// # [START cloud_sql_postgres_pdo_create]
// // $username = 'your_db_user';
// // $password = 'yoursupersecretpassword';
// // $dbName = 'your_db_name';
// // $cloud_sql_connection_name = getenv("CLOUD_SQL_CONNECTION_NAME");
// // $hostname = "127.0.0.1"; // Only used in TCP mode.

if ($hostname) {
    // Connect using TCP
    $dsn = sprintf('pgsql:dbname=%s;host=%s', $dbName, $hostname);
} else {
    // Connect using UNIX sockets
    $dsn = sprintf(
        'pgsql:dbname=%s;host=/cloudsql/%s',
        $dbName,
        $cloud_sql_connection_name
    );
}

// Connect to the database.
$conn = new PDO($dsn, $username, $password);
# [END cloud_sql_postgres_pdo_create]

# [START cloud_sql_postgres_pdo_timeout]
// Here we set the connection timeout to five seconds and ask PDO to
// throw an exception if any errors occur.
$conn->setAttribute(PDO::ATTR_TIMEOUT, 5);
$conn->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
# [END cloud_sql_postgres_pdo_timeout]
```

shubha-rajan: @kurtisvg thoughts?

kurtisvg: Does `new PDO(...)` initialize the connection on object init, or is it waiting until it's used? If the first, I think we have to include it as part of the initial call (otherwise it won't actually timeout).

If so, I think something like this:

```php
// [START cloud_sql_postgres_pdo_create]
// ...

// Connect to the database.
# [START cloud_sql_postgres_pdo_timeout]
// Here we set the connection timeout to five seconds and ask PDO to
// throw an exception if any errors occur.
$conn = new PDO($dsn, $username, $password, [
    PDO::ATTR_TIMEOUT => 5,
    PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
]);
# [END cloud_sql_postgres_pdo_timeout]
# [END cloud_sql_postgres_pdo_create]
```

jdpedrie: it initializes the connection when the object is created.

cloud_sql/postgres/pdo/README.md
```diff
@@ -39,14 +66,28 @@ $ php -S localhost:8080
...
+
+To run on App Engine Standard, create an App Engine project by following the setup for these [instructions](https://cloud.google.com/appengine/docs/standard/php7/quickstart#before-you-begin).
+
+First, update `app-standard.yaml` with the correct values to pass the environment variables into the runtime.
+
 Next, the following command will deploy the application to your Google Cloud project:
 
 ```bash
-$ gcloud app deploy
+$ gcloud app deploy app-standard.yaml
```
grayside, line 92: In the samples I've more typically seen `app.standard.yaml` or `app.flex[ible].yaml`

cloud_sql/postgres/pdo/README.md
grayside: Looks like this is still a beta setting for flex, but not standard: https://cloud.google.com/sql/docs/mysql/connect-app-engine#flexible

cloud_sql/postgres/pdo/src/app.php
shubha-rajan: In that case, let's nest the region tags. We should probably update the mysql and sql server samples too. Feel free to open a separate PR for that and I can approve and merge those as well.

jdpedrie: Hmm. Are we thinking something like this?

```php
// # [START cloud_sql_postgres_pdo_create]
// // $username = 'your_db_user';
// // $password = 'yoursupersecretpassword';
// // $dbName = 'your_db_name';
// // $cloud_sql_connection_name = getenv("CLOUD_SQL_CONNECTION_NAME");
// // $hostname = "127.0.0.1"; // Only used in TCP mode.

if ($hostname) {
    // Connect using TCP
    $dsn = sprintf('pgsql:dbname=%s;host=%s', $dbName, $hostname);
} else {
    // Connect using UNIX sockets
    $dsn = sprintf(
        'pgsql:dbname=%s;host=/cloudsql/%s',
        $dbName,
        $cloud_sql_connection_name
    );
}

// Connect to the database.
// Here we set the connection timeout to five seconds and ask PDO to
// throw an exception if any errors occur.
$conn = new PDO($dsn, $username, $password, [
    PDO::ATTR_TIMEOUT => 5,
    PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
]);

# [START_EXCLUDE]
# [START cloud_sql_postgres_pdo_timeout]
$conn->setAttribute(PDO::ATTR_TIMEOUT, 5);
$conn->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
# [END cloud_sql_postgres_pdo_timeout]
# [END_EXCLUDE]
# [END cloud_sql_postgres_pdo_create]
```

kurtisvg: That won't work because we are still showing our users the wrong way to do it. Was there something wrong with my example above?

The snippet lookup automatically pulls out comments that are region tags, so it'll render without the extra region tags (but the code will still be included.

For example:
```
# [START REGION1]
This shows in region 1. 
# [START REGION2]
This shows in region 1 and 2. 
# [END REGION2]
# [END REGION1]
```

jdpedrie: Ah, sorry I misunderstood at first. This is fixed now.

shubha-rajan:
@GoogleCloudPlatform/php-admins ping for codeowner review

cloud_sql/postgres/pdo/app.yaml
```diff
@@ -12,15 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-runtime: php72
+runtime: php
+env: flex
```
bshaffer, line 16: Shouldn't this be `app.flex.yaml`?  Or if we insist on having a default, the default should be App Engine Standard and not Flex.

cloud_sql/postgres/pdo/composer.json
```diff
@@ -0,0 +1,24 @@
...
+        "php": ">= 7.2",
+        "slim/slim": "^4.5",
+        "slim/twig-view": "^3.1",
+        "pimple/pimple": "^3.3",
+        "guzzlehttp/psr7": "^1.6",
+        "http-interop/http-factory-guzzle": "^1.0"
+    },
+    "require-dev": {
+        "phpunit/phpunit": "^8.5"
+    }
```
bshaffer, line 23: This is not necessary, all of the tests in this repo use the `phpunit/phpunit` dep declared in `testing/composer.json`

cloud_sql/postgres/pdo/phpunit.xml.dist
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit colors="true">
+  <testsuites>
+    <testsuite name="CloudSQLPostgresSample">
```
bshaffer, line 4: this should use the `bootstrap` attribute and have it set to the file in `testing/bootstrap/php`

shubha-rajan:
@jdpedrie just a few more minor changes before we can merge this. PTAL at the review comments

jdpedrie:
@bshaffer @shubha-rajan done

shubha-rajan:
@bshaffer ping for review

cloud_sql/postgres/pdo/composer.json
```diff
@@ -0,0 +1,21 @@
...
+            "Google\\Cloud\\Samples\\CloudSQL\\Postgres\\Tests\\": "src"
+        }
+    },
+    "require": {
+        "php": ">= 7.2",
+        "slim/slim": "^4.5",
+        "slim/twig-view": "^3.1",
+        "pimple/pimple": "^3.3",
+        "guzzlehttp/psr7": "^1.6",
+        "http-interop/http-factory-guzzle": "^1.0"
```
bshaffer, line 19: Do we really need all five of these dependencies for such a simple sample?

bshaffer: We could use these dependencies instead of the two guzzle ones:

```
    "slim/http": "^1.0",
    "slim/psr7": "^1.0"
```

Is there a good reason to choose one over the other? We should be consistent with all slim samples in this repo (see #1087)

shubha-rajan:
@bshaffer good to merge?

SaketramDurbha:
I've created PR #1130 that depends on the changes made in this pull request. @bshaffer, if you could review the new changes to this PR at your convenience, I would greatly appreciate it.

cloud_sql/postgres/pdo/README.md
```diff
@@ -39,14 +66,28 @@ $ php -S localhost:8080
...
+To run on App Engine Flex, create an App Engine project by following the setup for these [instructions](https://cloud.google.com/appengine/docs/standard/php7/quickstart#before-you-begin).
+
+First, update `app.flex.yaml` with the correct values to pass the environment variables into the runtime.
+
+Then, make sure that the service account `service-{PROJECT_NUMBER}>@gae-api-prod.google.com.iam.gserviceaccount.com` has the IAM role `Cloud SQL Client`.
+
+Next, the following command will deploy the application to your Google Cloud project:
+
+```bash
+$ gcloud beta app deploy
```
bshaffer, line 80: This should be `gcloud beta app deploy app.flex.yaml` since the name was changed

cloud_sql/postgres/pdo/README.md
```diff
@@ -39,14 +66,28 @@ $ php -S localhost:8080
...
+```
+
 ## Google App Engine Standard
 
-To run on GAE-Standard, create an App Engine project by following the setup for these [instructions](https://cloud.google.com/appengine/docs/standard/php7/quickstart#before-you-begin).
+To run on App Engine Standard, create an App Engine project by following the setup for these [instructions](https://cloud.google.com/appengine/docs/standard/php7/quickstart#before-you-begin).
 
-First, update `app.yaml` with the correct values to pass the environment variables into the runtime.
+First, update `app-standard.yaml` with the correct values to pass the environment variables into the runtime.
```
bshaffer, line 87: should be `app.standard.yaml`, not `app-standard.yaml`

cloud_sql/postgres/pdo/README.md
```diff
@@ -2,35 +2,62 @@
...
+
+```bash
+$ ./cloud_sql_proxy \
+    --instances=$CLOUD_SQL_CONNECTION_NAME=tcp:5432 \
+    --credential_file=$GOOGLE_APPLICATION_CREDENTIALS
+```
+
+### Set Configuration Values
+
+Set the required environment variables for your connection to Cloud SQL. If you are using TCP mode as described above, do not set the `CLOUD_SQL_CONNECTION_NAME` variable.
```
bshaffer, line 50: I believe this should be  

> If you are using TCP mode as described above, do not set the `DB_HOST` variable.

The code uses TCP if `DB_HOSTNAME` is not set.

cloud_sql/postgres/pdo/src/app.php
```diff
@@ -0,0 +1,118 @@
...
+$container['votes'] = function (Container $container) {
+    return new Votes($container['db']);
+};
+
+// Setup the database connection in the container.
+$container['db'] = function () {
+    $username = getenv('DB_USER');
+    $password = getenv('DB_PASS');
+    $dbName = getenv('DB_NAME');
+    $hostname = getenv('DB_HOSTNAME');
```
bshaffer, line 41: This uses `DB_HOSTNAME`, but the readme uses `DB_HOST`

cloud_sql/postgres/pdo/src/app.php
```diff
@@ -0,0 +1,118 @@
...
+        // // $cloud_sql_connection_name = getenv("CLOUD_SQL_CONNECTION_NAME");
+        // // $hostname = "127.0.0.1"; // Only used in TCP mode.
+
+        if ($hostname) {
+            // Connect using TCP
+            $dsn = sprintf('pgsql:dbname=%s;host=%s', $dbName, $hostname);
+        } else {
+            // Connect using UNIX sockets
+            $dsn = sprintf(
+                'pgsql:dbname=%s;host=/cloudsql/%s',
```
bshaffer, line 58: I had to change this to `/tmp/cloudsql` because my `/cloudsql` directory was not writable. Would `/tmp/cloudsql` be a more reliable default?

cloud_sql/postgres/pdo/README.md
```diff
@@ -2,35 +2,62 @@
...
+### Unix Socket mode
 
 ```bash
-$ ./cloud_sql_proxy -dir=/cloudsql --instances=$CLOUD_SQL_CONNECTION_NAME --credential_file=$GOOGLE_APPLICATION_CREDENTIALS
+$ ./cloud_sql_proxy -dir=/cloudsql \
+    --instances=$CLOUD_SQL_CONNECTION_NAME \
+    --credential_file=$GOOGLE_APPLICATION_CREDENTIALS
 ```
 
 Note: Make sure to run the command under a user with write access in the `/cloudsql` directory. This proxy will use this folder to create a unix socket the application will use to connect to Cloud SQL.
```
bshaffer, line 38: It would be better to say something like "Note: If your user cannot write to the `/cloudsql` directory, change this to a writable directory. You will also need to update the connection string in `src/app.php`"

cloud_sql/postgres/pdo/README.md
```diff
@@ -39,14 +66,28 @@ $ php -S localhost:8080
...
+
+Then, make sure that the service account `service-{PROJECT_NUMBER}>@gae-api-prod.google.com.iam.gserviceaccount.com` has the IAM role `Cloud SQL Client`.
+
+Next, the following command will deploy the application to your Google Cloud project:
+
+```bash
+$ gcloud beta app deploy
+```
+
 ## Google App Engine Standard
```
bshaffer, line 83: Let's move "App Engine Standard" to before "App Engine Flex"

cloud_sql/postgres/pdo/app.flex.yaml
```diff
@@ -0,0 +1,37 @@
...
+  cloud_sql_instances: "<MY-PROJECT>:<INSTANCE-REGION>:<INSTANCE-NAME>"
+
+runtime_config:
+  document_root: .
+
+# Defaults to "serve index.php" and "serve public/index.php". Can be used to
+# serve a custom PHP front controller (e.g. "serve backend/index.php") or to
+# run a long-running PHP script as a worker process (e.g. "php worker.php").
+#
+# entrypoint: serve index.php
```
bshaffer, line 37: I believe this comment is only valid for app engine standard, and not for app engine flex

cloud_sql/postgres/pdo/src/app.php
shubha-rajan: I believe `/cloudsql` is the default because it's the path that's expected by App Engine. In #1130 we added the ability to override the default path by setting an environment variable.

cloud_sql/postgres/pdo/README.md
shubha-rajan: That would make sense here, but in PR #1130 we're making it possible to override this default with an environment variable, and that's going to be the standard across languages for this sample.

cloud_sql/postgres/pdo/app.flex.yaml
shubha-rajan: fixed

cloud_sql/postgres/pdo/README.md
shubha-rajan: fixed

cloud_sql/postgres/pdo/src/app.php
shubha-rajan: fixed (I changed the readme to use `DB_HOSTNAME`)

cloud_sql/postgres/pdo/app.flex.yaml
bshaffer: I tested this and found some very strange behavior:

1. `entrypoint: serve index.php` => `entrypoint1: serve index.php` - throws error "unexpected attribute entrypoint2"
1. `entrypoint: serve index.php` => `entrypoint: serve index1.php` - does not throw error, application works as expected, even though `index1.php` doesn't exist.
1. `entrypoint: serve1 index.php` => `entrypoint: serve index1.php` - does not throw error, application works as expected, even though `serve1` doesn't exist.

My theory is that this attribute has been deprecated and while is "valid" is no longer being used in App Engine Flex.

Adding @tmatsuo for comment

cloud_sql/postgres/pdo/README.md
shubha-rajan: It should actually be 

> If you are using Unix socket mode as described above, do not set the `DB_HOSTNAME` variable. 

The code should use TCP only if DB_HOSTNAME is set:

```
if ($hostname) {
            // Connect using TCP
            $dsn = sprintf('pgsql:dbname=%s;host=%s', $dbName, $hostname);
        } else {
            // Connect using UNIX sockets
            $dsn = sprintf(
                'pgsql:dbname=%s;host=/cloudsql/%s',
```

Updated this in the last set of commits

cloud_sql/postgres/pdo/README.md
shubha-rajan: fixed

cloud_sql/postgres/pdo/README.md
shubha-rajan: fixed

cloud_sql/postgres/pdo/app.flex.yaml
bshaffer: Just realized I've been using `gcloud app deploy` instead of `gcloud beta app deploy`. I'll test everything again with the beta command.

cloud_sql/postgres/pdo/tests/VotesTest.php
bshaffer: please don't forget this one!!

cloud_sql/postgres/pdo/README.md
bshaffer: FYI I tested this using `gcloud app deploy` (no beta flag) for flex and it worked as expected with no discernible differences.

cloud_sql/postgres/pdo/app.flex.yaml
bshaffer: OK, done. The beta command had no effect.

cloud_sql/postgres/pdo/README.md
bshaffer: I removed `beta_settings` and the app failed. But it seems that using `beta` does not make a difference.

cloud_sql/postgres/pdo/app.flex.yaml
kurtisvg: Added a comment to handle that in #1130

PR #1101, opened 2020-05-06
Create CODEOWNERS

czahedi:
crwilcox@ requested teams for repo access across the board for SoDa teams. Adding a CODEOWNERS file with php-admins as default but with the storage team.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,12 @@
+# Code owners file.
+# This file controls who is tagged for review for any given pull request.
+#
+# For syntax help see:
+# https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
+
+
+# The php-admins team is the default owner for anything not
+# explicitly taken by someone else.
+*                               @GoogleCloudPlatform/php-admins
+
+/storage/                          @GoogleCloudPlatform/storage-dpe @GoogleCloudPlatform/php-admins
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -7,6 +7,10 @@
 
 # The php-admins team is the default owner for anything not
 # explicitly taken by someone else.
-*                               @GoogleCloudPlatform/php-admins
+*                         @GoogleCloudPlatform/php-admins
 
-/storage/                          @GoogleCloudPlatform/storage-dpe @GoogleCloudPlatform/php-admins
+/bigtable/                @GoogleCloudPlatform/bigtable-dpe @GoogleCloudPlatform/php-admins
+/cloud_sql/               @GoogleCloudPlatform/cloud-sql-dpes @GoogleCloudPlatform/php-admins
+/datastore/               @GoogleCloudPlatform/firestore-dpe @GoogleCloudPlatform/php-admins
+/firestore/               @GoogleCloudPlatform/firestore-dpe @GoogleCloudPlatform/php-admins
+/storage/                 @GoogleCloudPlatform/storage-dpe @GoogleCloudPlatform/php-admins


CODEOWNERS
```diff
@@ -0,0 +1,12 @@
...
+#
+# For syntax help see:
+# https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
+
+
+# The php-admins team is the default owner for anything not
+# explicitly taken by someone else.
+*                               @GoogleCloudPlatform/php-admins
+
+/storage/                          @GoogleCloudPlatform/storage-dpe @GoogleCloudPlatform/php-admins
```
bshaffer, line 12: Seems like we should add all the storage products here, such as `cloud_sql`, `datastore`, `firestore`, and `bigquery`

CODEOWNERS
```diff
@@ -0,0 +1,12 @@
+# Code owners file.
+# This file controls who is tagged for review for any given pull request.
+#
+# For syntax help see:
+# https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
+
+
+# The php-admins team is the default owner for anything not
+# explicitly taken by someone else.
+*                               @GoogleCloudPlatform/php-admins
```
bshaffer, line 10: can we get the indentation fixed here? Looks like it should be consistent with the other lines.

CODEOWNERS
```diff
@@ -0,0 +1,12 @@
...
+
```
bshaffer, line 11: Is there supposed to be an extra newline between `*` and `/storage`?

CODEOWNERS
czahedi: Yep noted (though not bigquery since that isn't a storage product). Adding those.

CODEOWNERS
czahedi: I took that convention from the python doc: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/.github/CODEOWNERS

I'm cool with whatever you think is best.

CODEOWNERS
czahedi: Fixed.

CODEOWNERS
bshaffer: Ok let's keep it!

CODEOWNERS
```diff
@@ -0,0 +1,16 @@
...
+# The php-admins team is the default owner for anything not
+# explicitly taken by someone else.
+*                         @GoogleCloudPlatform/php-admins
+
+/bigtable/                @GoogleCloudPlatform/bigtable-dpe @GoogleCloudPlatform/php-admins
+/cloud_sql/               @GoogleCloudPlatform/cloud-sql-dpes @GoogleCloudPlatform/php-admins
+/datastore/               @GoogleCloudPlatform/firestore-dpe @GoogleCloudPlatform/php-admins
+/firestore/               @GoogleCloudPlatform/firestore-dpe @GoogleCloudPlatform/php-admins
+/storage/                 @GoogleCloudPlatform/storage-dpe @GoogleCloudPlatform/php-admins
```
bshaffer, line 16: Hmm, we should also add `spanner`, `bigtable`, `bigquerydatatransfer`... that's probably all of em.

crwilcox: This is specifically storage/database focused to start.

bshaffer: wouldn't that include spanner and bigtable then?

czahedi: So we have bigtable, and don't need bigquery. I can add Spanner too but need to know who should be on the team. Will sync w/ WIlcox tomorrow.

crwilcox: Spanner is currently being worked on by a separate libraries team. It is very much the plan to add all products eventually, but we wanted to test it out with a smaller group to start.

PR #1102, opened 2020-05-15
PHP 7.3 runtime is no longer in beta

darrencarlton:
Removing the "beta" tag from PHP 7.3. Also, I'm updating the App Engine doc to call the runtime the "PHP 7" runtime, since it supports 7.2 and 7.3, and will soon launch a beta for 7.4. So I've made a slight edit here to match the language in the App Engine doc.

### DIFF
--- a/appengine/php72/helloworld/README.md
+++ b/appengine/php72/helloworld/README.md
@@ -1,6 +1,6 @@
-# Hello World on App Engine Standard for PHP 7
+# Hello World on App Engine Standard for PHP 7.2 and 7.3
 
 This sample demonstrates how to deploy a *very* basic application to the
-PHP 7 standard environment in Google App Engine.
+PHP 7.2 and 7.3 standard environment in Google App Engine.
 
 ## View the [full tutorial](https://cloud.google.com/appengine/docs/standard/php7/quickstart)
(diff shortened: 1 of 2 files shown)


appengine/php72/helloworld/README.md
```diff
@@ -1,6 +1,6 @@
-# Hello World on App Engine Standard for PHP 7.2
```
bshaffer, line 1: PHP 7 is too vague. This should be `PHP 7.2 and 7.3`

appengine/php72/helloworld/README.md
```diff
@@ -1,6 +1,6 @@
...
+# Hello World on App Engine Standard for PHP 7
 
-This sample demonstrates how to deploy a *very* basic application to Google
-App Engine for PHP 7.2.
+This sample demonstrates how to deploy a *very* basic application to the
+PHP 7 standard environment in Google App Engine.
```
bshaffer, line 4: This should be `PHP 7.2 and 7.3 standard environment`

darrencarlton: Hi Brent:

In the App Engine doc, we're calling it the "PHP 7" runtime, which supports PHP 7.2 and 7.3. Otherwise, each time we support a new language version, we have to change titles and so on.

Do you want to match the doc, or be more specific here and call out each language version?

bshaffer: We should be more specific, as nobody in the PHP community refers to it as "PHP 7". Minor versions are actually major versions in PHP releases, so lumping them together while not supporting all of them is confusing.

PR #1108, opened 2020-06-01
Improve Spanner delete_data sample coverage

larkee:
The Deleting rows in a table section gives four different delete methods. The PHP section pointed at the documentation for $database->delete(). This only demonstrated one of the methods.

This PR adds the missing sample and includes the missing delete methods.

### DIFF
--- a/spanner/README.md
+++ b/spanner/README.md
@@ -103,7 +103,7 @@ To run the Spanner Samples:
       create-storing-index                 Adds an storing index to the example database.
       create-table-timestamp               Creates a table with a commit timestamp column.
       create-table-with-datatypes          Creates a table with supported datatypes.
-      delete-data                          Deletes sample data from the given database.d
+      delete-data                          Deletes sample data from the given database.
       delete-data-with-dml                 Remove sample data from the given database with a DML statement.
       deleted-data-with-partitioned-dml    Deletes sample data in the database by partition with a DML statement.
       help                                 Displays help for a command
--- a/spanner/README.md
+++ b/spanner/README.md
@@ -103,6 +103,7 @@ To run the Spanner Samples:
       create-storing-index                 Adds an storing index to the example database.
       create-table-timestamp               Creates a table with a commit timestamp column.
       create-table-with-datatypes          Creates a table with supported datatypes.
+      delete-data                          Deletes sample data from the given database.d
       delete-data-with-dml                 Remove sample data from the given database with a DML statement.
       deleted-data-with-partitioned-dml    Deletes sample data in the database by partition with a DML statement.
       help                                 Displays help for a command
--- a/spanner/src/delete_data.php
+++ b/spanner/src/delete_data.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


spanner/src/delete_data.php
```diff
@@ -0,0 +1,71 @@
...
+    $albumsToDelete = $spanner->keySet([
+        'keys' => [[2, 1], [2, 3]]
+    ]);
+    $database->delete('Albums', $albumsToDelete);
+
+    // Delete a range of rows where the column key is >=3 and <5
+    $singersRange = $spanner->keyRange([
+        'startType' => KeyRange::TYPE_CLOSED,
+        'start' => [3],
+        'endType' => KeyRange::TYPE_OPEN,
```
bshaffer, line 54: when these are omitted, does the client use a default or will it throw an error? And can we add to the comment with that information?

spanner/test/spannerTest.php
```diff
@@ -126,6 +126,15 @@ public function testReadData()
         $this->assertContains('SingerId: 2, AlbumId: 3, AlbumTitle: Terrified', $output);
     }
 
+    /**
+     * @depends testInsertData
+     */
+    public function testDeleteData()
+    {
+        $output = $this->runCommand('delete-data');
+        $this->assertContains('Deleted data.' . PHP_EOL, $output);
```
bshaffer, line 135: We should add a confirmation via the client to confirm the deletion worked.

spanner/src/delete_data.php
larkee: KeyRange needs a start and end. The types default to KeyRange::TYPE_OPEN. I have added comments with those details.

spanner/test/spannerTest.php
larkee: Should this be part of the sample or the test? Including it in the test seems like it would go against the style but it seems strange to query the tables in the sample.

spanner/src/delete_data.php
bshaffer: This comment seems conflicting. The `KeyRange MUST include a start and end` but defaults to OPEN. I think the first line can be removed.

spanner/src/delete_data.php
```diff
@@ -0,0 +1,75 @@
...
+    $database->delete('Singers', $singersToDelete);
+
+    // Delete remaining Singers rows, which will also delete the remaining
+    // Albums rows because Albums was defined with ON DELETE CASCADE
+    $remainingSingers = $spanner->keySet([
+        'all' => true
+    ]);
+    $database->delete('Singers', $remainingSingers);
+
+
```
bshaffer, line 71: we do not need three newlines here. One is sufficient.

spanner/src/delete_data.php
larkee: So you need to include `start` and `end` keys but `startType` and `endType` default to OPEN. Do you think removing the first note conveys this?

spanner/src/delete_data.php
larkee: Done.

spanner/src/delete_data.php
bshaffer: I understand now. I think we can leave it. Thank you for clarifying!

spanner/test/spannerTest.php
bshaffer: It should be part of the test. We do this in other tests, so it does not go against the style.

larkee: Done.

larkee:
@bshaffer PTAL :)

spanner/README.md
```diff
@@ -103,6 +103,7 @@ To run the Spanner Samples:
       create-storing-index                 Adds an storing index to the example database.
       create-table-timestamp               Creates a table with a commit timestamp column.
       create-table-with-datatypes          Creates a table with supported datatypes.
+      delete-data                          Deletes sample data from the given database.d
```
bshaffer, line 106: Found a typo! There's a trailing "d"...

larkee: Good spotting! Fixed :)

PR #1109, opened 2020-06-02
Updates IAP samples for new IAP service URL

bshaffer:
And makes the command actually useful, as before it was dependent on the IAP URL respobse body having a specific format.


iap/iap.php
```diff
@@ -51,27 +51,19 @@
 
 // Create a validate Command.
 $application->add((new Command('validate'))
-    ->addArgument('url', InputArgument::REQUIRED, 'The Identity-Aware Proxy-protected URL to fetch.')
-    ->addArgument('clientId', InputArgument::REQUIRED, 'The client ID used by Identity-Aware Proxy.')
-    ->addArgument('serviceAccountPath', InputArgument::REQUIRED, 'Path for the service account you want to use.')
+    ->addArgument('jwt', InputArgument::REQUIRED, 'A JWT from the X-Goog-Iap-Jwt-Assertion heaer')
```
dwsupplee, line 54:
```suggestion
    ->addArgument('jwt', InputArgument::REQUIRED, 'A JWT from the X-Goog-Iap-Jwt-Assertion header')
```

PR #1121, opened 2020-06-24
Typo in snippet, incremet -> increment

jlara310:
Should be increment here, reference for method: https://googleapis.github.io/google-cloud-php/#/docs/cloud-firestore/v1.14.0/firestore/fieldvalue?method=increment


jlara310:
@BenWhitehead

PR #1122, opened 2020-06-26
Migrate DLP samples to use regional syntax in parent field

sushicw:
Update DLP samples to include `locations/global` in the parent field, so that the job name returned by the API matches what is used internally and what is used in the pubsub message being looked for.

See internal bug b/158100365

Also added a troubleshooting note to the DLP readme.

### DIFF
--- a/dlp/test/data/harmful.csv
+++ b/dlp/test/data/harmful.csv
@@ -1,0 +1,2 @@
+Name,Description
+John Doe,Name of a person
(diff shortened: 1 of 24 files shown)


dlp/test/data/harmful.csv
```diff
@@ -0,0 +1,2 @@
+Name,Description
```
bshaffer, line 1: Why is this added? This is in GCS

sushicw: It's in the project used by kokoro, but wasn't in my project so I had to add it before I could properly run tests locally. I can remove it from the PR, it's not needed. (Although it's separately a good idea to either add all the instructions you need to replicate the test env in your own project, or to have the tests themselves push any data into the project that's needed, other langs do it both ways).

bshaffer: gotcha. I don't mind keeping it, just wanted to make sure it was intentional.

PR #1128, opened 2020-07-01
storage: bucket lifecycle management

jkwlui:


storage/test/BucketLifecycleManagementTest.php
```diff
@@ -0,0 +1,91 @@
...
+    use ExecuteCommandTrait;
+
+    private static $commandFile = __DIR__ . '/../storage.php';
+    protected $storage;
+    protected $bucket;
+
+    public function setUp()
+    {
+        // Sleep to avoid the rate limit for creating/deleting.
+        sleep(5 + rand(2, 4));
```
bshaffer, line 40: This is strange, to sleep before we've even attempted to create the bucket. It would be better to use the [resource exhausted backoff](https://github.com/GoogleCloudPlatform/php-tools/blob/master/src/TestUtils/ExponentialBackoffTrait.php#L33):

```php
class BucketLifecycleManagementTest extends TestCase
{
        //...
    public function setUp()
    {
        $storage = new StorageClient();
        $bucketName = sprintf('php-olm-%s-%s', time(), rand(1000, 9999));
        $this->useResourceExhaustedBackoff();
        self::$backoff->execute(function() use ($storage, $bucketName) {
            $this->bucket = $storage->createBucket($bucketName);
        });
    }
```

storage/test/BucketLifecycleManagementTest.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public function testEnableBucketLifecycleManagement()
+    {
+        $bucketName = $this->bucket->name();
+        $output = $this->runCommand('bucket-lifecycle-management', [
+            'bucket' => $bucketName,
+            '--enable' => true,
+        ]);
+        $match = 'Lifecycle management is enabled for bucket ' . $bucketName . ' and the rules are:';
```
bshaffer, line 60: nit, but either of these are more idiomatic:
```php
$match = "Lifecycle management is enabled for bucket $bucketName and the rules are:";
$match = sprintf('Lifecycle management is enabled for bucket %s and the rules are:', $bucketName);
```

storage/test/BucketLifecycleManagementTest.php
```diff
@@ -0,0 +1,91 @@
...
+        $this->assertContains(
+            [
+                'action' => [
+                    'type' => 'Delete'
+                ],
+                'condition' => [
+                    'age' => 100
+                ]
+            ],
+            $rules);
```
bshaffer, line 74: formatting here should be:
```php
        $this->assertContains([
            'action' => [
                'type' => 'Delete'
            ],
            'condition' => [
                'age' => 100
            ]
        ], $rules);
```

storage/test/BucketLifecycleManagementTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+    /** @depends testEnableBucketLifecycleManagement */
+    public function testDisableBucketLifecycleManagement()
+    {
+        $bucketName = $this->bucket->name();
+        $output = $this->runCommand('bucket-lifecycle-management', [
+            'bucket' => $bucketName,
+            '--disable' => true,
+        ]);
+        $expectedOutput = 'Lifecycle management is disabled for bucket ' . $bucketName . '.' . PHP_EOL;
```
bshaffer, line 85: nit, but either of these are more idiomatic:
```php
$expectedOutput = "Lifecycle management is disabled for bucket $bucketName.\n";
$expectedOutput = sprintf('Lifecycle management is disabled for bucket %s.' . PHP_EOL, $bucketName);
```

storage/test/BucketLifecycleManagementTest.php
```diff
@@ -0,0 +1,91 @@
...
+
+    private static $commandFile = __DIR__ . '/../storage.php';
+    protected $storage;
+    protected $bucket;
+
+    public function setUp()
+    {
+        // Sleep to avoid the rate limit for creating/deleting.
+        sleep(5 + rand(2, 4));
+        $this->storage = new StorageClient();
```
bshaffer, line 41: this can just be `$storage`, as `$this->storage` is never used again.

storage/test/BucketLifecycleManagementTest.php
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Unit Tests for enable/disable bucket lifecycle management.
+ */
+class BucketLifecycleManagementTest extends TestCase
+{
+    use TestTrait;
+    use ExecuteCommandTrait;
+
+    private static $commandFile = __DIR__ . '/../storage.php';
+    protected $storage;
```
bshaffer, line 34: this local variable can be removed

storage/test/BucketLifecycleManagementTest.php
```diff
@@ -0,0 +1,91 @@
...
+    protected $bucket;
+
+    public function setUp()
+    {
+        // Sleep to avoid the rate limit for creating/deleting.
+        sleep(5 + rand(2, 4));
+        $this->storage = new StorageClient();
+
+        // Append random because tests for multiple PHP versions were running at the same time.
+        $bucketName = 'php-olm-' . time() . '-' . rand(1000, 9999);
```
bshaffer, line 44:
```php
$bucketName = sprintf('php-olm-%s-%s', time(), rand(1000, 9999));
```

storage/test/BucketLifecycleManagementTest.php
jkwlui: I have to be honest - I copied the setup code from UniformBucketLevelAccessCommandTest.php without much thought. This looks much better~

storage/src/disable_bucket_lifecycle_management.php
```diff
@@ -0,0 +1,46 @@
...
+function disable_bucket_lifecycle_management($bucketName)
+{
+    $storage = new StorageClient();
+    $bucket = $storage->bucket($bucketName);
+
+    $bucket->update([
+        'lifecycle' => null
+    ]);
+
+    printf('Lifecycle management is disabled for bucket ' . $bucketName . '.' . PHP_EOL);
```
bshaffer, line 44:
```php
printf('Lifecycle management is disabled for bucket %s.' . PHP_EOL, $bucketName);
```

storage/src/enable_bucket_lifecycle_management.php
```diff
@@ -0,0 +1,56 @@
...
+
+    $bucket->update([
+        'lifecycle' => $lifecycle
+    ]);
+
+    $lifecycle = $bucket->currentLifecycle();
+
+    printf('Lifecycle management is enabled for bucket ' . $bucketName . ' and the rules are:' . PHP_EOL);
+    foreach ($lifecycle as $rule) {
+        printf("%s" . PHP_EOL, print_r($rule));
```
bshaffer, line 53: this isn't going to work as expected, as `print_r` prints immediately unless you pass `true` as a second argument. What you want is:

```php
    foreach ($lifecycle as $rule) {
        print_r($rule);
        print(PHP_EOL);
    }
```

storage/src/enable_bucket_lifecycle_management.php
```diff
@@ -0,0 +1,56 @@
...
+            'age' => 100
+        ]);
+
+    $bucket->update([
+        'lifecycle' => $lifecycle
+    ]);
+
+    $lifecycle = $bucket->currentLifecycle();
+
+    printf('Lifecycle management is enabled for bucket ' . $bucketName . ' and the rules are:' . PHP_EOL);
```
bshaffer, line 51:
```php
printf('Lifecycle management is enabled for bucket %s and the rules are:' . PHP_EOL, $bucketName);
```

jkwlui:
Thanks @bshaffer!

PR #1160, opened 2020-08-24
fix(cloud-sql): fix region tag

shubha-rajan:
Fixing a commented out region tag.

### DIFF
--- a/cloud_sql/sqlserver/pdo/src/app.php
+++ b/cloud_sql/sqlserver/pdo/src/app.php
@@ -42,10 +42,10 @@ $container['db'] = function () {
 
     try {
         # [START cloud_sql_sqlserver_pdo_create_tcp]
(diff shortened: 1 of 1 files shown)


cloud_sql/sqlserver/pdo/src/app.php
```diff
@@ -41,7 +41,7 @@
     $host = getenv('DB_HOST');
 
     try {
-        // # [START cloud_sql_sqlserver_pdo_create_tcp]
+        # [START cloud_sql_sqlserver_pdo_create_tcp]
         // // $username = 'your_db_user';
```
bshaffer, line 45: While we're at it, can we get rid of the double comments on all these variables?

shubha-rajan: done.

PR #1166, opened 2020-09-03
feat: add custom time to object metadata

frankyn:
Adding a new metadata field to print out.

### DIFF
--- a/storage/src/object_metadata.php
+++ b/storage/src/object_metadata.php
@@ -77,19 +77,19 @@ function object_metadata($bucketName, $objectName)
         printf('Content-type: %s' . PHP_EOL, $info['contentType']);
     }
     if (isset($info['temporaryHold'])) {
-        printf("Temporary hold: " . ($info['temporaryHold'] ? "enabled" : "disabled") . PHP_EOL);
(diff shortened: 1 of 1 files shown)


storage/src/object_metadata.php
```diff
@@ -85,6 +85,9 @@ function object_metadata($bucketName, $objectName)
     if (isset($info['retentionExpirationTime'])) {
         printf("retentionExpirationTime: " . $info['retentionExpirationTime'] . PHP_EOL);
     }
+    if (isset($info['customTime'])) {
+        printf("customTime: " . $info['customTime'] . PHP_EOL);
```
tmatsuo, line 89: It's using `printf` so maybe:

```PHP
        printf("customTime: %s" . PHP_EOL, $info['customTime']);
```
?

tmatsuo (approved):
I think the added code is consistent with the rest, but I feel we should use the place holder if we use `printf`.

storage/src/object_metadata.php
frankyn: Good point, I'll send an update.

tmatsuo (approved):

frankyn:
Thanks @tmatsuo!

PR #1183, opened 2020-09-23
feat(spanner): add standalone samples for NUMERIC type

larkee:


spanner/spanner.php
```diff
@@ -303,6 +315,18 @@
     })
 );
 
+// Update data with timestamp column command
```
thiagotnunes, line 318: I think this comment should say "Update data with numeric column command"

larkee: Good spotting! Fixed :+1:

thiagotnunes (approved):

spanner/src/update_data_with_numeric_column.php
```diff
@@ -0,0 +1,59 @@
...
+ * @param string $instanceId The Spanner instance ID.
+ * @param string $databaseId The Spanner database ID.
+ */
+function update_data_with_numeric_column($instanceId, $databaseId)
+{
+    $spanner = new SpannerClient();
+    $instance = $spanner->instance($instanceId);
+    $database = $instance->database($databaseId);
+
+    $operation = $database->transaction(['singleUse' => true])
```
bshaffer, line 49: According to the documentatioon, this returns a `Google\Cloud\Core\Timestamp` object and not an `$operation`. Also this parameter is never used. So I'd recommend removing it.

larkee: Done. As a note, this template exists in other samples. I'll make a separate PR to correct this in those samples.

bshaffer (approved):

larkee:
This PR cannot be merged until GoogleCloudPlatform/php-docs-samples/pull/1184 is merged to skip the failing test.

skuruppu:
@bshaffer, if you can help us merge this in, that would be awesome. The tests now pass :)

PR #1184, opened 2020-09-28
fix(spanner): skip failing backup test

larkee:
The ListBackupOperations test is failing across all the languages due to a bug in production. This PR marks that test as skipped to unblock other PRs. Once the bug is fixed, this skip can be removed.

Towards #1186

### DIFF
--- a/spanner/test/spannerBackupTest.php
+++ b/spanner/test/spannerBackupTest.php
@@ -98,6 +98,10 @@ class spannerBackupTest extends TestCase
      */
     public function testListBackupOperations()
     {
+        $this->markTestSkipped(
+            "This test is failing due to a bug in production. Skipping until it is fixed."
+        );
+
         $databaseId2 = self::$databaseId . '-2';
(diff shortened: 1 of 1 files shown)


thiagotnunes (approved):
LGTM

bshaffer:
I don't want to skip these tests without anything in place to ensure they get reinstated

Please file a bug at the very least to remind us to unskip them

larkee:
@bshaffer I've created an issue for the test failure and assigned it to myself. I will keep it open until the test is unskipped and passing.

skuruppu:
@bshaffer it could take weeks to do a backend rollout with a fix so I think we have no choice but to skip this for now. We're doing the same across all the language repos. @larkee will keep the issue open until we revert the skip. Hope that's ok with you.

bshaffer (approved):
We are keeping https://github.com/GoogleCloudPlatform/php-docs-samples/issues/1186 open to track this

PR #1189, opened 2020-10-14
chore(appengine): remove unused samples

ace-n:
These samples are not used in cloudsite.

(Besides, removing them makes my DRIFT project easier! 😝 )

 @michaelawyu please confirm deleting the mail samples is OK.


grayside (changes requested):
php-docs-samples hosts some samples for PHP that are manually synchronized to community tutorials, and possibly some that are linked from the docs. This is a broader scope than most samples repo, in part to offset the smaller number of folks working in PHP. Removing these needs to be considered one at a time.

bshaffer (changes requested):
Let's keep the PHP 7.2 samples. For two reasons

 1. The Wordpress samples are used in a community tutorial
 2. CloudSQL/gRPC samples are useful for those exploring samples

I am happy to get rid the `php55` samples!

grayside:
Is the CloudSQL sample superseded by https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/cloud_sql?

bshaffer:
@grayside I believe so... we could update the README in `appengine/php72/cloudsql` to point there!

ace-n:
OK - to clarify: we should delete the same `php55` samples, while leaving `php72` alone?

(If so - I'll close this PR + do that in a separate one.)

bshaffer:
@ace-n yes that's right! But you can remove the php72 Cloud SQL samples if you'd like, as long as you put a note in the `README` in `appengine/php72/cloudsql` that they're now at `cloud_sql`...

ace-n:
Closed in favor of #1190

PR #1194, opened 2020-10-16
feat(functions): add functions_http_unit_test and functions_http_content

grayside:
This PR adds two additional samples.

* functions_http_content is a new function sample that demonstrates handling different content types.
* functions_http_unit_test shows how to write a unit test for a Cloud Function. The existing UnitTest is not a good fit for use as a sample, and rather than refactor it I decided to keep both as a cross-check that the unit test sample for developers and our own (more thorough) testing stay in alignment.

### DIFF
--- a/functions/helloworld_http/test/SampleUnitTest.php
+++ b/functions/helloworld_http/test/SampleUnitTest.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START functions_http_unit_test]
+
+namespace Google\Cloud\Samples\Functions\HelloworldHttp\Test;
+
+use GuzzleHttp\Psr7\ServerRequest;
+use PHPUnit\Framework\TestCase;
+
+/**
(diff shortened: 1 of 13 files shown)


ace-n (approved):
LGTM if @bshaffer approves.

functions/helloworld_http/test/SampleUnitTest.php
```diff
@@ -0,0 +1,47 @@
...
+ * Unit test for helloHttp.
+ */
+class SampleUnitTest extends TestCase
+{
+    public static function setUpBeforeClass() : void
+    {
+        require_once __DIR__ . '/../index.php';
+    }
+
+    public function testFunction() : void
```
bshaffer, line 37: nit: this should be `public function testFunction(): void`

functions/http_content_type/composer.json
```diff
@@ -0,0 +1,10 @@
+{
+    "require": {
+        "google/cloud-functions-framework": "^0.6"
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Google\\Cloud\\Samples\\Functions\\HttpContentType\\Test\\": "test"
+        }
+    }
```
bshaffer, line 9: Same nit as before, I'd prefer to just require the class in the test instead of autoloading them`

bshaffer (commented):
Looks good! Just a few nits

functions/helloworld_http/test/SampleUnitTest.php
grayside: Thanks! I don't think we have that in the linter.

functions/http_content_type/composer.json
grayside: Ah, I didn't see this nit previously, will make that change.

functions/helloworld_http/test/SampleUnitTest.php
bshaffer: We don't ... probably because it's new... I'll look into it!

bshaffer: done! 

https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1202

bshaffer (approved):

PR #1195, opened 2020-10-18
Firestore | Renaming all files so they are sorted by write, query, up…

drupalista-br:
Renaming all files so they are sorted by write, query, update and reference.

It makes it easier to browse and find a sample.


bshaffer:
We appreciate your submission! I do agree this is a better way to sort these samples, but as these are consistent across many languages, and referenced in our documentation (and tracked in our dashboards), the amount of effort to rename all these would unfortunately be non-trivial.

PR #1196, opened 2020-10-20
feat(functions): add functions_http_form_data sample

grayside:
This adds the functions_http_form_data sample, which is focused on managing file uploads.

* It uses the PSR-7 file upload implementation provided by Guzzle.
* I skipped unit tests. System tests are still very fast and run in CI.
* This includes structured logging, which is arguably outside sample scope. We can simplify to line logging if it seems too distracting.
* To facilitate experimentation, I added a composer script to start the function. Should this be added to the composer.json by default for all sample functions?


functions/http_form_data/index.php
```diff
@@ -0,0 +1,83 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use GuzzleHttp\Psr7\Response;
+
+function uploadFile(ServerRequestInterface $request): ResponseInterface
+{
+    if ($request->getMethod() != 'POST') {
+        return (new Response())
+            ->withBody(GuzzleHttp\Psr7\stream_for('Method Not Allowed: expected POST, found ' . $request->getMethod()))
+            ->withStatus(405);
```
bshaffer, line 29: This would look better if we use the `Response` constructor:

```php
return new Response(405, 'Method Not Allowed: expected POST, found ' . $request->getMethod());
```

And same with all the responses below

functions/http_form_data/test/DeployTest.php
```diff
@@ -0,0 +1,76 @@
...
+            $resp = $this->client->$method('', [
+                'multipart' => $test['multipart'],
+            ]);
+            $this->assertEquals($test['code'], $resp->getStatusCode(), $test['label'] . ' code:');
+            $actual = trim((string) $resp->getBody());
+            $this->assertContains($test['expected'], $actual, $test['label'] . ':');
+        }
+    }
+
+    public function testErrorCases() : void
```
bshaffer, line 54: nit: should be `public function testErrorCases(): void`

functions/http_form_data/composer.json
```diff
@@ -0,0 +1,17 @@
...
+        "start": [
+            "Composer\\Config::disableProcessTimeout",
+            "TMPDIR=./tmp FUNCTION_TARGET=uploadFile php -S localhost:${PORT:-8080} vendor/bin/router.php"
+        ]
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Google\\Cloud\\Samples\\Functions\\HelloworldHttp\\Test\\": "test"
+        }
+    }
```
bshaffer, line 16: nit, but I'd prefer to just manually include the `TestCasesTrait` file in the test cases than use the autoloader. This is because this file is visible to our users as part of the sample, and having the test logic here may be confusing. It's not a hill I'd die on, but since we're only talking about a single class, I'd prefer to use `require_once`

bshaffer (changes requested):

functions/http_form_data/index.php
grayside: Ah, thanks for this feedback! This approach is what I saw in the guzzle docs, but I think what you suggest is much cleaner.

grayside: Note in practice it will look like this:

```php
return new Response(405, [], 'Method Not Allowed: expected POST, found ' . $request->getMethod());
```

The second argument is an associative array of headers. Slightly unfortunate, that really should have been after the body.

functions/http_form_data/composer.json
grayside: Fixed.

functions/http_form_data/test/DeployTest.php
grayside: Fixed

functions/http_form_data/index.php
grayside: Fixed.

functions/http_form_data/index.php
```diff
@@ -0,0 +1,73 @@
...
+
+    $contentType = $request->getHeader('Content-Type')[0];
+    if (strpos($contentType, 'multipart/form-data') !== 0) {
+        return new Response(400, [], 'Bad Request: content of type "multipart/form-data" not provided, found ' . $contentType);
+    }
+
+    $fileList = [];
+    /** @var $file Psr\Http\Message\UploadedFileInterface */
+    foreach ($request->getUploadedFiles() as $name => $file) {
+        // Use caution when trusting the client-provided filename:
```
ace-n, line 38: Nit: this should probably be a callout in the doc itself, not (just?) the sample.

ace-n (approved):

bshaffer (approved):

functions/http_form_data/test/SystemTest.php
```diff
@@ -0,0 +1,70 @@
...
+            $resp = $this->client->$method('/', [
+                'multipart' => $test['multipart'],
+            ]);
+            $this->assertEquals($test['code'], $resp->getStatusCode(), $test['label'] . ' code:');
+            $actual = trim((string) $resp->getBody());
+            $this->assertContains($test['expected'], $actual, $test['label'] . ':');
+        }
+    }
+
+    public function testErrorCases() : void
```
bshaffer, line 51: just a few more of these to fix (3) and we can merge!

bshaffer (approved):

functions/http_form_data/index.php
grayside: Acknowledged, will send a CL to add this to the docs also.

PR #1197, opened 2020-10-20
feat(functions): Add HTTP samples, round 1

ace-n:


functions/concepts_filesystem/index.php
```diff
@@ -0,0 +1,36 @@
...
+    $directory = dirname(__FILE__);
+    $contents = scandir($directory);
+
+    echo "Files:" . PHP_EOL;
+
+    foreach ($contents as $file) {
+        echo "\t" . $file . PHP_EOL;
+    }
+
+    return '';
```
grayside, line 33: Per discussion in #1198, we're not directly outputting, don't use echo. Assemble and return a string or a Response object.

functions/concepts_filesystem/index.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+// [START functions_concepts_filesystem]
+
+use Psr\Http\Message\ServerRequestInterface;
+
+function listFiles(ServerRequestInterface $request): string
+{
+    $directory = dirname(__FILE__);
+    $contents = scandir($directory);
```
grayside, line 25: nit: circuitous code.

`__DIR__` is equivalent to `dirname(__FILE__)`. Further, `scandir('.')` is equivalent to `scandir(__DIR__)`. Since this is sometimes ambiguous, I suggest we directly use __DIR__.


```suggestion
    $contents = scandir(__DIR__);
```

functions/concepts_filesystem/test/DeployTest.php
```diff
@@ -0,0 +1,56 @@
...
+        ]);
+
+        // Assert status code.
+        $this->assertEquals('200', $resp->getStatusCode());
+
+        // Assert function output.
+        $expected = 'index.php';
+        $actual = trim((string) $resp->getBody());
+        // Failures often lead to a large HTML page in the response body.
+        $this->assertContains($expected, $actual);
```
grayside, line 54: Rather than duplicate the test cases, can we use the TestCasesTrait pattern as a default practice between the test types? Where we add test cases outside the common denominator we can add additional tests.

functions/concepts_filesystem/test/UnitTest.php
```diff
@@ -0,0 +1,47 @@
...
+    public static function setUpBeforeClass() : void
+    {
+        require_once __DIR__ . '/../index.php';
+    }
+
+    public function testFunction() : void
+    {
+        $request = new ServerRequest('GET', '/');
+        $this->runFunction(self::$name, [$request]);
+        $this->expectOutputRegex("/index\.php/");
```
grayside, line 40: Why expectedOutputRegex instead of the easier to read $this->assertContains?

functions/concepts_requests/composer.json
```diff
@@ -0,0 +1,6 @@
+{
+    "require": {
+        "google/cloud-functions-framework": "^0.6",
+        "curl/curl": "^2.3"
```
grayside, line 4: Please use guzzlehttp/guzzle as the HTTP request client.

functions/concepts_requests/index.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+// [START functions_concepts_requests]
+
+use Psr\Http\Message\ServerRequestInterface;
+use GuzzleHttp\Psr7\Response;
+
+function makeRequest(ServerRequestInterface $request): Response
+{
+    $curl = new Curl\Curl();
```
grayside, line 25: Per note on composer.json, please switch to use the Guzzle request client.
https://packagist.org/packages/guzzlehttp/guzzle

functions/http_cors/test/DeployTest.php
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Class DeployTest.
+ *
+ * This test is not run by the CI system.
+ *
+ * To skip deployment of a new function, run with "GOOGLE_SKIP_DEPLOYMENT=true".
+ * To skip deletion of the tested function, run with "GOOGLE_KEEP_DEPLOYMENT=true".
+ */
+class DeployTest extends TestCase
```
grayside, line 33: Expect System and Deploy test to specifically verify OPTIONS behavior. CORS is never sufficiently tested and we should minimally be sure our OPTIONS logic works as expected.

functions/http_method/index.php
```diff
@@ -0,0 +1,53 @@
...
+                'Forbidden!' . PHP_EOL
+            );
+            break;
+        default:
+            $json_payload = json_encode([
+                'error' => 'something blew up!'
+            ]);
+            return new Response(
+                405,
+                [],
```
grayside, line 46: Is something smart enough to inject `Content-Type: application/json`? If not this is needed.

functions/http_method/index.php
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START functions_http_method]
+
+use Psr\Http\Message\ServerRequestInterface;
+use GuzzleHttp\Psr7\Response;
+
+function helloHttp(ServerRequestInterface $request): Response
```
grayside, line 23: There is already a helloHttp function. To avoid conflicts and confusion one should be renamed.

grayside, line 23: @bshaffer do you have thoughts on whether we should typehint as Response or the ResponseInterface?

functions/http_method/test/DeployTest.php
```diff
@@ -0,0 +1,56 @@
...
+ * To skip deployment of a new function, run with "GOOGLE_SKIP_DEPLOYMENT=true".
+ * To skip deletion of the tested function, run with "GOOGLE_KEEP_DEPLOYMENT=true".
+ */
+class DeployTest extends TestCase
+{
+    use CloudFunctionDeploymentTrait;
+
+    private static $name = 'helloHttp';
+
+    public function testFunction() : void
```
grayside, line 39: My code showed the wrong way, remove the space between parens and colon.

```suggestion
    public function testFunction(): void
```

grayside (changes requested):
Did a first low-res pass.

functions/concepts_filesystem/index.php
ace-n: Done.

functions/concepts_filesystem/index.php
ace-n: Done - I was wondering if such a constant existed, thanks!

functions/http_method/index.php
ace-n: I'm not sure; I'll add it in case.

functions/http_method/index.php
ace-n: Done.

functions/http_method/test/DeployTest.php
ace-n: Done.

functions/concepts_filesystem/test/UnitTest.php
ace-n: Done - stored `$response` as a variable and used `assertContains()` with that.

functions/concepts_requests/composer.json
ace-n: Done.

functions/concepts_requests/index.php
ace-n: Done.

functions/http_cors/test/DeployTest.php
ace-n: Will implement this using `TestCasesTrait`.

functions/http_method/index.php
bshaffer: We should recommend `ResponseInterface` as using the guzzle library is optional, but returning a `ResponseInterface` is not.

functions/concepts_filesystem/test/DeployTest.php
ace-n: Done.

grayside:
Hey Ace, a number of comments are resolved but I don't see pushed commits. Holding off on further review for now.

functions/http_method/index.php
ace-n: Can y'all point me to an example of how to do this? Doesn't look like [directly returning an Interface is possible](https://stackoverflow.com/questions/8084169/return-an-interface-in-php).

(IIUC we might be able to wrap our functions themselves in a `Class`, but I'm not sure if that's the cleanest approach.)

ace-n:
> Hey Ace, a number of comments are resolved but I don't see pushed commits. Holding off on further review for now.

Sorry about the confusion - commits should be live now. Thanks!

functions/http_method/index.php
bshaffer: @ace-n the "naive way" is possible for the Type Hint. I am not suggesting you change what the function returns, only the return type hint, e.g.:

```php
function helloHttp(ServerRequestInterface $request): ResponseInterface
```

functions/concepts_requests/index.php
```diff
@@ -0,0 +1,41 @@
...
+        'base_uri' => 'https://example.com',
+    ]);
+
+    $url_response = $client->get('/');
+
+    $function_response = new Response(
+        $url_response->getStatusCode(),
+        [],
+        ''
+    );
```
bshaffer, line 36: How about comments for the parameters that are not obvious?

```php
    $function_response = new Response(
        $url_response->getStatusCode(),
        [], // headers
        ''  // body
    );
```

functions/concepts_requests/test/UnitTest.php
```diff
@@ -0,0 +1,57 @@
...
+    private static $name = 'makeRequest';
+
+    public static function setUpBeforeClass(): void
+    {
+        require_once __DIR__ . '/../index.php';
+    }
+
+    public function testFunction(): void
+    {
+        foreach (self::cases() as $test) {
```
bshaffer, line 43: We could refactor these to use PHPUnit data providers and they'd be a bit cleaner:

```php
    /** 
      * @dataProvider cases
      */
    public function testFunction($url, $statusCode): void
    {
        $request = new ServerRequest('GET', $url]);
        $output = $this->runFunction(self::$name, [$request]);
        $this->assertEquals(
            $statusCode,
            $output->getStatusCode()
        );
    }
```

functions/http_method/index.php
```diff
@@ -0,0 +1,53 @@
...
+                [],
+                'Forbidden!' . PHP_EOL
+            );
+            break;
+        default:
+            $json_payload = json_encode([
+                'error' => 'something blew up!'
+            ]);
+            return new Response(
+                405,
```
bshaffer, line 45: "Something Blew Up" isn't the actual status code, it's "Method Not Allowed... I think we should at least add a comment after `405` so it's clear why we're using this status code.

functions/http_method/index.php
```diff
@@ -0,0 +1,53 @@
...
+                200,
+                [],
+                'Hello, World!' . PHP_EOL
+            );
+            break;
+        case 'PUT':
+            return new Response(
+                403,
+                [],
+                'Forbidden!' . PHP_EOL
```
bshaffer, line 37: it's not clear why PUT is 403 "Forbidden", but the other methods are "Method Not Allowed". Could we add a comment explaining the reason for the difference?

bshaffer (commented):
These look great!! I just have a few requests:

 - A few spots could use comments for clarification
 - Instead of looping over `cases` in these tests, let's use the `@dataProvider` functionality in PHPUnit
 - Let's make all these functions use `Psr\Http\Message\RequestInterface` return types

functions/http_method/index.php
bshaffer: See https://github.com/GoogleCloudPlatform/functions-framework-php/blob/master/src/CloudEventFunctionWrapper.php#L43

functions/concepts_requests/test/UnitTest.php
grayside: @bshaffer want to file a follow-up to do this to the existing tests? Ace is copying a pattern I laid out.

functions/env_vars/test/DeployTest.php
```diff
@@ -0,0 +1,65 @@
...
+        // Assert function output.
+        $expected = 'bar';
+        $actual = trim((string) $resp->getBody());
+        // Failures often lead to a large HTML page in the response body.
+        $this->assertEquals($expected, $actual);
+    }
+
+    protected static function deployFlags(array $flags = []): array
+    {
+        $flags['--update-env-vars'] = 'FOO=bar;';
```
grayside, line 62: Why the semicolon inside the quote mark? This has the potential to break the deploy command.

functions/helloworld_get/test/SystemTest.php
```diff
@@ -21,26 +21,31 @@
...
 {
     use CloudFunctionLocalTestTrait;
+    use TestCasesTrait;
 
     private static $name = 'helloGet';
 
     public function testFunction(): void
     {
-        // Send a request to the function.
-        $resp = $this->client->get('/');
```
grayside, line 36: Thanks for updating helloworld_get to this pattern.

functions/http_method/index.php
ace-n: Done - figured this out by looking at @grayside's examples.

functions/concepts_requests/index.php
ace-n: Done.

functions/http_method/index.php
ace-n: Added explanations for each HTTP error code.

functions/http_method/index.php
ace-n: Added use-case examples.

functions/env_vars/test/DeployTest.php
ace-n: Removed it.

ace-n:
Tests are failing because the `FOO` environment variable is missing. (By default, it's set to `bar`.)

functions/concepts_requests/index.php
```diff
@@ -0,0 +1,42 @@
...
+use GuzzleHttp\Client;
+use GuzzleHttp\Psr7\Response;
+
+function makeRequest(ServerRequestInterface $request): ResponseInterface
+{
+    $client = new Client([
+        'base_uri' => 'https://example.com',
+    ]);
+
+    $url_response = $client->get('/');
```
grayside, line 31: nit: inline comment linking to documentation on the guzzle http client? Maybe https://docs.guzzlephp.org/en/stable/.

functions/concepts_requests/test/DeployTest.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    use CloudFunctionDeploymentTrait;
+    use TestCasesTrait;
+
+    private static $name = 'makeRequest';
+
+    /**
+      * @dataProvider cases
+      */
+    public function testFunction($url, $statusCode): void
```
grayside, line 45: I'm against URL being abstracted to the separate TestCases, because:

1. It's not something which should ever vary
2. If you set the value to anything except empty string or the function name, the request gets 302 redirected to a Google login page, obscuring errors. This is a divergence in behavior between DeployTest and SystemTest.

functions/http_method/index.php
```diff
@@ -0,0 +1,57 @@
...
+            return new Response(
+                403, // Permission denied
+                [],
+                'Forbidden!' . PHP_EOL
+            );
+            break;
+        default:
+            // Example: request type not supported by the application
+            $json_payload = json_encode([
+                'error' => 'something blew up!'
```
grayside, line 46: nit: If the developer sent an invalid method, it's not that the server blew up. It's that the client did something wrong.

grayside (changes requested):
Getting really close. I've made a point around url being in TestCasesTrait, everything else is minor.

functions/concepts_requests/test/DeployTest.php
ace-n: Done - I moved `url` back into specific test cases.

functions/http_method/index.php
ace-n: This is the error message we use [in other languages](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/functions/http/index.js#L75), so I'd rather keep it the same for consistency.

functions/concepts_requests/index.php
ace-n: Added comments to this sample.

bshaffer (approved):

functions/http_method/index.php
grayside: We probably should, but it's not worth 5 PRs just for this change.

grayside (approved):

PR #1198, opened 2020-10-21
feat(functions): add functions_log_helloworld

ace-n:


functions/log_helloworld/index.php
```diff
@@ -0,0 +1,43 @@
...
+    // These will be sent back as part
+    // of the function's HTTP response
+    echo("HTTP message from echo().");
+    print("HTTP message from print().");
+
+    // Note: this is different than STDOUT
+    // See this page for more info:
+    // https://www.php.net/manual/en/wrappers.php.php
+    $output = fopen('php://output', 'wb');
+    fwrite($output, "HTTP message from fwrite().\n");
```
bshaffer, line 33: This should result in a "headers already sent" warning because output should be either returned as a string or in the Response object. Outputing the response in this way, if it's possible, is not something we want to highlight.

functions/log_helloworld/test/DeployTest.php
```diff
@@ -0,0 +1,57 @@
...
+
+        // Assert status code.
+        $this->assertEquals('200', $resp->getStatusCode());
+
+        // Assert function output.
+        $response = trim((string) $resp->getBody());
+
+        $this->assertContains('print()', $response);
+        $this->assertContains('echo()', $response);
+        $this->assertContains('fwrite()', $response);
```
bshaffer, line 55: I suggest making this check for the full message, since two different lines output `fwrite()`. I'd also suggest including an `assertNotContains` for the `stderr` log message.

functions/log_helloworld/test/SystemTest.php
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Class SystemTest.
+ */
+class SystemTest extends TestCase
+{
+    use CloudFunctionLocalTestTrait;
+
+    private static $name = 'helloWorld';
+
+    public function testFunction() : void
```
bshaffer, line 33: nit, should be `testFunction(): void`

bshaffer (changes requested):

functions/log_helloworld/index.php
```diff
@@ -0,0 +1,43 @@
...
+
+// [START functions_log_helloworld]
+
+use Psr\Http\Message\ServerRequestInterface;
+
+function helloWorld(ServerRequestInterface $request): string
+{
+    // These will be sent back as part
+    // of the function's HTTP response
+    echo("HTTP message from echo().");
```
grayside, line 26: It's not typical to use echo with parentheses. Per Brent's comment though, we should consider removing anyway.

functions/log_helloworld/index.php
```diff
@@ -0,0 +1,43 @@
...
+    // Note: this is different than STDOUT
+    // See this page for more info:
+    // https://www.php.net/manual/en/wrappers.php.php
+    $output = fopen('php://output', 'wb');
+    fwrite($output, "HTTP message from fwrite().\n");
+
+    // Code running in Google Cloud Functions itself
+    // writes log entries to Stackdriver Logging
+    $error = fopen('php://stderr', 'wb');
+    fwrite($error, "Log entry from fwrite().\n");
```
grayside, line 38: This is not specifically errors in any way, STDOUT and STDERR are treated identically by GCF.

I suggest we use a more general variable name, preferably concise:

```suggestion
    $log = fopen('php://stderr', 'wb');
    fwrite($log, "Log entry from fwrite().\n");
```

functions/log_helloworld/index.php
```diff
@@ -0,0 +1,43 @@
...
+
+    return '';
```
grayside, line 40: Ideally this sample might also make a statement about `error_log()`, does it work? Does it need special config?

functions/log_helloworld/index.php
```diff
@@ -0,0 +1,43 @@
...
+}
```
grayside, line 41: While it does push at the scope limits, I think it would be worth considering showcasing how to log with severity:

```
fwrite($log, json_encode([
  'message' => 'Structured log with error severity',
  'severity' => 'error'
]);
```

functions/log_helloworld/index.php
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START functions_log_helloworld]
+
+use Psr\Http\Message\ServerRequestInterface;
+
+function helloWorld(ServerRequestInterface $request): string
```
grayside, line 22: Should this not be `helloLogging` or similar? This seems like it has potential to lead to conflicts if there's another helloWorld function under test.

functions/log_helloworld/test/UnitTest.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public function testFunction() : void
+    {
+        $request = new ServerRequest('GET', '/');
+        $output = $this->runFunction(self::$name, [$request]);
+
+        $this->expectOutputRegex("/echo()/");
+        $this->expectOutputRegex("/print()/");
+        $this->expectOutputRegex("/fwrite()/");
```
grayside, line 43: Why not assertContains on these?

functions/log_helloworld/index.php
grayside: In which case, you are suggesting all the code in this function above this line be removed?

Also, (FF behavior question) if we don't want developers to write to php://output in their function, should we buffer and discard that output? Or even redirect to stderr to simplify logging?

grayside (changes requested):

functions/log_helloworld/index.php
ace-n: The main thing I'm trying to showcase here is "How can I write `error`-level (or at least `warning`-level) log messages to Stackdriver?"

functions/log_helloworld/test/UnitTest.php
bshaffer: To use assertContains, we'd have to wrap the above include file in output buffers. This is preferable, as the above code (it's not obvious) only checks the final `expectOutputRegex`. Something like this would work:

```php

ob_start();
$funcReturn = self::runFunction(self::$name, [$request]);
$output = ob_get_clean();

$this->assertRegexp("/echo()/", output);
```

functions/log_helloworld/index.php
grayside: The code you currently have only applies default severity level. I include code on how to add severity in one of my comments below, as well as in the form data PR.

functions/log_helloworld/test/UnitTest.php
grayside: This assumes not switching to return a string value, which I thought was preferred?

functions/log_helloworld/test/DeployTest.php
ace-n: Done.

functions/log_helloworld/test/UnitTest.php
ace-n: I was aiming for brevity last time. Changed to `assertContains`.

functions/log_helloworld/index.php
ace-n: That's the function name Node uses. 🤷 

I'll change it to `helloLogging` for PHP though.

functions/log_helloworld/index.php
ace-n: Will test this on Cloud Functions itself.

functions/log_helloworld/index.php
ace-n: It's a no-op. (I'll mention that though.)

functions/log_helloworld/index.php
ace-n: Didn't work - it logged the JSON payload as an `INFO` message.

functions/log_helloworld/index.php
ace-n: Done.

functions/log_helloworld/index.php
ace-n: Removed the parenthesis for now.

@bshaffer LMK if we should remove the `echo` outright.

functions/log_helloworld/test/SystemTest.php
ace-n: Done.

functions/log_helloworld/index.php
grayside: Sorry, I forget to point the need for a newline termination, such as PHP_EOL.

Here's my experiment:

```
use Psr\Http\Message\ServerRequestInterface;

function severity(ServerRequestInterface $request): string
{
    $stream = fopen('php://stderr', 'wb');
    $severities = [
        'DEBUG',
        'NOTICE',
        'INFO',
        'WARNING',
        'ERROR',
        'CRITICAL',
        'ALERT',
        'EMERGENCY',
    ];

    foreach ($severities as $severity) {
        $msg = [
            'message' => 'Log with severity ' . $severity,
            'severity' => $severity,
        ];
        fwrite($stream, json_encode($msg) . PHP_EOL);
    }
}
```

Then, checking stackdriver:
Screen Shot 2020-10-21 at 11 35 34 PM

functions/log_helloworld/index.php
bshaffer: Yes, I think it's best to remove it!

functions/log_helloworld/index.php
bshaffer: beautiful

functions/log_helloworld/test/UnitTest.php
```diff
@@ -36,11 +36,14 @@ public static function setUpBeforeClass() : void
     public function testFunction() : void
     {
         $request = new ServerRequest('GET', '/');
-        $output = $this->runFunction(self::$name, [$request]);
 
-        $this->expectOutputRegex("/echo()/");
-        $this->expectOutputRegex("/print()/");
-        $this->expectOutputRegex("/fwrite()/");
+        $response = $this->runFunction(self::$name, [$request]);
+        $output = $this->getActualOutput();
```
bshaffer, line 41: oh, that's better eh? 👍

functions/log_helloworld/index.php
```diff
@@ -0,0 +1,47 @@
...
+    fwrite($output, "HTTP message from fwrite().\n");
+
+    // Code running in Google Cloud Functions itself
+    // writes log entries to Stackdriver Logging
+    // (Log severity is INFO, even for stderr logs.)
+    $log = fopen('php://stderr', 'wb');
+    fwrite($log, "Log entry from fwrite().\n");
+
+    // This doesn't log anything
+    error_log('error_log does not log in Cloud Functions!');
```
grayside, line 42: If error_log doesn't log in cloud functions, this sounds like the environment configuration should be improved to do so. Could you file a bug or shall I?

functions/log_helloworld/index.php
ace-n: Done.

functions/log_helloworld/index.php
ace-n: @grayside Thanks for investigating - done.

functions/log_helloworld/test/UnitTest.php
ace-n: (Yeah, we had to get the output _somehow_.)

functions/log_helloworld/index.php
ace-n: Done - bug filed. 🙂

functions/log_helloworld/index.php
ace-n: Looks like the newlines were the problem. Added.

ace-n:
@grayside N.B: I chose not to use `TestCaseTrait` here, since this sample is fairly simple. LMK if I should change that!

functions/log_helloworld/composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": {
+        "google/cloud-functions-framework": "^0.6"
+    }
```
grayside, line 4: Per chat discussion, let's add a start script. The plan doc has an example.

functions/log_helloworld/index.php
grayside: Per Brent's initial comment, we **do not* want to highlight that developers can create a function response by writing to stdout. In that case, the print and php://output statements should be removed.

functions/log_helloworld/index.php
```diff
@@ -0,0 +1,55 @@
...
+    // of the function's HTTP response
+    print("HTTP message from print().");
+
+    // Note: this is different than STDOUT
+    // See this page for more info:
+    // https://www.php.net/manual/en/wrappers.php.php
+    $output = fopen('php://output', 'wb');
+    fwrite($output, "HTTP message from fwrite().\n");
+
+    // Code running in Google Cloud Functions itself
```
grayside, line 34: nit: Not sure what the line length policy is in this repo, usually in PHP at least 80 characters.

functions/log_helloworld/index.php
```diff
@@ -0,0 +1,55 @@
...
+    //   https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity
+    fwrite($log, json_encode([
+      'message' => 'Structured log with error severity',
+      'severity' => 'error'
+    ]) . PHP_EOL);
+
+    // This doesn't log anything
+    error_log('error_log does not log in Cloud Functions!');
+
+    return '';
```
grayside, line 52: We should be able to remove or change the typehint to `void` and skip this return statement.

functions/log_helloworld/test/UnitTest.php
```diff
@@ -0,0 +1,61 @@
...
+            if (isset($test['contains'])) {
+                $this->assertContains($test['contains'], $output);
+            }
+            if (isset($test['not_contains'])) {
+                $this->assertNotContains($test['not_contains'], $output);
+            }
+        }
+    }
+
+    private static function runFunction($functionName, array $params = []): string
```
grayside, line 57: Note if we change the function to return void, this will also need to be changed, if only to `return ''` here instead.

grayside (changes requested):

functions/log_helloworld/index.php
ace-n: This is a URL, not sure if I can make it that much shorter (other than a `goo.gl` or similar link).

functions/log_helloworld/index.php
ace-n: Done.

functions/log_helloworld/test/UnitTest.php
ace-n: Done. (I had this function return `void` as well.)

functions/log_helloworld/index.php
grayside: Not talking about a URL being too long, I'm talking about the comment lines in general being too short.

functions/log_helloworld/index.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+// [START functions_log_helloworld]
+
+use Psr\Http\Message\ServerRequestInterface;
+
+function helloLogging(ServerRequestInterface $request): string
+{
+    // Code running in Google Cloud Functions itself
+    // writes log entries to Stackdriver Logging
```
grayside, line 25: "Stackdriver" is no longer used. Switch to "Cloud Logging".

ace-n: Done.

functions/log_helloworld/index.php
ace-n: Ah - I usually use a smaller text editor window, and erroneously figured this couldn't hurt. (Fixed.)

grayside (approved):

bshaffer (approved):

PR #1199, opened 2020-10-21
feat(functions): add scopes sample

ace-n:
~**Do not merge** until we determine how to handle global variables.~


functions/tips_scopes/index.php
```diff
@@ -0,0 +1,50 @@
...
+    return 1 + 2 + 3 + 4 + 5;
+}
+
+// [START functions_tips_scopes]
+
+use Psr\Http\Message\ServerRequestInterface;
+
+// Global (instance-wide) scope
+// This computation runs at instance cold-start
+$instanceVar = _heavyComputation();
```
ace-n, line 34: This line is **not** executed when running the tests locally, but it **is** executed by `DeployTest` (i.e. on Cloud Functions itself).

@grayside @bshaffer I presume this is a bug with the local test setup?

grayside: I can't see how local setup affects this offhand. However, this may be related to a flaw in the approach of the sample. Each request causes the PHP code to be reloaded, so there's no optimization in using global scope in this way.

In order to preserve a heavy computation for reuse, I suggest writing the data to the filesystem with `file_put_contents()` and then `file_get_contents()` when needed. No need for global variables, just a defined constant on where to look for this cache. Would also be useful to add a comment that the filesystem is in-memory, so expected to be fast.

grayside (changes requested):

functions/tips_scopes/index.php
ace-n: @bshaffer is that an appropriately idiomatic solution?

(We can also say "PHP does not support caching variables globally".)

functions/tips_scopes/index.php
```diff
@@ -0,0 +1,59 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+
+function scopeDemo(ServerRequestInterface $request): string
+{
+    // Heavy computations should be cached between invocations.
+    // The PHP runtime does NOT preserve variables between invocations, so we
+    // must write their values to a file.
+    // (All writable directories in Cloud Functions are in-memory, so these
+    // operations are typically fast.)
+    $cachePath = '/tmp/cached_value.txt';
```
grayside, line 39: In support of local development, we should not assume `/tmp` is the temp filesystem.

I suggest one of two paths:

* Use PHP's built-in functions to identify the temp directory and create a file.

https://github.com/GoogleCloudPlatform/php-docs-samples/blob/02da6bac19126b68136bdfa52e28549ded891084/functions/http_form_data/index.php#L43

* Create a temp directory inside the sample directory to facilitate developers examining the cache when they run the function locally.

grayside (changes requested):

functions/tips_scopes/index.php
ace-n: Used a built-in function here.

functions/tips_scopes/index.php
```diff
@@ -0,0 +1,59 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+
+function scopeDemo(ServerRequestInterface $request): string
+{
+    // Heavy computations should be cached between invocations.
+    // The PHP runtime does NOT preserve variables between invocations, so we
+    // must write their values to a file.
+    // (All writable directories in Cloud Functions are in-memory, so these
+    // operations are typically fast.)
+    $cachePath = sys_get_temp_dir() . 'cached_value.txt';
```
bshaffer, line 39: You need a directory separator, so `/cached_value.txt`

bshaffer (commented):
I do think it'd be better to use a caching library (PSR-6) for these samples

functions/tips_scopes/test/SystemTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Google\Cloud\Samples\Functions\HelloworldGet\Test;
+
+use PHPUnit\Framework\TestCase;
+use Google\Cloud\TestUtils\CloudFunctionLocalTestTrait;
+
+/**
+ * Class SystemTest.
+ */
+class SystemTest extends TestCase
```
bshaffer, line 27: nit: can we call this `LocalTest.php`? That's the naming I use elsewhere in this repo

bshaffer (approved):

functions/tips_scopes/test/SystemTest.php
ace-n: Done.

functions/tips_scopes/test/DeployTest.php
```diff
@@ -0,0 +1,55 @@
...
+ * To skip deployment of a new function, run with "GOOGLE_SKIP_DEPLOYMENT=true".
+ * To skip deletion of the tested function, run with "GOOGLE_KEEP_DEPLOYMENT=true".
+ */
+class DeployTest extends TestCase
+{
+    use CloudFunctionDeploymentTrait;
+
+    private static $name = 'scopeDemo';
+
+    public function testFunction() : void
```
grayside, line 39: nit
```suggestion
    public function testFunction(): void
```

functions/tips_scopes/test/LocalTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Google\Cloud\Samples\Functions\HelloworldGet\Test;
+
+use PHPUnit\Framework\TestCase;
+use Google\Cloud\TestUtils\CloudFunctionLocalTestTrait;
+
+/**
+ * Class SystemTest.
+ */
+class SystemTest extends TestCase
```
grayside, line 27: Align class name to file name.

Per discussion in #1213, we're talking about renaming this file & class to IntegrationTest, but the current name is SystemTest.

functions/tips_scopes/test/LocalTest.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    use CloudFunctionLocalTestTrait;
+
+    private static $name = 'scopeDemo';
```
grayside, line 31: Requires rename.

```suggestion
    private static $entryPoint = 'scopeDemo';
```

functions/tips_scopes/test/DeployTest.php
```diff
@@ -0,0 +1,55 @@
...
+ * This test is not run by the CI system.
+ *
+ * To skip deployment of a new function, run with "GOOGLE_SKIP_DEPLOYMENT=true".
+ * To skip deletion of the tested function, run with "GOOGLE_KEEP_DEPLOYMENT=true".
+ */
+class DeployTest extends TestCase
+{
+    use CloudFunctionDeploymentTrait;
+
+    private static $name = 'scopeDemo';
```
grayside, line 37: Requires rename.

```suggestion
    private static $entryPoint = 'scopeDemo';
```

functions/tips_scopes/test/DeployTest.php
```diff
@@ -0,0 +1,55 @@
...
+            // Uncomment and CURLOPT_VERBOSE debug content will be sent to stdout.
+            // 'debug' => true
+        ]);
+
+        // Assert status code.
+        $this->assertEquals('200', $resp->getStatusCode());
+
+        // Assert function output.
+        $output = trim((string) $resp->getBody());
+        $this->assertContains('Per instance: 120', $output);
```
grayside, line 52: This is a fairly shallow test here. Would running the request twice to confirm persistence of in-memory data be worthwhile? The risk of spinning up a second instance could be mitigated by running with --max-instances=1

functions/tips_scopes/test/DeployTest.php
```diff
@@ -0,0 +1,55 @@
...
+class DeployTest extends TestCase
+{
+    use CloudFunctionDeploymentTrait;
+
+    private static $name = 'scopeDemo';
+
+    public function testFunction() : void
+    {
+        // Send a request to the function.
+        $resp = $this->client->get('', [
```
grayside, line 42: This should be a POST to ensure no edge cases replay the GET request and throw off the per-instance result.

grayside (changes requested):

functions/tips_scopes/test/LocalTest.php
ace-n: Done.

functions/tips_scopes/test/DeployTest.php
ace-n: How should we do that? Maybe a timestamp-based and/or random value?

functions/tips_scopes/test/DeployTest.php
ace-n: Changed all `GET`s to `POST`s for this sample.

functions/tips_scopes/test/DeployTest.php
ace-n: Done.

functions/tips_scopes/test/DeployTest.php
ace-n: Done.

functions/tips_scopes/test/LocalTest.php
ace-n: Done.

functions/tips_scopes/test/DeployTest.php
grayside: Ah, I had it in my head that heavyComputation was adding to the previous value. Given that, the way we might test this is if there's a log entry when priming the cache and otherwise not. Note I'm more interested in this for the Integration Test than the DeployTest.

functions/tips_scopes/composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": {
+        "google/cloud-functions-framework": "^0.6"
```
bshaffer, line 3: This should be updated to `^0.7.1` now!

functions/tips_scopes/test/UnitTest.php
```diff
@@ -0,0 +1,48 @@
...
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Unit tests for the Cloud Function.
+ */
+class UnitTest extends TestCase
+{
+    private static $entryPoint = 'scopeDemo';
+
+    public static function setUpBeforeClass() : void
```
bshaffer, line 31: nit: should be `public static function setUpBeforeClass(): void`

functions/tips_scopes/test/UnitTest.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        require_once __DIR__ . '/../index.php';
+    }
+
+    public function testFunction() : void
```
bshaffer, line 36: nit: should be `public function testFunction(): void`

functions/tips_scopes/test/UnitTest.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $request = new ServerRequest('POST', '/');
+        $output = $this->runFunction(self::$entryPoint, [$request]);
+        $this->assertContains('Per instance: 120', $output);
+        $this->assertContains('Per function: 15', $output);
+    }
+
+    private static function runFunction($functionName, array $params = []) : string
```
bshaffer, line 44: `private static function runFunction($functionName, array $params = []): string`

bshaffer (commented):
a bunch of nits but otherwise this still looks great!

functions/tips_scopes/composer.json
ace-n: Done.

functions/tips_scopes/test/UnitTest.php
ace-n: Done.

functions/tips_scopes/test/UnitTest.php
ace-n: Done.

functions/tips_scopes/test/UnitTest.php
ace-n: Done.

functions/tips_scopes/index.php
ace-n: Done.

functions/tips_scopes/test/DeployTest.php
ace-n: Got it - I went ahead and added a `Response` test to both `IntegrationTest` and `DeployTest`.

grayside (approved):

Issue #1204, opened 2020-10-22
Run Hello World Sample Instructions Slightly Off


grant:
Running the Hello World Sample for Cloud Run is slightly incorrect:

https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/run

```
export SAMPLE='helloworld'
cd php-docs-samples/run/$SAMPLE
docker build --tag $SAMPLE .
cd: no such file or directory: php-docs-samples/run/helloworld
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat ~/Documents/github/googlecloudplatform/php-docs-samples/run/Dockerfile: no such file or directory
```

PR #1205, opened 2020-10-22
docs(run): fix run helloworld instructions

grant:
Fixes cd path for sample to work properly.

Fixes https://github.com/GoogleCloudPlatform/php-docs-samples/issues/1204

### DIFF
--- a/run/README.md
+++ b/run/README.md
@@ -20,6 +20,7 @@ For more Cloud Run samples beyond PHP, see the main list in the [Cloud Run Sampl
 
     ```sh
     git clone https://github.com/GoogleCloudPlatform/php-docs-samples.git
+    cd php-docs-samples/run
     ```
 
 ## How to run a sample locally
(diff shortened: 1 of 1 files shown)


grayside (commented):
Per request from @bshaffer this command assumes a starting working directory of the directory from where you would have run git clone. I have no opinion either way.

grant:
Thanks for the background.

IMO we should please state that assumption then. Or state the `cd` right after the `git clone` step.

If a user is linked to this README, the instructions will not work. Usually READMEs are self-isolated and expect you run commands from the directory you're reading the README in.

bshaffer:
@grant If we include instructions on cloning the repo, we include instructions on changing directory into the repo. This is consistent with all of our tutorials.

If you'd like, split up `cd php-docs-samples/run` into the previous step where the repo is cloned. This should be clear to everyone.

grant:
> If you'd like, split up `cd php-docs-samples/run` into the previous step where the repo is cloned. This should be clear to everyone.

Yeah, that's what I'd suggest.

bshaffer:
This looks good but you'll need to merge master for it to pass the tests

bshaffer (approved):

grant:
Sure. Rebased.

PR #1208, opened 2020-10-26
Add new updated SQL Server sample region tags.

jsimonweb:


bshaffer (commented):
Looks good but do we need region tags for both `cloud_sql_server_pdo_connection` and `cloud_sql_sqlserver_pdo_connection`?

jsimonweb:
After I've imported the code into docs using `cloud_sql_sqlserver_pdo_connection`, I'll create another PR to remove the old `cloud_sql_server_pdo_connection`

bshaffer (approved):

PR #1212, opened 2020-10-30
feat(functions): add Slack samples

ace-n:
**Update:** `DeployTest` has been fixed, so this is clear to merge. 😄


functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,177 @@
...
+}
+// [END functions_slack_request]
+
+// [START functions_slack_search]
+/**
+* Receive a Slash Command request from Slack.
+*/
+function receiveRequest(ServerRequestInterface $request): ResponseInterface
+{
+
```
bshaffer, line 141: nit: unnecessary newline

functions/slack_slash_command/composer.json
```diff
@@ -0,0 +1,6 @@
+{
+    "require": {
+        "google/cloud-functions-framework": "^0.6",
+        "google/apiclient": "^2.8"
+    }
```
bshaffer, line 5: We should add the helpful tool which removes all client services except the ones we [explicitly ask for](https://github.com/googleapis/google-api-php-client#cleaning-up-unused-services):

```json
{
    "require": {
        "google/cloud-functions-framework": "^0.6",
        "google/apiclient": "^2.8"
    },
    "scripts": {
        "post-update-cmd": "Google\\Task\\Composer::cleanup"
    },
    "extra": {
        "google/apiclient-services": [
            "Kgsearch"
        ]
    }
}
```

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,177 @@
...
+}
+// [END functions_slack_format]
+
+// [START functions_slack_request]
+/**
+* Send the user's search query to the Knowledge Graph API.
+*/
+function searchKnowledgeGraph($query): Google_Service_Kgsearch_SearchResponse
+{
+
```
bshaffer, line 115: nit: extra newline

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,177 @@
...
+        $timestamp = $timestamp[0];
+        $signature = $signature[0];
+    }
+
+
+    //$nowInMs = round(microtime(true) * 1000);
+
+    // Compute signature
+    $plaintext = 'v0:' . $timestamp . ':' . (string) $request->getBody();
+    //var_dump($plaintext);
```
bshaffer, line 53: remove `var_dump` comment

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,177 @@
...
+    $signature = $request->getHeader("X-Slack-Signature");
+    if (!$timestamp || !$signature) {
+        return false;
+    } else {
+        $timestamp = $timestamp[0];
+        $signature = $signature[0];
+    }
+
+
+    //$nowInMs = round(microtime(true) * 1000);
```
bshaffer, line 49: I don't think this is supposed to be here...

ace-n: Forgot to push the commit that removed this. 😄 Done.

functions/slack_slash_command/index.php
ace-n: Done.

functions/slack_slash_command/index.php
ace-n: Removed.

functions/slack_slash_command/index.php
ace-n: Removed.

functions/slack_slash_command/composer.json
ace-n: Added.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,171 @@
...
+
+// [START functions_slack_request]
+/**
+* Send the user's search query to the Knowledge Graph API.
+*/
+function searchKnowledgeGraph($query): Google_Service_Kgsearch_SearchResponse
+{
+    /**
+    * PHP Functions Framework does not support "global"/instance-scoped
+    * constants, so we fetch these values within PHP functions themselves.
```
grayside, line 113: This comment isn't quite right. PHP supports global constants and global state, it just doesn't preserve that state between requests. If you had a use case of multiple API requests to Knowledge Graph as part of a single function execution, you'd still want to stick the service variable in a global, or as a static variable in the function. (static variables inside a function stick around between function calls, though not between requests).

Actions:
1. Adjust comment content.
2. Consider if we want to teach reuse by default, even if it's not helpful in this use case.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,171 @@
...
+// [START functions_slack_search]
+/**
+* Receive a Slash Command request from Slack.
+*/
+function receiveRequest(ServerRequestInterface $request): ResponseInterface
+{
+    // Validate request
+    if ($request->getMethod() !== 'POST') {
+        // [] = empty headers
+        return new Response(405, [], '');
```
grayside, line 139: I think being explicit with the first parameters was useful for the initial HTTP introductions, but at this point more conciseness makes sense. PHP doesn't require you to specify all the parameters on the caller's side where there's a default in the function signature. Therefore, this could be:

```suggestion
        return new Response(405);
```

However, without any text in the response, this might confuse developers trying out the function to see no response. It would be more helpful to return a response that will show up in non-verbose `curl`, even if only "Method not allowed".

This feedback also applies to the 400 and 403 responses below.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    // Slack sends requests as URL-encoded strings
```
grayside, line 142: Specifically `application/x-www-form-urlencoded` or something else?

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,171 @@
...
+
+    if (!isValidSlackWebhook($request, $bodyStr)) {
+        // [] = empty headers
+        return new Response(403, [], '');
+    }
+
+    $query = $bodyParams["text"];
+
+    // Call knowledge graph API
+    $kgResponse = searchKnowledgeGraph($query);
```
grayside, line 159: What if searchKnowledgeGraph fails or throws an exception? Does that change the message to be sent for formatting?

@bshaffer here's an example of an HTTP function needing support from https://github.com/GoogleCloudPlatform/functions-framework-php/issues/60.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,171 @@
...
+
+    return $hash === $signature;
+}
+// [END functions_verify_webhook]
+
+// [START functions_slack_format]
+/**
+* Format the Knowledge Graph API response into a richly formatted Slack message.
+*/
+function formatSlackMessage($kgResponse, $query): string
```
grayside, line 60: I started us off by typehinting the return values, but we should probably declare strict types and include typehints for the function parameters as well.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,171 @@
...
+{
+    $responseJson = [
+        "response_type" => "in_channel",
+        "text" => "Query: " . $query
+    ];
+
+    $entityList = $kgResponse["itemListElement"];
```
grayside, line 67: Does $kgResponse have a failure mode which is not a thrown exception but for which itemListElement is not populated?

```
PHP Notice:  Undefined index: itemListElement in php shell code on line 1
```

This warning can be suppressed with a check or a default value, such as a null coalescing operator.

```suggestion
    $entityList = $kgResponse["itemListElement"] ?? null;
```

grayside (changes requested):

functions/slack_slash_command/test/TestCasesTrait.php
```diff
@@ -0,0 +1,114 @@
...
+        'X-Slack-Request-Timestamp' => '0',
+        'X-Slack-Signature' => $hash,
+    ];
+}
+
+trait TestCasesTrait
+{
+    public static function cases(): array
+    {
+        $SLACK_SIGNATURE = getenv('SLACK_TEST_SIGNATURE');
```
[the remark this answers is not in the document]
ace-n: Good catch - removed.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,171 @@
...
+    $responseJson["attachments"] = array($attachmentJson);
+
+    return json_encode($responseJson);
+}
+// [END functions_slack_format]
+
+// [START functions_slack_request]
+/**
+* Send the user's search query to the Knowledge Graph API.
+*/
```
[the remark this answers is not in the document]
ace-n: Done.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,171 @@
...
+    $kgResults = $service->entities->search($params);
+
+    return $kgResults;
+}
+// [END functions_slack_request]
+
+// [START functions_slack_search]
+/**
+* Receive a Slash Command request from Slack.
+*/
```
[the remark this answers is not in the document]
ace-n: Done.

functions/slack_slash_command/index.php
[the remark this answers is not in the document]
ace-n: Done.

functions/slack_slash_command/index.php
ace-n: Done.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,171 @@
...
+
+    return json_encode($responseJson);
+}
+// [END functions_slack_format]
+
+// [START functions_slack_request]
+/**
+* Send the user's search query to the Knowledge Graph API.
+*/
+function searchKnowledgeGraph($query): Google_Service_Kgsearch_SearchResponse
```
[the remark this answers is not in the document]
ace-n: I presume you're referring to `Google_Service_Kgsearch_SearchResponse` here?

(I added a `use` statement for that in the `functions_slack_setup` section near the top of the file, but I'm fairly sure it's a no-op.)

functions/slack_slash_command/index.php
ace-n: I think my main concern here is actually that the local testing framework (in particular the `UnitTest`s) doesn't execute these "global" statements.

If we can find a way to address that, I can move these into global scope + delete the comment outright. (Given that **a)** making these global seems idiomatic in PHP, and **b)** we call out PHP's caching behavior - or lack thereof - in `functions_tips_scopes`, I'm happy to drop the mention of it here altogether.)

functions/slack_slash_command/index.php
ace-n: Happy to do this for primitives and objects, but is there a good way to do this for collections (arrays/maps)? (It [doesn't seem like](https://stackoverflow.com/a/34273821) PHP collections allow their constituent values to be type-checked.)

functions/slack_slash_command/index.php
ace-n: I believe `itemListElement` is supposed to be defined (but empty!) _even if_ there are no matching results. (There _is_ a test case for this behavior.)

Happy to add a null-coalescing operator for defensive programming though (e.g. to mitigate errors and/or unexpected changes at the API level), if you think that's worth doing.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,171 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// [START functions_slack_setup]
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use GuzzleHttp\Psr7\Response;
+
+require_once './vendor/autoload.php';
```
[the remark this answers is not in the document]
ace-n: Done.

functions/slack_slash_command/test/TestCasesTrait.php
```diff
@@ -0,0 +1,114 @@
...
+
+        return [
```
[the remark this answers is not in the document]
ace-n: Done.

functions/slack_slash_command/index.php
ace-n: Yes - I've added a comment to this effect.

functions/slack_slash_command/index.php
ace-n: For better or for worse - currently, the other samples don't explicitly handle this (and instead pass it off to the runtime itself to handle).

That said, I'm open to changing this pattern if you think it's worthwhile. (Aside: I don't know what the consensus is on "whether (or not) to handle exceptional behavior in samples". Maybe @jmdobry or @JustinBeckwith have some opinions on this?)

functions/slack_slash_command/index.php
grayside: We can typehint as array, and sometimes worth defaulting to an empty array as well.

functions/slack_slash_command/index.php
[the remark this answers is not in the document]
grayside: Yes. Everything should have a use statement. If nothing else it's how the IDE will identify what you're referencing.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,174 @@
...
+
+// [START functions_verify_webhook]
+/**
+ * Verify that the webhook request came from Slack.
+ */
+function isValidSlackWebhook(ServerRequestInterface $request): bool
+{
+    /**
+     * PHP Functions Framework does not support "global"/instance-scoped
+     * constants, so we fetch these values within PHP functtions themselves.
```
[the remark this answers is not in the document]
bshaffer: Yes, I agree. Only if in the context of the documentation, we need to add the callout to differentiate from the other languages.

functions/slack_slash_command/index.php
ace-n: @grayside agreed - I'll stick to cross-sample consistency for now.

Perhaps it's worth filing a bug internally to revise this later (across languages), though?

functions/slack_slash_command/index.php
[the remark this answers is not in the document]
ace-n: > Only if in the context of the documentation, we need to add the callout to differentiate from the other languages.

Yep - that's my concern. I'll reword this.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,174 @@
...
+// [START functions_slack_request]
+/**
+ * Send the user's search query to the Knowledge Graph API.
+ */
+function searchKnowledgeGraph($query): Google_Service_Kgsearch_SearchResponse
+{
+    /**
+     * PHP Functions Framework does not support "global"/instance-scoped
+     * constants, so we fetch these values within PHP functions themselves.
+     */
```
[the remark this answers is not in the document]
ace-n: Reworded, per the other thread.

functions/slack_slash_command/index.php
ace-n: Used `object` here.

functions/slack_slash_command/test/SystemTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Google\Cloud\TestUtils\CloudFunctionLocalTestTrait;
+
+require_once __DIR__ . '/TestCasesTrait.php';
+
+/**
+ * Class SystemTest.
+ */
+class SystemTest extends TestCase
```
bshaffer, line 30: Request to rename this to `LocalTest.php`

bshaffer (approved):
One nit, otherwise nice work!

functions/slack_slash_command/test/SystemTest.php
ace-n: Done.

ace-n:
The tests are failing with:

> InvalidArgumentException: **Missing required property: entryPoint** in /tmpfs/src/github/php-docs-samples/testing/vendor/google/cloud-tools/src/TestUtils/GcloudWrapper/CloudFunction.php:83

@grayside is this something a Kokoro re-run will solve? If so, would you mind kicking that off once you've finished merging changes to `php-tools`?

(If not...happy to update this PR accordingly. 🙂)

functions/slack_slash_command/test/LocalTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Google\Cloud\TestUtils\CloudFunctionLocalTestTrait;
+
+require_once __DIR__ . '/TestCasesTrait.php';
+
+/**
+ * Class SystemTest.
+ */
+class SystemTest extends TestCase
```
grayside, line 30: File Name doesn't match class name. I believe in #1213 we are discussing standardizing on `IntegrationTest`.

functions/slack_slash_command/test/TestCasesTrait.php
```diff
@@ -0,0 +1,106 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+declare(strict_types=1);
+
+namespace Google\Cloud\Samples\Functions\SlackSlashCommand\Test;
+
+function _valid_headers($body): array
+{
+
```
grayside, line 23: nit: extra line

functions/slack_slash_command/test/TestCasesTrait.php
```diff
@@ -0,0 +1,106 @@
...
+    // Calculate test case signature
+    $key = getenv('SLACK_SECRET');
```
grayside, line 25: nit: this is an undocumented requirement. There should be a $this->requireEnv() statement somewhere.

functions/slack_slash_command/test/UnitTest.php
```diff
@@ -0,0 +1,70 @@
...
+namespace Google\Cloud\Samples\Functions\SlackSlashCommand\Test;
+
+use GuzzleHttp\Psr7\ServerRequest;
+use Psr\Http\Message\ResponseInterface;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Unit tests for the Cloud Function.
+ */
+class UnitTest extends TestCase
```
grayside, line 29: Unit test shouldn't be dependent on integrations with Slack/Google APIs. If we're not going to implement mocks and the integration/system/etc test isn't too slow, I think we can delete the UnitTest.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Send the user's search query to the Knowledge Graph API.
+ */
+function searchKnowledgeGraph(string $query): Google_Service_Kgsearch_SearchResponse
+{
+    /**
+     * Unlike most other Cloud Functions languages, PHP does not preserve
+     * global values between requests. Since we don't use this value elsewhere
+     * in the sample, it's easiest to declare it within this method itself.
+     */
```
grayside, line 115: Ditto from `isValidSlackWebhook()`.

grayside (changes requested):

functions/slack_slash_command/test/TestCasesTrait.php
ace-n: Done.

functions/slack_slash_command/test/LocalTest.php
ace-n: Renamed to `IntegrationTest`

functions/slack_slash_command/test/UnitTest.php
ace-n: I'm not opposed. @bshaffer sound OK to you?

functions/slack_slash_command/test/DeployTest.php
```diff
@@ -0,0 +1,79 @@
...
+            $output = (string) $response->getBody();
+            $this->assertContains($expected, $output, $label . ': contains');
+        }
+    }
+
+    protected static function deployFlags(array $flags = []): array
+    {
+        // Forward required env variables to Cloud Functions
+        $envVars = 'SLACK_SECRET=' . getenv('SLACK_SECRET') . ',';
+        $envVars .= 'KG_API_KEY=' . getenv('KG_API_KEY');
```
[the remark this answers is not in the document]
ace-n: Done.

functions/slack_slash_command/index.php
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Verify that the webhook request came from Slack.
+ */
+function isValidSlackWebhook(ServerRequestInterface $request): bool
+{
+    /**
+     * Unlike most other Cloud Functions languages, PHP does not preserve
+     * global values between requests. Since we don't use this value elsewhere
+     * in the sample, it's easiest to declare it within this method itself.
+     */
```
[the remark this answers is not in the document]
ace-n: Got it - deleted this comment.

functions/slack_slash_command/index.php
ace-n: Comment removed.

functions/slack_slash_command/test/TestCasesTrait.php
ace-n: Added it in the tests themselves.

(I think this is attached to some of the `Trait`s we use, so I'm doubtful if we can use it within `TestCaseTrait.php` - though that might be the ideal solution.)

functions/slack_slash_command/test/DeployTest.php
```diff
@@ -0,0 +1,79 @@
...
+            $label . ': status code'
+        );
+
+        if ($expected !== null) {
+            $output = (string) $response->getBody();
+            $this->assertContains($expected, $output, $label . ': contains');
+        }
+    }
+
+    protected static function deployFlags(array $flags = []): array
```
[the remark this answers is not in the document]
ace-n: Good catch, updated.

functions/slack_slash_command/test/DeployTest.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        // Forward required env variables to Cloud Functions
+        $envVars = 'SLACK_SECRET=' . getenv('SLACK_SECRET') . ',';
+        $envVars .= 'KG_API_KEY=' . getenv('KG_API_KEY');
+
+        $flags['--update-env-vars'] = '"' . $envVars . '"';
```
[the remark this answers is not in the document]
ace-n: Apparently not! (I switched to `doDeploy()`, which doesn't appear to require this.)

functions/slack_slash_command/test/UnitTest.php
bshaffer: I am ambivalent, but it sounds ok to me! The integration tests should be sufficient.

ace-n: OK - I'll remove this.

functions/slack_slash_command/test/DeployTest.php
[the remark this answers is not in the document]
grayside: Require should be before use: this will create a somewhat fuzzy error on gcloud deploy and never call testFunction().

functions/slack_slash_command/test/IntegrationTest.php
```diff
@@ -0,0 +1,67 @@
...
+    public function testFunction(
+        $label,
+        $body,
+        $method,
+        $expected,
+        $statusCode,
+        $headers
+    ): void {
+        $this->requireEnv('SLACK_SECRET');
+        $this->requireEnv('KG_API_KEY');
```
grayside, line 49: I don't see a customization of the `doRun()` method, so presumably the function framework is being started without these variables. Is test coverage deep enough if it passes without those variables present in the function?

grayside, line 49: nit: Should check these variables before the function is started/deployed to ensure fast failure and clear error message.

grayside (changes requested):

functions/slack_slash_command/test/IntegrationTest.php
ace-n: Thanks for clarifying - I wasn't quite sure where to put these. Moved to `doRun()`.

ace-n: Done - moved checks to `doRun()` as suggested above.

functions/slack_slash_command/test/DeployTest.php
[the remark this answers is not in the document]
ace-n: Moved to `doRun()` - I assume that will suffice?

grayside (approved):
Thanks for addressing the many rounds of feedback on this PR!

PR #1213, opened 2020-10-30
feat(functions/helloworld_storage): initial commit

grayside:
The tests for this depend on https://github.com/GoogleCloudPlatform/php-tools/pull/91.

## Change Summary

* Adds `functions_helloworld_storage` sample, canonical for CloudEvent functions.
* Solves unit testing log output by defining the stream as a global and swapping it out with php://memory in the test.
* There are four test classes:
  * **SampleUnitTest**: Self-contained implementation for `functions_storage_unit_test`
  * **SampleIntegrationTest**: Self-contained implementation for `functions_storage_integration_test`.
  * **IntegrationTest**: php-tools based implementation, cross-checks the system tests and provides canonical implementation for CloudEvent functions that do not host a sample test.
  * **DeployTest:** Canonical deployment test for CloudEvent functions.

## Emergent Patterns

* Suggests standardizing on `$log` variable for stream name where we use `fwrite()` to create logs.
* DeployTest includes two helpers: `triggerStorageUpload()` and `processFunctionLogs()` that may be worth considering for inclusion in CloudFunctionDeploymentTrait.
* Realized Functions Testing guide & samples use different test naming. To avoid confusion, we should follow-up with rename of SystemTest classes to IntegrationTest.


functions/helloworld_storage/test/DeployTest.php
```diff
@@ -0,0 +1,161 @@
+<?php
+/**
+ * Copyright 2020 Google LLC.
```
ace-n, line 3: Nit: perhaps we can use this for `functions_storage_system_test`?

(All we'd have to do is add a region tag, I think.)

ace-n (approved):

ace-n:
> DeployTest includes two helpers: triggerStorageUpload() and processFunctionLogs() that may be worth considering for inclusion in CloudFunctionDeploymentTrait.
> Realized Functions Testing guide & samples use different test naming. To avoid confusion, we should follow-up with rename of SystemTest classes to IntegrationTest.

Agreed on both. @grayside mind creating GitHub issues (or PRs) for these (for tracking purposes)?

functions/helloworld_storage/test/DeployTest.php
grayside: Potentially, however I'm a bit uncomfortable in that some of the use statements and traits are magically responsible for deployment. I'd rather include the deployment directly in the test sample.

Because of those considerations leaving that iteration outside scope for this PR, which already is pretty heavy as a canonical cloudevent sample and carrying two different snippets.

grayside:
I will file issue follow-ups once this gets a bit further through review by @ace-n and @bshaffer

ace-n (approved):

bshaffer (approved):

functions/helloworld_storage/test/SampleUnitTest.php
```diff
@@ -0,0 +1,91 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+declare(strict_types=1);
+
+// [START functions_storage_unit_test]
```
bshaffer, line 20: Do we really want region tags here? Is this showing up in the docs?

bshaffer: If these are actual samples showing up in the docs, then they shouldn't be in the `test` directory.

grayside: Similar to [helloworld_http/test/SampleUnitTest.php](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/functions/helloworld_http/test/SampleUnitTest.php), they are actual samples in the docs of working tests of the function sample.

I'm happy to move them out of the test directory, but I still want phpunit to run them. What's your preference?

bshaffer: Interesting! Looks like these samples aren't using `google/cloud-tools` in the tests... that's good, let's make sure we don't actually recommend that library to our users, since it's internal for now.

I'd say move the ones that are samples with region tags into the root (so `helloworld_http/SampleUnitTest.php`) and just ensure it's run via the `phpunit.xml.dist` file (changing ` <directory>test</directory>` to ` `<directory>.</directory>` should be sufficient)

grayside:
Various minor fixes, and deletes the created Cloud Storage object at the end of the test.

functions/helloworld_storage/test/SampleUnitTest.php
grayside: Yep, I took care to avoid external dependencies and simplify the code as much as possible. I'll go ahead and move that up in this and the helloworld_http case.

grayside:
This will not pass CI testing until the branch is updated after #1215 lands.

functions/helloworld_storage/SampleIntegrationTest.php
```diff
@@ -0,0 +1,136 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+declare(strict_types=1);
+
+// [START functions_storage_integration_test]
+
+namespace Google\Cloud\Samples\Functions\HelloworldStorage\Test;
```
bshaffer, line 22: nit: Should the namespace `Test` be removed to match the other sample test?

functions/helloworld_storage/index.php
```diff
@@ -0,0 +1,37 @@
...
+
+// [START functions_helloworld_storage]
+
+use Google\CloudFunctions\CloudEvent;
+
+$log = fopen('php://stderr', 'wb');
+
+function helloGCS(CloudEvent $cloudevent)
+{
+    global $log;
```
bshaffer, line 26: Using globals in PHP is typically not a good practice. 

I'd prefer to define `$log` inside the function. If we'd like to change the output for testing, then we can use an environment variable to change the log output:

```php
$tmpFile = sys_get_temp_dir() . '/index.php';
file_put_contents(
    $tmpFile,
    str_replace('php://stderr', 'php://memory', file_get_contents(__DIR__ . '/index.php))
);
require_once($tmpFile);
```

It seems like a very uncommon thing for someone to need to test the log output of their application. If this is actually a common thing, we should come up with a more robust solution, such as using monolog.

bshaffer (changes requested):

functions/helloworld_storage/index.php
grayside: Yes, I think this will be very common.

Tests which focus on integration with other systems can check things like new objects in Cloud Storage or use mocks, but log tests are very common. That's why the DeployTest helper in this PR is a candidate to move to a trait later.

bshaffer: Then I think this is fine for now but we really should provide a better solution in the function framework!

grayside: In discussion with Brent yesterday, we decided to switch to an env override model instead of a global variable. I think it looks a small bit more complicated in the function but uses a cleaner practice.

In doing that, I also added an example to SampleUnitTest of the code you need in the function implementation for this to work.

functions/helloworld_http/phpunit.xml.dist
```diff
@@ -17,7 +17,7 @@
 <phpunit bootstrap="../../testing/bootstrap.php" convertWarningsToExceptions="false">
     <testsuites>
         <testsuite name="Cloud Functions Hello World HTTP Test Suite">
-            <directory>test</directory>
+            <directory>.</directory>
```
bshaffer, line 20: needs `<exclude>vendor</exclude>`

bshaffer (approved):
One change needed, otherwise LGTM!

functions/helloworld_http/phpunit.xml.dist
grayside: 😱

PR #1214, opened 2020-11-04
chore(functions): Update tests to use new php-tools extension methods

grayside:
Note I expect the `env_vars` System Test to fail unless the CI environment has a "FOO" variable set.

This is a pre-existing bug, and likewise a remaining gap in the php-tools implementation of `CloudFunction::run()`

### DIFF
--- a/functions/concepts_filesystem/test/DeployTest.php
+++ b/functions/concepts_filesystem/test/DeployTest.php
@@ -37,7 +37,11 @@ class DeployTest extends TestCase
     use CloudFunctionDeploymentTrait;
     use TestCasesTrait;
 
-    private static $name = 'listFiles';
+    private static function initFunctionProperties(array $props = [])
+    {
+        $props['entryPoint'] = 'listFiles';
+        return $props;
+    }
 
     /**
       * @dataProvider cases
--- a/functions/env_vars/test/DeployTest.php
+++ b/functions/env_vars/test/DeployTest.php
@@ -44,8 +44,21 @@ class DeployTest extends TestCase
     }
 
     /**
-      * @dataProvider cases
-      */
+     * Deploy the Cloud Function, called from DeploymentTrait::deployApp().
+     *
+     * Overrides CloudFunctionDeploymentTrait::doDeploy().
+     */
+    private static function doDeploy()
+    {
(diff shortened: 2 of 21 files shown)


functions/concepts_filesystem/test/DeployTest.php
```diff
@@ -37,7 +37,11 @@ class DeployTest extends TestCase
     use CloudFunctionDeploymentTrait;
     use TestCasesTrait;
 
-    private static $name = 'listFiles';
+    private static function initFunctionProperties(array $props = [])
+    {
+        $props['entryPoint'] = 'listFiles';
```
ace-n, line 42: Nit: I'm guessing it doesn't make sense to simply set `$props['entryPoint'] = $name`, because we may want to specify other values in this function?

functions/env_vars/test/DeployTest.php
```diff
@@ -37,11 +37,28 @@ class DeployTest extends TestCase
...
+        self::$bucket = self::requireEnv('GOOGLE_STORAGE_BUCKET');
+        return self::$fn->deploy([
+            '--update-env-vars' => 'FOO=bar',
+        ]);
+    }
 
     /**
-      * @dataProvider cases
-      */
+     * @dataProvider cases
```
ace-n, line 60: Nit: can we enforce this via `php-cs-fixer` (or a similar automated method) if we aren't already?

ace-n (approved):

functions/concepts_filesystem/test/DeployTest.php
grayside: Discussing the patterns should be done in https://github.com/GoogleCloudPlatform/php-tools/pull/91.

functions/env_vars/test/DeployTest.php
grayside: Outside scope, but I also thought about that, would be good to find out.

grayside:
This includes a fix for the envVar sample, which was depending on test environment to have the correct environment variables.

However, the structure of the test is flawed, in that only the existing test case can be tested. In order to iterate the test with multiple env vars we need to stand up and shut down the functions framework server multiple times. Outside the scope of this PR.

grayside:
We took another approach.

PR #1215, opened 2020-11-05
chore(functions): Revise test coverage for CI utility updates

grayside:
* Rename $name in functions tests to $entryPoint
* Fix env_vars sample DeployTest and SystemTest to use first test case env var.

### DIFF
--- a/functions/env_vars/test/DeployTest.php
+++ b/functions/env_vars/test/DeployTest.php
@@ -42,7 +42,7 @@ class DeployTest extends TestCase
     /**
      * Deploy the Function.
      *
-     * Overrides CloudFunctionLocalTestTrait::doDeploy().
+     * Overrides CloudFunctionDeploymentTrait::doDeploy().
      */
     private static function doDeploy()
     {
--- a/functions/env_vars/test/DeployTest.php
+++ b/functions/env_vars/test/DeployTest.php
@@ -46,9 +46,10 @@ class DeployTest extends TestCase
      */
     private static function doDeploy()
     {
+        // Use the first test case as the source of the deployed environment variable.
         $cases = self::cases();
-        $case = reset($cases);
-        self::$fn->deploy([$case['varName'] => $case['varValue']]);
+        $env = sprintf("%s=%s", $cases[0]['varName'], $cases[0]['varValue']);
+        self::$fn->deploy(['--update-env-vars' => $env]);
     }
 
     /**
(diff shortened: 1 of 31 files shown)


functions/env_vars/test/DeployTest.php
```diff
@@ -37,7 +37,19 @@ class DeployTest extends TestCase
...
+    /**
+     * Deploy the Function.
+     *
+     * Overrides CloudFunctionLocalTestTrait::doDeploy().
+     */
+    private static function doDeploy()
+    {
+        $cases = self::cases();
+        $case = reset($cases);
+        self::$fn->deploy([$case['varName'] => $case['varValue']]);
```
bshaffer, line 51: I am confused as to what's happening here

functions/helloworld_http/composer.json
```diff
@@ -1,5 +1,8 @@
 {
     "require": {
         "google/cloud-functions-framework": "^0.6"
+    },
+    "require-dev": {
+        "phpunit/phpunit": "^9.4"
```
bshaffer, line 6: We should use the phpunit in `../../testing/vendor/bin/phpunit`. Do you have a reason you're adding this here?

grayside: Accident, will revert.

functions/env_vars/test/DeployTest.php
grayside: This has a bug, which I've fixed. Also clarified the code a bit and added a comment.

tl;dr: The env_var sample got merged but the DeployTest and SystemTest had to have been failing without a custom variable in the running environment. I've switched the code to pull that var from the dataProvider, which only works because there's just one test case.

Fixing this more completely is out of scope, I only did this to make sure this PR could pass.

functions/env_vars/test/DeployTest.php
```diff
@@ -37,7 +37,20 @@ class DeployTest extends TestCase
     use CloudFunctionDeploymentTrait;
     use TestCasesTrait;
 
-    private static $name = 'envVar';
+    private static $entryPoint = 'envVar';
+
+    /**
+     * Deploy the Function.
+     *
+     * Overrides CloudFunctionLocalTestTrait::doDeploy().
```
ace-n, line 45: Nit: this seems like `CloudFunctionDeploymentTrait`?

functions/helloworld_http/composer.json
```diff
@@ -1,5 +1,7 @@
 {
     "require": {
         "google/cloud-functions-framework": "^0.6"
+    },
+    "require-dev": {
```
bshaffer: remove unneeded `require-dev`!

bshaffer (approved):
one nit, otherwise looks good!

PR #1216, opened 2020-11-05
docs(run): update region tag

averikitsch:


grayside (approved):
Expected phase 1 of region tag prefix migration for Cloud Run samples.

bshaffer (approved):

PR #1218, opened 2020-11-09
chore(functions/log_helloworld): mention var_dump

ace-n:
Ran into this while debugging; figured it was worth documenting.

PR #1220, opened 2020-11-11
chore(functions/log): add more logging methods

### LINKED ISSUE #1219: functions/log_helloworld: Add more logging/troubleshooting guidance
This sample goes a bit beyond logging to cover common log-based troubleshooting techniques. In that case, I would expect some more details to be explored in the sample:

1. What does `trigger_error()` do?
2. How do I get `var_dump()` output to the logs?

```php
ob_start();
var_dump('var_dump details');
$dump = ob_get_contents();
ob_end_clean();
fwrite($log, $entry);
```

3. What do we recommend as a way to use logs for troubleshooting?

```php
$entry = [
  'message' => 'var_dump output',
  'var_dump' => json_encode($var_to_dump),
  'severity' => 'debug',
];
fwrite($log, $entry);
```

In Cloud Logging UI, expand the log entry to see the details and var_dump has the variable in structured form. This also makes it easier to build log queries.

ace-n:
Fixes #1219

(I was going to do this in #1218, but @bshaffer ninja'd me. 🙂

### DIFF
--- a/functions/log_helloworld/index.php
+++ b/functions/log_helloworld/index.php
@@ -37,9 +37,18 @@ function helloLogging(ServerRequestInterface $request): string
     // This doesn't log anything
     error_log('error_log does not log in Cloud Functions!');
     
+    // This will log an error message and immediately terminate the function execution
+    // trigger_error('fatal errors are logged!');
+    
     // For HTTP functions, this is added to the HTTP response
     // For CloudEvent functions, this does nothing
     var_dump('var_dump goes to HTTP response for HTTP functions');
+    
+    // You can also dump variables using var_export() and forward
+    // the resulting string to Cloud Logging via an fwrite() call.
+    $entry = '';
+    var_export('var_export output can be captured.', $entry);
+    fwrite($log, $entry);
 
     // Functions must return a String or PSR-7 Response object
     return '';
--- a/functions/log_helloworld/index.php
+++ b/functions/log_helloworld/index.php
@@ -46,8 +46,7 @@ function helloLogging(ServerRequestInterface $request): string
     
     // You can also dump variables using var_export() and forward
     // the resulting string to Cloud Logging via an fwrite() call.
-    $entry = '';
-    var_export('var_export output can be captured.', $entry);
+    $entry = var_export('var_export output can be captured.', true);
     fwrite($log, $entry);
 
     // Functions must return a String or PSR-7 Response object


grayside (approved):

functions/log_helloworld/index.php
```diff
@@ -37,9 +37,21 @@ function helloLogging(ServerRequestInterface $request): string
...
     var_dump('var_dump goes to HTTP response for HTTP functions');
+    
+    // You can also capture var_dump output using Output Buffering
+    // and forward it to Cloud Logging via an fwrite() call.
+    // See https://www.php.net/manual/en/book.outcontrol.php
+    ob_start();
+    var_dump('Captured var_dump calls can be logged.');
+    $entry = ob_get_contents();
+    ob_end_clean();
+    fwrite($log, $entry);  // Forward var_dump entry to fwrite.
```
bshaffer, line 54: This really doesn't seem necessary... why not just recommend `var_export`? I don't think we need to cover output buffering as a part of this example...

bshaffer (changes requested):

functions/log_helloworld/index.php
grayside: I assumed Ace had a reason to emphasize this, I think showcasing `var_export()` (forgot about that one!) and my json_encode() snippet from the #1219 would be more valuable.

bshaffer: I agree!

ace-n: `var_export()`...hmm, TIL. 🤔

I'll use that instead.

tmatsuo:
@ace-n 
FYI, it seems like the snippet-bot is updating the comment which is deleted.

functions/log_helloworld/index.php
```diff
@@ -37,9 +37,18 @@ function helloLogging(ServerRequestInterface $request): string
...
+    // trigger_error('fatal errors are logged!');
+    
     // For HTTP functions, this is added to the HTTP response
     // For CloudEvent functions, this does nothing
     var_dump('var_dump goes to HTTP response for HTTP functions');
+    
+    // You can also dump variables using var_export() and forward
+    // the resulting string to Cloud Logging via an fwrite() call.
+    $entry = '';
+    var_export('var_export output can be captured.', $entry);
```
bshaffer, line 50: This won't work as written here... it should be this

```php
$entry = var_export('var_export output can be captured.', true);
fwrite($log, $entry);
```

ace-n: Done.

bshaffer (approved):

PR #1227, opened 2020-11-20
docs: cleanup README for eventarc

bshaffer:
This replaces the manual steps to deploy to Cloud Run with the Cloud Run button

### DIFF
--- a/eventarc/generic/README.md
+++ b/eventarc/generic/README.md
@@ -83,20 +83,6 @@ source for our sample.
 
     *This will take a few minutes to build and deploy.*
 
-1. [Build the container using Cloud Build](https://cloud.google.com/run/docs/building/containers#builder)
-
-    ```sh
-    gcloud builds submit --tag gcr.io/$(gcloud config get-value project)/eventarc-generic-php
-    ```
-
-1. [Deploy the container](https://cloud.google.com/run/docs/deploying#service)
-
-    ```sh
-    gcloud run deploy eventarc-generic-php \
-      --image gcr.io/$(gcloud config get-value project)/eventarc-generic-php \
-      --allow-unauthenticated
-    ```
-
     The command line will display the service URL when deployment is complete.
 
 ### Create an Eventarc Trigger
--- a/eventarc/generic/README.md
+++ b/eventarc/generic/README.md
@@ -79,11 +79,11 @@ source for our sample.
 
 ## Run the sample on Cloud Run
 
-1. [<img src="https://storage.googleapis.com/cloudrun/button.svg" alt="Run on Google Cloud" height="30"/>][run_button_generic]
+### Deploy to Cloud Run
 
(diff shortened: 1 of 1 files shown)


grant (commented):
I don't think moving the Run button into the steps makes sense. The steps show how to deploy manually and the button does that for you, so they are exclusive. The button isn't the first step of running the sample on Cloud Run.

bshaffer:
@grant take a look now. It replaces the cloud build / manual deploy instructions. Having the button at the top without context doesn't serve a purpose. So this is a better flow all around.

grant (changes requested):
Sorry, I believe the PR description and title are incorrect. This PR is not cleaning up the sample. It looks like it's forcing the use of the Cloud Run button rather than typical `gcloud` commands.

Using a UI / the Cloud Run button should not be a required step. That will make this sample not usable in a terminal environment for example. In most samples from what I know, we include `gcloud` steps for required actions via the terminal.

What's the goal of this change?

grant:
> @grant take a look now. It replaces the cloud build / manual deploy instructions. Having the button at the top without context doesn't serve a purpose. So this is a better flow all around.

Maybe simply remove the button?

bshaffer:
@grant Why would we remove the button? That's how they deploy to Cloud Run.

bshaffer:
Deploying using the cloud run button is more convenient than using gcloud commands. It's a better user experience.

We could include both, which complicates things a little bit, or just link to the docs on deploying with gcloud. I think for the purpose of the README, either would be sufficient.

grant:
Thanks for the PR again.

> @grant Why would we remove the button? That's how they deploy to Cloud Run. And it's easier than running the `gcloud` steps manually.

We could remove the button as the button doesn't make sense here and it does not function. Does the button function with this PR? Removing it might make sense if it doesn't function. Has the button been tested – and what is the result?

I disagree from a user experience POV that removing the `gcloud build` and `gcloud run deploy` commands provide a better UX. Why aren't we doing that in the Run sample? – that sample is simpler and I could imagine first changing that sample, then modifying this one.

eventarc/generic/README.md
```diff
@@ -81,21 +79,11 @@ source for our sample.
...
+### Deploy to Cloud Run
 
-1. [Deploy the container](https://cloud.google.com/run/docs/deploying#service)
-
-    ```sh
-    gcloud run deploy eventarc-generic-php \
-      --image gcr.io/$(gcloud config get-value project)/eventarc-generic-php \
-      --allow-unauthenticated
-    ```
+[<img src="https://storage.googleapis.com/cloudrun/button.svg" alt="Run on Google Cloud" height="30"/>][run_button_generic]
```
grant, line 84: This button doesn't deploy a Run service called `eventarc-generic-php`, so the rest of this README fails to work.

PR #1234, opened 2020-11-30
feat(spanner): add helper for sample function execution

bshaffer:
RFC for a way to get rid of `symfony/command` component in our samples by allowing users to execute each sample directly:

```sh 
// execute with improper params
$ php spanner/src/add_column.php
Usage: add_column.php $instanceId $databaseId

	@param string $instanceId The Spanner instance ID.
	@param string $databaseId The Spanner database ID.
```
```sh
// execute with missing vendor/autoload
$ php spanner/src/add_column.php instanceId databseId
You must run "composer install" in the sample root (/Users/betterbrent/Code/php-docs-samples/spanner/)
```

This will remove a lot of boilerplate code and overhead we currently have with the various `spanner.php` CLI tools in most of our samples. All that is required is for the following lines to be added at the bottom of every `[sample].php` file:

```php
require_once __DIR__ . '/../../testing/sample_helpers.php';
\Google\Cloud\Samples\execute_sample(__FILE__, __NAMESPACE__);
```


skuruppu:
FYI @larkee

dwsupplee (approved):
Clever!

PR #1236, opened 2020-12-02
feat: added a profanity filter sample

b-loved-dreamer:
New sample that shows the use of profanity filtering


speech/src/profanity_filter_gcs.php
```diff
@@ -0,0 +1,53 @@
...
+// set string as audio content
+$audio = (new RecognitionAudio())
+    ->setUri($uri);
+
+// set config
+$config = (new RecognitionConfig())
+    ->setEncoding(AudioEncoding::FLAC)
+    ->setSampleRateHertz(16000)
+    ->setLanguageCode('en-US')
+    ->setProfanityFilter(TRUE);
```
bshaffer, line 37: `true` needs to be lowercase

speech/src/profanity_filter.php
```diff
@@ -0,0 +1,53 @@
...
+// set string as audio content
+$audio = (new RecognitionAudio())
+    ->setUri($uri);
+
+// set config
+$config = (new RecognitionConfig())
+    ->setEncoding(AudioEncoding::FLAC)
+    ->setSampleRateHertz(16000)
+    ->setLanguageCode('en-US')
+    ->setProfanityFilter(TRUE);
```
bshaffer, line 37: `true` needs to be lowercase

speech/test/speechTest.php
```diff
@@ -22,7 +22,7 @@
 
 class speechTest extends TestCase
 {
-    use TestTrait;
+    // use TestTrait;
```
bshaffer, line 25: This should not be commented out

speech/src/profanity_filter_gcs.php
```diff
@@ -0,0 +1,53 @@
...
+# [START profanity_filter]
+# Includes the autoloader for libraries installed with composer
+require __DIR__ . '/../vendor/autoload.php';
+
+use Google\Cloud\Speech\V1\SpeechClient;
+use Google\Cloud\Speech\V1\RecognitionAudio;
+use Google\Cloud\Speech\V1\RecognitionConfig;
+use Google\Cloud\Speech\V1\RecognitionConfig\AudioEncoding;
+
+
```
bshaffer, line 24: nit: remove unnecessary newline

speech/src/profanity_filter.php
```diff
@@ -0,0 +1,53 @@
...
+# [START profanity_filter]
+# Includes the autoloader for libraries installed with composer
+require __DIR__ . '/../vendor/autoload.php';
+
+use Google\Cloud\Speech\V1\SpeechClient;
+use Google\Cloud\Speech\V1\RecognitionAudio;
+use Google\Cloud\Speech\V1\RecognitionConfig;
+use Google\Cloud\Speech\V1\RecognitionConfig\AudioEncoding;
+
+
```
bshaffer, line 24: nit: remove unnecessary newline

speech/src/profanity_filter.php
```diff
@@ -0,0 +1,53 @@
...
+/** The Cloud Storage object to transcribe */
+$uri = 'gs://cloud-samples-tests/speech/brooklyn.flac';
```
bshaffer, line 26: This should be using a local file rather than a file from gcs.

speech/test/speechTest.php
```diff
@@ -92,6 +92,7 @@ public function provideTranscribe()
             ['transcribe_async', __DIR__ . '/data/audio32KHz.raw'],
             ['transcribe_async_gcs', 'gs://' . self::$bucketName . '/speech/audio32KHz.raw'],
             ['transcribe_async_words', __DIR__ . '/data/audio32KHz.raw'],
+            ['profanity_filter', 'gs://' . self::$bucketName . '/speech/audio32KHz.raw'],
```
bshaffer, line 95: There should also be a `profanity_filter_gcs` to test both samples.

Also, you're passing an argument here but your sample does not allow for an argument. See one of the other speech samples for an example. At the top of the samples, it looks like this:

```php
if (count($argv) != 2) {
    return print("Usage: php base64_encode_audio.php AUDIO_FILE\n");
}
list($_, $audioFile) = $argv;

# [START base64_audio]
// $audioFile = 'path to an audio file';

//...
```

bshaffer (changes requested):

bshaffer (changes requested):
Again, you need to uncomment `use TestTrait;`. The tests are failing because `runSnippet` is undefined.

speech/test/speechTest.php
```diff
@@ -98,6 +98,8 @@ public function provideTranscribe()
             ['transcribe_async', __DIR__ . '/data/audio32KHz.raw'],
             ['transcribe_async_gcs', 'gs://' . self::$bucketName . '/speech/audio32KHz.raw'],
             ['transcribe_async_words', __DIR__ . '/data/audio32KHz.raw'],
+            ['profanity_filter_gcs', 'gs://' . self::$bucketName . '/speech/audio32KHz.raw'],
+            ['profanity_filter', __DIR__ . '/data/audio32KHz.raw'],
```
bshaffer, line 102: These tests are failing... possibly because there is no profanity in it, but I'm not sure. Is there a way we can test the profanity filter samples are working?

speech/test/speechTest.php
```diff
@@ -76,7 +76,7 @@ public function testTranscribe($command, $audioFile, $requireGrpc = false)
         if ($requireGrpc && !extension_loaded('grpc')) {
             self::markTestSkipped('Must enable grpc extension.');
         }
-        if (!self::$bucketName && in_array($command, ['transcribe_gcs', 'transcribe_async_gcs'])) {
+        if (!self::$bucketName && in_array($command, ['transcribe_gcs', 'transcribe_async_gcs', 'profanity_filter_gcs', 'profanity_filter'])) {
```
bshaffer, line 79: You can remove `profanity_filter` from this line... this line is only for samples using cloud storage (e.g. ending in `_gcs`)

bshaffer (changes requested):

speech/test/speechTest.php
b-loved-dreamer: I wouldn't know what would be acceptable for a file containing profanity. I will ask.

bshaffer: Me neither, but we should find a way to test this! And for some reason the tests you added are failing...

b-loved-dreamer: Thank you. I will look into this more closely.

bshaffer:
The behavior I'm seeing is that this works for FLAC but not for RAW.

bshaffer (approved):

tmatsuo:
FYI, python uses product prefixes like `speech_recognize_with_profanity_filter_gcs`

https://github.com/googleapis/python-speech/pull/93

PR #1238, opened 2020-12-10
refactor: updated the quickstart to use gcs

b-loved-dreamer:
This sample now takes a gcs resource audio file instead of a local


tmatsuo:
FYI, other languages are using region tags like `speech_transcribe_with_profanity_filter_gcs`. I recommend that you do the same.

bshaffer (approved):

PR #1239, opened 2020-12-11
docs: Wrap samples with future prefix

Strykrol:
Standardizing Security Command Center samples to use 'securitycenter' prefixing. 

- Wrapped existing samples to keep published doclinks unbroken
- Replace region tags that aren't published. 

Once this PR is through, published sample inclusions will be updated to use the new prefix, then I'll come through again and remove the unused block wraps.


securitycenter/src/receive_notification.php
```diff
@@ -22,7 +22,7 @@
 }
 list($_, $projectId, $subscriptionId) = $argv;
 
-// [START scc_receive_notifications]
+// [START securitycenter_receive_notifications]
```
bshaffer, line 25: Curious why this one was removed while the others were wrapped, but otherwise LGTM

bshaffer (approved):

securitycenter/src/receive_notification.php
Strykrol: This PHP sample is not published live in any docs, so fully changing this should be no-harm-no-foul.

securitycenter/src/list_notification.php
```diff
@@ -22,6 +22,7 @@
 }
 list($_, $organizationId) = $argv;
 
+// [START securitycenter_list_notification_configs]
```
bshaffer, line 25: The tests are failing because there's trailing whitespace at the end of this line

Strykrol: Done, thank you!

securitycenter/src/create_notification.php
```diff
@@ -22,6 +22,7 @@
 }
 list($_, $organizationId, $notificationConfigId, $projectId, $topicName) = $argv;
 
+// [START securitycenter_ create_notification_config]
```
tmatsuo, line 25: `[START securitycenter_ create_notification_config]`

There is a unnecessary space in the region tag part. Please fix.

securitycenter/src/create_notification.php
```diff
@@ -51,3 +52,4 @@
 printf('Notification config was created: %s' . PHP_EOL, $response->getName());
 
 // [END scc_create_notification_config]
+// [END securitycenter_ create_notification_config]
```
tmatsuo, line 55: ditto

tmatsuo (changes requested):

securitycenter/src/create_notification.php
Strykrol: Thank you!

securitycenter/src/create_notification.php
Strykrol: Good catch, thanks.

tmatsuo (approved):

PR #1241, opened 2020-12-15
feature(functions): add Firebase remote config sample

ace-n:


ace-n:
TODO: use stream wrappers.

functions/firebase_remote_config/composer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "require": {
+        "google/cloud-functions-framework": "^0.7.1",
+        "google/cloud-firestore": "^1.18",
+        "kreait/firebase-php": "^5.14"
+    },
+    "scripts": {
+        "start": [
+            "Composer\\Config::disableProcessTimeout",
+            "FUNCTION_SIGNATURE_TYPE=cloudevent FUNCTION_TARGET=firebaseFirestore php -S localhost:${PORT:-8080} vendor/bin/router.php"
```
grayside, line 10: FUNCTION_TARGET needs to match current function name.

functions/firebase_remote_config/index.php
```diff
@@ -0,0 +1,32 @@
...
+function firebaseRemoteConfig(CloudEvent $cloudevent)
+{
+    $log = fopen(getenv('LOGGER_OUTPUT') ?: 'php://stderr', 'wb');
+    
+    $data = $cloudevent->getData();
+
+    fwrite($log, 'Update type: ' . $data['updateType'] . PHP_EOL);
+    fwrite($log, 'Origin: ' . $data['updateOrigin'] . PHP_EOL);
+    fwrite($log, 'Version: ' . $data['versionNumber'] . PHP_EOL);
+}
```
grayside, line 31: Future consideration: This sample seems to output specific fields from this event. It seems like it may be changed with one of the following:

1. Replace with event documentation & generic sample to log top-level fields from an event.
2. Generate based on an event schema.

functions/firebase_remote_config/phpunit.xml.dist
```diff
@@ -0,0 +1,35 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<phpunit bootstrap="../../testing/bootstrap.php" convertWarningsToExceptions="false">
+    <testsuites>
+        <testsuite name="Cloud Functions Firebase Remote Config Test Suite">
+            <directory>test</directory>
+            <exclude>vendor</exclude>
```
grayside, line 21: `vendor` is not needed as an exclusion from the test directory.

functions/firebase_remote_config/composer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "require": {
+        "google/cloud-functions-framework": "^0.7.1",
+        "google/cloud-firestore": "^1.18",
+        "kreait/firebase-php": "^5.14"
```
grayside, line 5: Starting out-of-band discussion on this. Seems odd to use an unofficial Firebase library in these samples.

grayside (changes requested):
A quick pass. Overall looks good but a couple tweaks needed.

functions/firebase_remote_config/composer.json
ace-n: cc @samtstern - is this the recommended library for interacting with Firebase?

samtstern: It's the highest-quality third party Firebase SDK that I know of but not something we officially support or recommend in any of our samples.

I would not recommend using it in an official Cloud sample.

ace-n: Got it. What would the best alternative be? (Ideally something other than raw API requests!)

samtstern:
Unfortunately raw HTTP requests is my answer here.  The only product we
have a PHP SDK for is Firestore.

functions/firebase_remote_config/composer.json
ace-n: From @bshaffer - we can use the code below to do auth'd API requests:
```php
use Google\Auth\ApplicationDefaultCredentials;
use Google\Auth\CredentialsLoader;

$credentials = ApplicationDefaultCredentials::getCredentials();
$client = ClientLoader::makeHttpClient($credentials, [
    'base_url' => 'https://www.googleapis.com/taskqueue/v1beta2/projects/'
]);
$response = $client->get('myproject/taskqueues/myqueue');
```

functions/firebase_remote_config/composer.json
ace-n: Done.

functions/firebase_remote_config/phpunit.xml.dist
ace-n: Done.

functions/firebase_remote_config/composer.json
ace-n: This has been addressed. (We removed this library in favor of raw HTTP requests, as discussed above.)

functions/firebase_remote_config/index.php
ace-n: I wouldn't be opposed to a generation system. (Maybe ping @busunkim96 or @JustinBeckwith internally, as they might have more context.)

functions/firebase_remote_config/test/IntegrationTest.php
```diff
@@ -0,0 +1,84 @@
...
+
+        // The Cloud Function logs all data to stderr.
+        $actual = self::$localhost->getIncrementalErrorOutput();
+
+        // Confirm the status code.
+        $this->assertEquals($statusCode, $resp->getStatusCode());
+
+        // Verify the data properties are logged by the function.
+        foreach ($cloudevent->getData() as $property => $value) {
+            if (is_string($value)) {
```
grayside, line 79: Should there be a failing assertion if the value is not a string?

ace-n: In that case, `value` should `=== 2`. Added an assertion accordingly.

grayside (approved):

functions/tips_phpinfo/.gcloudignore
```diff
@@ -0,0 +1,3 @@
+test/
+build/
+vendor/
```
bshaffer, line 3: I think this file can be deleted?

functions/firebase_remote_config/composer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "require": {
+        "google/cloud-functions-framework": "^0.7.1",
+        "google/cloud-firestore": "^1.18",
+        "google/auth": "^1.14"
```
bshaffer, line 5: I believe this can be moved to `require-dev`, as it's only a dependency used in tests AFAICT

functions/firebase_remote_config/composer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "require": {
+        "google/cloud-functions-framework": "^0.7.1",
+        "google/cloud-firestore": "^1.18",
```
bshaffer, line 4: I don't see this dep being used anywhere

bshaffer (changes requested):
I might be missing something, but I don't see anything firebase in this sample - it looks like it just prints the cloud event data. Is that intentional?

functions/tips_phpinfo/.gcloudignore
ace-n: A stowaway! 🏴‍☠️ Removed.

functions/firebase_remote_config/composer.json
ace-n: Removed.

functions/firebase_remote_config/composer.json
ace-n: Done.

ace-n:
@bshaffer yeah - the "Firebase" here comes from the events [being triggered by](https://cloud.google.com/functions/docs/calling/firebase-remote-config) Firebase Remote Config.

bshaffer (approved):

PR #1246, opened 2020-12-17
feat: adds a new multi region sample

b-loved-dreamer:
Shows how to change the default endpoint


speech/src/multi_region_gcs.php
```diff
@@ -0,0 +1,61 @@
+<?php
+/**
+ * Copyright 2016 Google Inc.
```
bshaffer, line 3: New files should be 2021

speech/src/multi_region_gcs.php
```diff
@@ -0,0 +1,61 @@
...
+
+# The audio file's encoding, sample rate and language
+$config = new RecognitionConfig([
+    'encoding' => AudioEncoding::LINEAR16,
+    'sample_rate_hertz' => 16000,
+    'language_code' => 'en-US'
+]);
+
+# Specify a new endpoint.
+$options = array("apiEndpoint" => "eu-speech.googleapis.com");
```
bshaffer, line 43: This should use new PHP array syntax and single quotes:

```php
$options = ['apiEndpoint' => 'eu-speech.googleapis.com'];
```

speech/src/multi_region_gcs.php
```diff
@@ -0,0 +1,61 @@
...
+# Print most likely transcription
+foreach ($response->getResults() as $result) {
+    $alternatives = $result->getAlternatives();
+    $mostLikely = $alternatives[0];
+    $transcript = $mostLikely->getTranscript();
+    printf('Transcript: %s' . PHP_EOL, $transcript);
+}
+
+$client->close();
+
```
bshaffer, line 60: nit: remove extra newline after `$client->close()`

speech/test/speechTest.php
```diff
@@ -99,6 +99,7 @@ public function provideTranscribe()
             ['transcribe_async_gcs', 'gs://' . self::$bucketName . '/speech/audio32KHz.raw'],
             ['transcribe_async_words', __DIR__ . '/data/audio32KHz.raw'],
             ['profanity_filter_gcs', 'gs://' . self::$bucketName . '/speech/audio32KHz.raw'],
+            ['multi_region_gcs', 'gs://' . self::$bucketName . '/speech/audio32KHz.raw'],
```
bshaffer, line 102: This still needs to be added to `testTranscribe` (line 79 above).

So that this doesn't keep happening, change the `if` to the following:

```php
if (!self::$bucketName && '_gcs' == substr($command, -4)) {
```

bshaffer (changes requested):
A few small changes and this will be good to go!

bshaffer (approved):

PR #1256, opened 2021-01-14
docs: standardize region tagging pt.2

Strykrol:
This finishes up the work from PR #1247. New tags have proliferated on DevSite, and these old ones are no longer needed.


Strykrol:
@bshaffer when you get a moment can you help push this through to finish up? Thank you!

bshaffer (approved):

PR #1257, opened 2021-01-16
chore(functions): add missing HTTP test snippet

ace-n:
**Note:** this snippet would refer to some of our testing infrastructure (namely `CloudFunctionLocalTestTrait`) publicly. We may want to consider using something else in this sample if we don't want to suggest that to customers.


grayside:
This test was not intended to be the sample system test, which is why almost the entire test depends on our internal infrastructure and it's inside the test directory.

Issue #1240, opened 2020-12-15
Move processFunctionLogs functions into a library


ace-n:
I've noticed several (and counting) of our Cloud Functions samples' `DeployTest`s define a `processFunctionLogs` function.

Since these functions are largely similar (if not the same) across all these files, we should move this logic to a shared library (e.g. `php-tools`) or a single location within the repo (maybe the root-level `testing` directory?).

grayside:
I suggest following the existing pattern and putting this inside the CloudFunctionsDeployTrait or creating a new trait to encapsulate utilities for async testing. When the first example was merged, we discussed that being a useful follow-up as part of the 2nd or 3rd usage to make sure we understood the needed logic.

PR #1258, opened 2021-01-16
feat(functions): add retriable events sample

ace-n:
Addresses b/154660249


functions/tips_retry/index.php
```diff
@@ -0,0 +1,40 @@
...
+
+    if ($tryAgain) {
+        // Retry execution
+        throw new Exception('Retrying...');
+    } else {
+        // Don't retry execution
+        $log = fopen(getenv('LOGGER_OUTPUT') ?: 'php://stderr', 'wb');
+        fwrite($log, "Not retrying" . PHP_EOL);
+
+        return "";
```
grayside, line 37: nit: Are we unable to simply `return;`? The explicit string seems odd.

functions/tips_retry/test/DeployTest.php
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * Deploy the Cloud Function, called from DeploymentTrait::deployApp().
+     *
+     * Overrides CloudFunctionDeploymentTrait::doDeploy().
+     */
+    private static function doDeploy()
+    {
+        self::$projectId = self::requireEnv('GOOGLE_CLOUD_PROJECT');
+        self::$topicName = self::requireEnv('FUNCTIONS_TOPIC');
+        return self::$fn->deploy(['--retry' => ''], '--trigger-topic=' . self::$topicName);
```
grayside, line 145: nit: What is this `--retry` flag controlling? Could you add a comment?

functions/tips_retry/test/DeployTest.php
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private function publishMessage(): void
+    {
+        // Construct Pub/Sub message
+        $message = json_encode(['retry' => true]);
+
+        // Publish a message to the function.
+        $pubsub = new PubSubClient([
+            'projectId' => self::$projectId,
```
grayside, line 86: nit: Is passing the project ID explicitly still required for Pub/Sub client?

functions/tips_retry/test/DeployTest.php
```diff
@@ -0,0 +1,147 @@
...
+        $this->processFunctionLogs(self::$fn, $fiveMinAgo, function (\Iterator $logs) {
+            // Concatenate all relevant log messages.
+            $actual = '';
+            foreach ($logs as $log) {
+                $info = $log->info();
+                $actual .= $info['textPayload'];
+            }
+
+            // Check that multiple invocations of the function have occurred.
+            $retryCount = substr_count($actual, 'Retrying...');
```
grayside, line 74: This test cannot pass unless Pub/Sub tries, retries, and logs from both requests are processed by Cloud Logging. Is 5 seconds an adequate wait time?

functions/tips_retry/test/IntegrationTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+namespace Google\Cloud\Samples\Functions\TipsRetry\Test;
+
+use PHPUnit\Framework\TestCase;
+use Google\Cloud\TestUtils\CloudFunctionLocalTestTrait;
+
+/**
+ * Class IntegrationTest.
+ *
+ * Integration Test for helloGCS.
```
grayside, line 28: nit: Not a helloGCS test

functions/tips_retry/index.php
```diff
@@ -0,0 +1,40 @@
...
+// [START functions_tips_retry]
+
+use Google\CloudFunctions\CloudEvent;
+
+function tipsRetry(CloudEvent $event): string
+{
+    $data = $event->getData()['data'];
+    $data = json_decode(base64_decode($data), true);
+
+    $tryAgain = $data['retry'];
```
grayside, line 27: I think we need a comment here to clarify this "retry" property is an arbitrary value in the Pub/Sub message used to represent custom retry logic. I'd probably combine this with changing the property name to something like `yourCustomRetryPolicy`.

functions/tips_retry/index.php
```diff
@@ -0,0 +1,40 @@
...
+
+    if ($tryAgain) {
+        // Retry execution
```
grayside, line 30: nit: expand this comment e.g., "Thrown exceptions or 5xx status codes result in a retry."

grayside (changes requested):
Most of my comments are polish nits, with two blocking concerns on how we present the `retry` property and making sure the system test is sufficiently robust.

functions/tips_retry/index.php
ace-n: Nope - that causes a PHP error: `A function with return type must return a value`

functions/tips_retry/test/DeployTest.php
ace-n: Done - added a comment.

functions/tips_retry/test/DeployTest.php
ace-n: Akin to #1257, I increased this to 60 seconds.

functions/tips_retry/test/IntegrationTest.php
ace-n: Good catch - fixed.

functions/tips_retry/index.php
ace-n: Done.

functions/tips_retry/index.php
ace-n: Changed to `some_parameter` + added comments about how this works to the sample itself.

functions/tips_retry/test/DeployTest.php
ace-n: Apparently not - removed.

functions/tips_retry/test/DeployTest.php
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * Retrieve and process logs for the defined function.
+     *
+     * @param CloudFunction $fn function whose logs should be checked.
+     * @param string $startTime RFC3339 timestamp marking start of time range to retrieve.
+     * @param callable $process callback function to run on the logs.
+     */
+    private function processFunctionLogs(CloudFunction $fn, string $startTime, callable $process)
```
bshaffer, line 97: Seems like we should put this in a `FunctionsTestTrait` in `google/cloud-tools` at some point

bshaffer (approved):

functions/tips_retry/test/DeployTest.php
ace-n: See #1240

functions/tips_retry/composer.json
```diff
@@ -0,0 +1,9 @@
+{
+    "require": {
+        "google/cloud-functions-framework": "^0.7.1"
+    },
+    "require-dev": {
```
grayside, line 5: Did we want to add the start script to this sample?

functions/tips_retry/index.php
```diff
@@ -0,0 +1,49 @@
...
+    $tryAgain = $data['some_parameter'];
+
+    if ($tryAgain) {
+        /**
+         * Functions with automatic retries enabled should throw exceptions to
+         * indicate temporary failures that a retry might fix. In this case,
+         * a thrown exception will cause the original function invocation to be
+         * re-sent.
+         */
+        throw new Exception('Retrying...');
```
grayside, line 37: nit: While "Retrying..." here is the result of sending the exception, all the function knows is:

1. Something went wrong
2. We want a retry to happen.

I'm unsure if the goal of this snippet should be hyper-focused on training retry expectations, or if we should also include the angle of making the retry semantics friendly to troubleshooting. The exception ideally includes context on why the retry is needed so Cloud Logging and Cloud Error Reporting can provide more details:

```suggestion
        throw new Exception('Try Again triggered, expecting retry...');
```

functions/tips_retry/index.php
```diff
@@ -0,0 +1,49 @@
...
+    } else {
+        /**
+         * If a function with retries enabled encounters a non-retriable
+         * failure, it should return *without* throwing an exception.
+         */
+        $log = fopen(getenv('LOGGER_OUTPUT') ?: 'php://stderr', 'wb');
+        fwrite($log, "Not retrying" . PHP_EOL);
```
grayside, line 44: nit: if this would be out of alignment with other languages, it's fine, but the function doesn't control the retry, it only controls whether or not it's asking the system to perform a retry.

```suggestion
        fwrite($log, "Not requesting retry" . PHP_EOL);
```

functions/tips_retry/test/IntegrationTest.php
```diff
@@ -0,0 +1,113 @@
...
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider dataProvider
+     */
+    public function testTipsRetry(array $cloudevent, array $data, string $statusCode, string $expected, string $label): void
+    {
+        // Prepare the HTTP headers for a CloudEvent.
```
grayside, line 85: Note we extracted this code to a method in the helper trait, so all these sections in the sample tests should be updated:

https://github.com/GoogleCloudPlatform/php-tools/blob/a3bd49833bb21a41e67f6decba654edff0508e4c/src/TestUtils/CloudFunctionLocalTestTrait.php#L121

New sample tests should use that trait function.

grayside (approved):
All feedback is minor, but there are two lower-priority conventions not included:

1. Start script in composer.json
2. Use of trait method to facilitate integration test

I'm going to approve because it's unclear if these should be blocking considerations, but we should consider that lack of consistency on these seems like tech debt to me.

functions/tips_retry/index.php
ace-n: ACK - this is out of alignment, so I'll keep it as-is.

functions/tips_retry/test/IntegrationTest.php
ace-n: Done (for this sample).

Does this need to be refactored into existing samples? If so, mind filing an issue for that?

ace-n:
Added `start` script - @grayside can you review, and **merge this** if it looks good? Thanks!

functions/tips_retry/composer.json
ace-n: Done.

ace-n:
@bshaffer FYI - looks like the PHP 8.0 build isn't configured correctly.

grayside (approved):

PR #1259, opened 2021-01-16
chore(appengine): remove php55

ace-n:
🚨 **This may break docs staging + require hotfixes**, so we should coordinate the merge process with TWs. 🚨 

cc @llatif @qchong - _please LGTM._

Part of b/173545239


ace-n:
I believe most of those tags are pinned. (I filed [this issue](https://github.com/googleapis/repo-automation-bots/issues/1283) to add pinned-vs-unpinned logic to `snippet-bot`.)

grayside:
I believe some of this directory is used as source for community tutorials and materials in the cloud.google.com/php docs. @bshaffer should weigh in.

bshaffer (changes requested):
I do not believe any community articles exist for php 5.5, however there may be tutorials on cloudsite, and there certainly are MANY docs on cloudsite which reference these samples.

We **MUST** pin all those docs to a commit before we make this change.

Additionally, `php55` is referenced in this repo in the following places: 

 - `renovate.json`
 -  `appengine/standard/README.md`
 -  `appengine/wordpress/README.md`
 - `testing/run_test_suite.sh`

And note that the following samples still use `php55` outside of the `appengine/php55` directory, and need to be upgraded:

 - `auth/app-standard.yaml`
 - `pubsub/app/app.yaml`

ace-n:
FYI @bshaffer pinning CL (cl/353332231) has been submitted.

bshaffer:
@ace-n yes please, the first four need to be addressed as they're referencing the files removed in this PR. The last two are not required! Just thought I'd mention them.

bshaffer (approved):

PR #1270, opened 2021-01-28
feat!(appengine): php72 -> php folder

ace-n:
~**Do not merge** until we have the appropriate docs changes prepared.~ See cl/355045170


bshaffer (changes requested):
There are a LOT of instances of the path `appengine/php72/` in various sample `READMEs` and sample headers. We need to update them!

Examples:
 - https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/php72/storage/src/read_metadata.php#L21
 - https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/php72/auth/README.md#deploy-to-app-engine

ace-n:
@bshaffer good catch! Fixed those via `sed`.

appengine/standard/README.md
```diff
@@ -1,4 +1,4 @@
 ## App Engine Standard samples
-Samples are now in appengine/php72](../php72) for the PHP 7.2 version of App Engine Standard.
+Samples are now in appengine/php](../php72) for the PHP 7.2 version of App Engine Standard.
```
bshaffer, line 2: This also needs to be changed to `(../php)`

bshaffer (commented):
Do we think `appengine/php` is where these should go? Since we've removed `php55` samples, I would advocate `appengine/standard` is clearer and more informative.

appengine/standard/README.md
ace-n: This was replaced with the `7.2` readme, which (now) has correct links.

bshaffer (approved):
LGTM!

ace-n:
Updated `secrets` files and force-pushed changes, as requested by @bshaffer.

@grayside over to you. 🛩️

grayside (approved):

bshaffer:
Adding do not merge as this is waiting on an internal docs change

PR #1242, opened 2020-12-15
chore(appengine): enable deployment tests for php74

ace-n:
Addresses b/173545239

PR #1275, opened 2021-02-06
chore(testing): add ability to skip GAE Flex tests

ace-n:
Related to b/173545239 and #1242

### DIFF
--- a/.kokoro/common.cfg
+++ b/.kokoro/common.cfg
@@ -14,9 +14,3 @@ env_vars: {
     key: "TRAMPOLINE_BUILD_FILE"
     value: "github/php-docs-samples/.kokoro/system_tests.sh"
 }
-
-# Skip GAE Flexible deployment tests.
-env_vars: {
-    key: "RUN_APPENGINE_FLEX_TESTS"
-    value: "false"
-}
--- a/.kokoro/common.cfg
+++ b/.kokoro/common.cfg
@@ -14,3 +14,9 @@ env_vars: {
     key: "TRAMPOLINE_BUILD_FILE"
     value: "github/php-docs-samples/.kokoro/system_tests.sh"
 }
+
+# Skip GAE Flexible deployment tests.
+env_vars: {
+    key: "RUN_APPENGINE_FLEX_TESTS"
+    value: "false"
+}
--- a/appengine/flexible/datastore/test.sh
+++ b/appengine/flexible/datastore/test.sh
@@ -1,0 +1,8 @@
+export RUN_APPENGINE_FLEX_TESTS=false
+export DIR=$(pwd)
+
+if [ "$RUN_APPENGINE_FLEX_TESTS" != "true" ] &&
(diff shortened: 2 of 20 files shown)


.kokoro/common.cfg
```diff
@@ -14,3 +14,9 @@ env_vars: {
     key: "TRAMPOLINE_BUILD_FILE"
     value: "github/php-docs-samples/.kokoro/system_tests.sh"
 }
+
+# Skip GAE Flexible deployment tests.
+env_vars: {
+    key: "RUN_APPENGINE_FLEX_TESTS"
+    value: "false"
+}
```
bshaffer, line 22: We can remove this one!

appengine/flexible/datastore/test.sh
```diff
@@ -0,0 +1,8 @@
+export RUN_APPENGINE_FLEX_TESTS=false
+export DIR=$(pwd)
+
+if [ "$RUN_APPENGINE_FLEX_TESTS" != "true" ] &&
+       [[ -n $(grep -ri 'env:.+flex' app.*ml $DIR ) ]]; then
+        echo "Skipping tests in $DIR (no App Engine Flex tests)"
+        continue
+fi
```
bshaffer, line 8: Do we need this file? I don't think so.

bshaffer (changes requested):
Hmm, so we don't actually want to skip *all* flex tests. We just want to skip flex DEPLOYMENT tests. I think we could do this by:

1. Adding a check in each GAE flex deployment test to verify `RUN_FLEX_DEPLOYMENT_TESTS` should be `true`. This could be done by adding this to every `appengine/flexible/*/test/DeployTest.php` file:
    ```php
    /** @beforeClass */
    public static function checkProjectEnvVarBeforeClass()
    {
        self::requireEnv('RUN_FLEX_DEPLOYMENT_TESTS');
        self::checkProjectEnvVars();
    }
    ```

2. Add a block to `run_test_suite.sh` to temporarily set `RUN_DEPLOYMENT_TESTS` to false if the directory contains an `app.yaml` with `env: flex`. I can't really see a good/elegant way to do that with the current script though, so I prefer option 1

ace-n:
> Add a block to run_test_suite.sh to temporarily set RUN_DEPLOYMENT_TESTS to false if the directory contains an app.yaml with env: flex. I can't really see a good/elegant way to do that with the current script though, so I prefer option 1

We already do this - that's what the `grep` in `run_test_suite.sh` is for. 🙂

appengine/flexible/datastore/test.sh
ace-n: Good catch - reverted.

.kokoro/common.cfg
ace-n: Done.

bshaffer (approved):
Looks great! Should we skip the deployment test in endpoints/getting-started as well? I believe so!

PR #1291, opened 2021-02-25
chore: descriptive comment for path variable

b-loved-dreamer:
I am adding a new comment for the $path variable


texttospeech/src/synthesize_text_file.php
```diff
@@ -38,7 +38,7 @@
 use Google\Cloud\TextToSpeech\V1\VoiceSelectionParams;
 
 /** Uncomment and populate these variables in your code */
-// $path = 'Path to file to synthesize';
+// $path = 'textFile the text file to be synthesized. (e.g., hello.txt)';
```
bshaffer, line 41: please remove `textFile` from this, I think it's confusing:

```php
// $path = 'The text file to be synthesized. (e.g., hello.txt)';
```

bshaffer (approved):

PR #1293, opened 2021-03-01
samples(spanner): PITR samples backup fix

thiagotnunes:
Instead of using the earliest version time of the database, uses the current time (from spanner). If we used the earliest version time of the database instead we would be creating an empty backup, since the database we are backing up is not 1 hour old (default version retention period).

### DIFF
--- a/spanner/src/create_backup.php
+++ b/spanner/src/create_backup.php
@@ -44,7 +44,7 @@ function create_backup($instanceId, $databaseId, $backupId)
     $instance = $spanner->instance($instanceId);
     $database = $instance->database($databaseId);
 
-    $results = $database->execute("SELECT CURRENT_TIMESTAMP() as Timestamp");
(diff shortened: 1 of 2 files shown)


larkee (approved):
LGTM

spanner/src/create_backup.php
```diff
@@ -44,8 +44,11 @@ function create_backup($instanceId, $databaseId, $backupId)
     $instance = $spanner->instance($instanceId);
     $database = $instance->database($databaseId);
 
+    $results = $database->execute("SELECT TIMESTAMP_TRUNC(CURRENT_TIMESTAMP(), MICROSECOND) as Timestamp");
```
larkee, line 47: @jdpedrie The timestamp is being truncated because `DateTime` does not support nanosecond precision. Is there an alternative that does support nanosecond precision?

bshaffer (approved):

PR #1296, opened 2021-03-03
chore: replace deprecated silex with slim

bshaffer:


appengine/flexible/logging/phpunit.xml.dist
```diff
@@ -14,7 +14,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-<phpunit bootstrap="../../../testing/bootstrap.php">
+<phpunit bootstrap="../../../testing/bootstrap.php" convertWarningsToExceptions="false">
```
bshaffer, line 17: See https://github.com/googleapis/google-cloud-php/issues/3831

bshaffer: Also see https://github.com/googleapis/google-cloud-php/pull/3896

ace-n (approved):
Skimmed this; seems fine (at a high level) if tests pass.

appengine/flexible/websockets/index.php
```diff
@@ -1,44 +0,0 @@
-<?php
```
grayside, line 1: Did you mean to delete this file?

appengine/standard/getting-started/src/app.php
```diff
@@ -17,28 +17,27 @@
...
 // [END gae_php_app_storage_client_import]
+use Slim\Factory\AppFactory;
+use Slim\Views\Twig;
 
-$app = Slim\Factory\AppFactory::create();
-$app->addErrorMiddleware(true, true, true);
-
-// Get container
-$container = $app->getContainer();
+AppFactory::setContainer($container = new Container());
```
grayside, line 31: nit: Should Twig have a comment still since the other services being registered do?

appengine/standard/getting-started/src/app.php
```diff
@@ -17,28 +17,27 @@
...
+$container->set('view', function () {
+    return Twig::create(__DIR__ . '/../templates');
+});
 
-// Register Twig
-$container['view'] = function ($container) {
-    return new Slim\Views\Twig(__DIR__ . '/../templates');
-};
+$app = AppFactory::create();
+$app->addErrorMiddleware(true, true, true);
```
grayside, line 37: This looks like it's outside any region tags so probably ok, but I have no idea what this is toggling other than guessing it's turning on verbose & detailed error handling.

appengine/standard/getting-started/composer.json
```diff
@@ -2,15 +2,13 @@
   "require": {
     "google/cloud-storage": "^1.6",
     "slim/slim": "^4.0",
+    "slim/psr7": "^1.0",
     "slim/twig-view": "^3.0",
-    "slim/http": "^1.0",
-    "slim/psr7": "^1.0"
+    "php-di/slim-bridge": "^3.1"
```
grayside, line 7: Thanks for introducing me to php-di! I did some reading in https://php-di.org/doc/frameworks/slim.html to learn more about it, but when I check controllers.php I don't see what features we're using beyond Slim's built-in behavior. Am I missing something?

grayside (commented):
Spent 30 minutes with this PR and background reading, left a few comments but out of time for now.

appengine/flexible/websockets/index.php
bshaffer: Yes because the Silex action for `/` just prints `'index.html`. Without it, the server _should_ serve `index.html`... but I never actually tested it deploying to flex, only running locally. 

I can configure `nginx-app.conf` to serve `index.html`, or add an `index.php` that prints `index.html` w/o using Silex... either way, we don't need a microframework!

appengine/standard/getting-started/src/app.php
bshaffer: Yes, it's just to make sure errors are displayed. I'll add comments.

appengine/flexible/websockets/index.php
bshaffer: I just deployed to App engine to verify, and added the following to the README:

**Important**: After you deploy to App Engine Flexible, you will need to request the page `/index.html` directly to access the sample

appengine/standard/getting-started/src/app.php
bshaffer: done

appengine/standard/getting-started/composer.json
bshaffer: In this example, we use `$container->set` for both `storage` and `cloudsql` in `app.php` as a way to elegantly access those services in the controller layer (`controllers.php`)

grayside: Looks like that's also in https://www.slimframework.com/docs/v3/concepts/di.html?

PR #1298, opened 2021-03-04
fix(functions/http_content_type): PHP warning when header doesn't exist

bshaffer:


ace-n (approved):
LGTM.

functions/http_content_type/index.php
```diff
@@ -23,7 +23,7 @@ function helloContent(ServerRequestInterface $request): string
 {
     $name = 'World';
     $body = $request->getBody()->getContents();
-    switch ($request->getHeader('content-type')[0]) {
+    switch ($request->getHeaderLine('content-type')) {
```
grayside, line 26: Double-checked the [spec](https://www.php-fig.org/psr/psr-7/#headers-with-multiple-values), and for headers that don't accept comma-delimited values, we should use `getHeader()`.

A further problem, this header accepts further attributes (such as charset) after a semicolon.

Something like this:

```suggestion
    $header = $request->getHeader('content-type');
    $content_type = trim(explode(';', $header)[0]);
    switch ($content_type) {
```

grayside (changes requested):

functions/http_content_type/index.php
bshaffer: The `getHeaderLine` function is to provide convenience for working with headers as strings. It shouldn't be used with `Cookie-Set` because multiple headers with `Cookie-Set` can exist in a single request and `Cookie-Set` _cannot_ be concatenated with a comma.

However, `Content-Type` is not expected to have multiple headers sent in one request, so we can use `getHeaderLine` instead. In our example, if someone send in multiple `Content-Type` lines, the switch will not match. We could throw an exception under the `default` switch case if you'd prefer, but I do think `getHeaderLine` is acceptable to use here.

dwsupplee (approved):

functions/http_content_type/index.php
grayside: I'm fine going with that part. Should I file a follow-up issue for handling further attributes like charset?

grayside (approved):

functions/http_content_type/index.php
bshaffer: Sounds good to me!

PR #1304, opened 2021-03-13
RFC: feat(functions): PHP 7.4 preloading

bshaffer:
WIP for preloading in PHP 7.4 Cloud Functions

This demo suggests the users do the following:

1. Add the following to `composer.json`
    ```json
    {
        "config": {
            "optimize-autoloader": true
        },
    }
    ```

2. Add a `preload.php` file containing the following:
    ```php
    require_once __DIR__ . '/vendor/autoload.php';
    
    // By running "composer dump-autoload --optimize", the classmap file contains a
    // mapping of optimized class files.
    $composerClassmap = __DIR__ . '/vendor/composer/autoload_classmap.php';
    
    $classesToPreload = require $composerClassmap;
    
    foreach ($classesToPreload as $class => $file) {
        require_once $file;
    }
    ```

3. Add a `php.ini` containing the following:
    ```ini
    opcache.preload=preload.php
    ```

And that's it! Preloading will work for any/all classes handled by Composer's autoloader


bshaffer:
I am getting the following error when running the deployment:

```
PHP Fatal error:  Uncaught Error: Class 'PHPUnit\Framework\TestCase' not found in /workspace/test/IntegrationTest.php:28
Stack trace:
#0 /workspace/preload.php(12): require_once()
#1 {main}
  thrown in /workspace/test/IntegrationTest.php on line 28 
```

This is very strange, since `test/IntegrationTest.php` does not show up (locally) when I run the same composer commands.

I've tried ignoring `test` via `.gcloudignore`, and the error changes to `Failed opening required '/workspace/test/IntegrationTest.php'`, which isn't any better.

When I run `composer update` locally, I receive the following warnings:

```
→ composer update
Class Google\Cloud\Samples\Functions\TipsPreload\Test\IntegrationTest located in ./test/IntegrationTest.php does not comply with psr-4 autoloading standard. Skipping.
Class Google\Cloud\Samples\Functions\TipsPreload\Test\DeployTest located in ./test/DeployTest.php does not comply with psr-4 autoloading standard. Skipping.
```

Is it possible the composer version used by the server is adding these to the classmap instead of ignoring them, and that it's doing so before `.gcloudignore` is accounted for? Whatever the case, this is very strange behavior.

grayside:
Do we have any data on the benefits? E.g., A/B performance?

bshaffer:
@grayside no benchmarks, no, but we should definitely do some sanity-checking here. In theory, it should provide significant improvements on all requests, as the entire classmap is loaded into memory before the first request.

vishwarajanand:
Closing since we are not actively tracking this work. Maybe we'll want to come back to this if optimize flag proves optimistic.

bshaffer:
This is frustrating, since a lot of work was done to add this sample which could help the performance of Cloud Functions for a lot of our users. Maybe @grayside could suggest someone that might want to pick this up and add it to our docs?

rogerthatdev (approved):
Verified that this is deployable via the following:

`gcloud functions deploy php-http-function-w-preload --gen2 --runtime=php74 --region=us-central1 --source=. --entry-point=preloadDemo --trigger-http --allow-unauthenticated`

PR #1307, opened 2021-03-19
chore: removed unsupported samples

munkhuushmgl:
go/cloud-ai-2021-snippet-cleanup

Devsite:

cl/363930391 [pending]


bshaffer:
I understand removing product search, but removing all samples for speech seems a bit heavy handed.

Our [Speech Client](https://packagist.org/packages/google/cloud-speech) has 222K installs over the course of its lifetime. I do not think we should remove all its samples.

tmatsuo:
Nit: https://cloud.google.com/speech-to-text/docs/libraries still has `PHP` section in the Install section.

munkhuushmgl:
> Nit: https://cloud.google.com/speech-to-text/docs/libraries still has `PHP` section in the Install section.

I will fix that, thanks

munkhuushmgl:
> > Nit: https://cloud.google.com/speech-to-text/docs/libraries still has `PHP` section in the Install section.
> 
> I will fix that, thanks

Looks that will be used in the future, for tabs that I am going to add in deviste.
So i think we should leave  https://cloud.google.com/speech-to-text/docs/libraries as it is.

munkhuushmgl:
@tmatsuo PTAL

PR #1315, opened 2021-03-24
chore: (vision) removed unsupported samples

munkhuushmgl:
Devsite update: cl/363499751 [publlshed]


munkhuushmgl:
PTAL @bshaffer

bshaffer:
I do not agree that these samples should be removed so you will need to find someone else to approve and merge this

PR #1320, opened 2021-03-30
feat: add deploy group

bshaffer:


ace-n (approved):
Added some missing files ➡️ LGTM.

(I used `grep -L "@group deploy" **/*eploy*est.php` to find them.)

PR #1323, opened 2021-03-31
feat(Kokoro): add sub-directory support + split out lint build

ace-n:
- [x] Add support for `SUB_DIRECTORY` environment variable
- [x] Move `cs_check` into its own `lint` build.

@engelke @grayside FYI

(I'll leave further changes for subsequent PRs.)


bshaffer (approved):

.kokoro/system_tests.sh
```diff
@@ -64,5 +64,11 @@ fi
 # Install global test dependencies
 composer install -d testing/
 
+# cd into specific subdirectory (if appropriate)
+MAIN_DIR=$(pwd)
+if [[ -z "${SUB_DIRECTORY}" ]]; then
```
bshaffer, line 69: @ace-n kokoro tests are failing on [this line](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/testing/run_test_suite.sh#L79), which runs `git diff`.  It's because this bash test is incorrect, and as a result the script is running `cd` and ending up in the home directory.

Try this instead:

```suggestion
if [ ! -z "${SUB_DIRECTORY}" ]; then
```

PR #1333, opened 2021-04-05
fix: kokoro lint builds

bshaffer:
Simplifying the lint kokoro job

 - remove `RUN_CS_CHECK` env var
 - remove `lint.sh`


ace-n:
Are you directly invoking the lint script from Kokoro itself?

(I'm not sure if we should be removing the call to `run_cs_check.sh`.)

bshaffer:
@ace-n I am not removing the call to `run_cs_check.sh`, I'm just setting it to the trampoline script directly, as the file `lint.sh` was not really doing anything

ace-n (approved):
Ah, I missed that. Thanks!

LGTM once `lint` build passes.

PR #1334, opened 2021-04-05
fix: kokoro lint builds and lint script

bshaffer:


ace-n (approved):
LGTM once the presubmit check passes.

PR #1339, opened 2021-04-12
chore: add blunderbuss config to cloud iot label

gcseh:


bshaffer (changes requested):
We already have a file `.github/.blunderbuss.yaml` (notice this file in this PR is missing the dot prefix). I suggest using the existing file unless that one is improperly named, in which case please include the change of renaming the file as part of this PR!

bshaffer:
Looks like the existing file is incorrect (judging by every other repo with a `blunderbuss.yml`)

Fixed in https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1427

bshaffer (approved):

PR #1342, opened 2021-04-13
fix: functions retry sample and test

bshaffer:
- Fix the handling of pubsub data 
- Cleanup tests a tiny bit


functions/tips_retry/index.php
```diff
@@ -21,11 +21,13 @@
 
 function tipsRetry(CloudEvent $event): void
 {
-    $data = $event->getData()['data'];
-    $data = json_decode(base64_decode($data), true);
+    $cloudEventData = $event->getData();
+    $pubSubData = $cloudEventData['message']['data'];
+
+    $config = json_decode(base64_decode($pubSubData), true);
```
ace-n, line 27: Could we call this either `$retryConfig` or `$json`?

ace-n (changes requested):

functions/tips_retry/index.php
bshaffer: I like both of those better!

ace-n (approved):

PR #1345, opened 2021-04-13
fix: functions slack sample

bshaffer:
- fix timestamp - it was evaluating as empty because it was set to the string `'0'`
- Wrote a unit test for debugging and figured we might as well keep it
- misc test / function cleanup


ace-n (approved):
Thanks!

PR #1353, opened 2021-05-12
chore: cloud-compute samples preparation to incorporate them into cloud.google.com documentation

rsamborski:
Adding region tags for use on Google Cloud documentation pages + minor comments tweaks.
Updating cloud-compute library version to the newest stable one.


bshaffer (approved):

rsamborski:
@bshaffer I confirm that we are working in parallel on the documentation to use the aforementioned region tags

PR #1355, opened 2021-05-14
feat: wait for cloud-compute operation to finish

rsamborski:
Adding GCE sample to wait for operation to finish with tests

Please mind that tests for waiting operations (synchronous) can take up to couple minutes to finish:
Screenshot 2021-05-14 at 15 17 39

Therefore we may consider not running them every time. They are marked as group 'waitForOperation' for phpunit.


rsamborski:
To be merged after #1353

bshaffer (changes requested):
This is very helpful! However, adding a helper function like that is not how I'd like to handle it. Rather, we are going to add more complete LRO support to the library itself. In the meantime, I've added #1358 to show in the case of Instances how to wait for an operation to complete.

PR #1356, opened 2021-05-15
chore: update pubsub samples to new test runner

jdpedrie:


jdpedrie:
@bshaffer not sure how to fix the redeclared function error. Locally it seems like I can get it working by modifying [this line](https://github.com/GoogleCloudPlatform/php-tools/blob/master/src/TestUtils/TestTrait.php#L108) to use `require_once`, but I'm not sure if there's some context I'm missing.

bshaffer (approved):

PR #1358, opened 2021-05-19
add LRO to compute samples

bshaffer:


dwsupplee (approved):

rsamborski:
@bshaffer thanks for this example, unfortunately getStatus function returns an integer, so the code will never run. I've prepared a small fix and tested it on my environment here: #1361

PR #1363, opened 2021-05-23
feat: update region tags and add missing samples

jdpedrie:
cc @crwilcox 

This change updates region tags to match the correct locations, and renames files and functions to use the new tags. It also adds a couple of missing tags.

Additionally, I updated all the samples to use the new and improved runner and did a bit of refactoring of how data is stored in the database to lower potential flakiness.


bshaffer (approved):

samtstern:
This PR seems to have broken a lot of the Firestore docs. I see 10 unpublishable files with errors like this:

```
  docs/firestore/data-model.md: Exception: google3.devrel.tools.includecode.libraries.gerrit.GitRpcError: Error loading the GitHub content from: github/GoogleCloudPlatform/php-docs-samples/HEAD:firestore/src/document_ref.php Please verify you provided a valid content path. If you are trying to fetch GitHub content check that the repository is listed in the mirror https://github.git.corp.google.com. If you are trying to fetch gerrit content make sure that mdb/devrel-devsite-builder has read permissions on the repository using http://go/gob-ctl-acl The underlying exception was: /GitFS.FetchFileContent to [2002:abe:aec1:0:b029:1e7:285c:64da]:4504 : APP_ERROR(107)
```

samtstern:
Fixing the breakages here: http://cl/378831233

We need to be more careful going forward though.

bshaffer:
@samtstern good catch! Thanks for finding and fixing. 

The bot only says a bunch were added, so I checked for the one which was deleted and made the wrong assumption that was all we needed. It may be a bug in the bot because the files were renamed, they're labeled as added instead of changed or modified. I'll look into it.

PR #1366, opened 2021-05-25
fix: set a default firebase project if an alt project isnt used

bshaffer:
Adds the following logic in `secrets.sh` (no diff because it's binary and encrypted, but the following does not contain sensitive data:

```sh
# Firestore
if [ "$GOOGLE_PROJECT_ID" = "$GOOGLE_ALT_PROJECT_ID" ]; then
    # If an alt project hasn't been specified, we need to pick one because
    # there is no default Firestore project
    export FIRESTORE_PROJECT_ID=${GOOGLE_ALT_PROJECT_ID}1-fs
else
    export FIRESTORE_PROJECT_ID=${GOOGLE_ALT_PROJECT_ID}-fs
fi
```


jdpedrie (approved):

ace-n (approved):
LGTM if deployment tests pass.

PR #1379, opened 2021-06-11
chore: upgrade IOT samples to new format

bshaffer:


iot/README.md
```diff
@@ -28,44 +28,15 @@ IOT API from PHP. These samples are best seen in the context of the
 4.  **Install dependencies** via [Composer](http://getcomposer.org/doc/00-intro.md).
     Run `php composer.phar install` (if composer is installed locally) or `composer install`
     (if composer is installed globally).
-5.  Run `php iot.php`. The following commands are available:
```
laszlokorossy, line 31: Why did you remove the available commands? This list would be useful for reference.

iot/README.md
```diff
@@ -28,44 +28,15 @@ IOT API from PHP. These samples are best seen in the context of the
...
-    $ php iot.php create-registry my-registry my-pubsub-topic
-    Creating Registry
-    Id: my-registry, Name: projects/my-project/locations/us-central1/registries/my-registry
-    ```
+    @param string $projectId Google Cloud project ID
+    @param string $location (Optional) Google Cloud region
+```
 
 
 6. Run `php iot.php COMMAND --help` to print information about the usage of each command.
```
laszlokorossy, line 42: As the 'iot.php' file has been removed in this pull request, this command won't be valid anymore. Could you please update this part also and describe how it is possible to get information on each command?

iot/README.md
bshaffer: Because the commands no longer exist. The user executes one of the files. 

We could include a list of the files in src, but that's redundant and would become outdated and there's really no point.

iot/README.md
bshaffer: Good catch!! I'll remove this step!

the step above this shows how to get information on each command.

bshaffer: I added "Run them without arguments to print usage instructions" to make this more clear

laszlokorossy (approved):

gcseh (approved):

iot/README.md
laszlokorossy: Thanks, that is much better now.

PR #1383, opened 2021-06-14
feat: Compute Engine sample to present default values behaviour

rsamborski:
Compute Engine sample showcasing how to set usage reports bucket which presents the default values behaviour.


compute/cloud-client/instances/src/set_usage_export_bucket.php
```diff
@@ -0,0 +1,150 @@
...
+        // generated to have the default prefix value "usage_gce".
+        // ref: https://cloud.google.com/compute/docs/reference/rest/v1/projects/setUsageExportBucket
+        print("Setting report_name_prefix to empty value will cause the " .
+            "report to have the default value of `usage_gce`.");
+    }
+
+    // Set the usage export location.
+    $projectsClient = new ProjectsClient();
+    return $projectsClient->setUsageExportBucket($projectId, $usageExportLocation);
+}
```
bshaffer, line 78: To be consistent with the other samples in this repo, let's have only a single function in each file, and name the file the same name. So this needs to be three files, `src/set_usage_export_bucket.php`, `src/get_usage_export_bucket.php`, and `disable_usage_export_bucket.php`.

bshaffer (changes requested):

compute/cloud-client/instances/src/set_usage_export_bucket.php
rsamborski: I was thinking of how to do that and there is a slight problem with the split approach. We are trying to kill two birds with these samples: 

 1) Prepare individual samples (region tags: compute_usage_report_set, compute_usage_report_get, compute_usage_report_disable) for this page: https://cloud.google.com/compute/docs/logging/usage-export
No issue with splitting.

 2) Showcase the default value behavior for GCE GAPIC release (region tag: compute_instances_verify_default_value) which should present the set_usage_export_bucket and get_usage_export_bucket functions.
This can be split only in case one region tag can exist in two different files. Is this something that is possible?

Regardless to the answer above the disable function is not used in 2, so it can be moved to another file.

CODEOWNERS
```diff
@@ -30,3 +30,6 @@
 /texttospeech/            @bshaffer
 /vision/                  @bshaffer
 /video/                   @bshaffer
+
+# Compute samples owned by Remik
+/compute/cloud-client/
```
bshaffer, line 35: I believe you're missing the owning entity after ther directory

bshaffer (approved):
One small change otherwise this looks great

CODEOWNERS
rsamborski: Fixed

PR #1390, opened 2021-06-16
chore: ignore appengine/flexible dependencies

bshaffer:
@GoogleCloudPlatform/serverless-runtimes-php
@GoogleCloudPlatform/cdpe-serverless

Renovate-bot keeps churning out updates for these samples but I do not have the resources to maintain them. Also, the flex updates are bundled together with samples we DO want to update, which makes keeping other samples up-to-date more difficult.

I'd rather just ignore flex, but I'd like approval from the teams who maintain them before doing so, namely Serverless DPE.

See https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1360 and https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1310


ace-n:
Hmm - I can definitely empathize with the `renovate` concerns, but I don't think our team "officially" owns Flex samples. (The frustrating part: I don't know who does; they may in fact be orphaned.)

@iennae would know for sure.

PR #1394, opened 2021-06-17
feat(tests): use a single container with multiple php versions

bshaffer:
The goal here is to eventually break the tests down by sample directory instead of by PHP version. This will greatly simplify the complexity of our testing and speed up test execution.

### DIFF
--- a/.kokoro/php80.cfg
+++ b/.kokoro/php80.cfg
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
     key: "TRAMPOLINE_IMAGE"
-    value: "gcr.io/cloud-devrel-kokoro-resources/php80"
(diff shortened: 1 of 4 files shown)


ace-n (approved):
`phpbrew` seems nifty. ☕

.kokoro/php80.cfg
```diff
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
     key: "TRAMPOLINE_IMAGE"
-    value: "gcr.io/cloud-devrel-kokoro-resources/php80"
+    value: "gcr.io/cloud-devrel-kokoro-resources/php-multi-test"
```
vishwarajanand, line 6: I cannot find it here: https://github.com/googleapis/testing-infra-docker/blob/main/php/php80/Dockerfile

Is it located somewhere else?

PR #1425, opened 2021-06-23
feat(spanner): update query option samples with optimizerStatisticsPackage

larkee:


skuruppu (approved):

dwsupplee (approved):

spanner/src/create_client_with_query_options.php
```diff
@@ -41,7 +41,8 @@ function create_client_with_query_options($instanceId, $databaseId)
 {
     $spanner = new SpannerClient([
         'queryOptions' => [
-            'optimizerVersion' => "1"
+            'optimizerVersion' => '1',
+            'optimizerStatisticsPackage' => 'auto_20191128_14_47_22UTC'
```
bshaffer, line 45: This is a strange  value to have here (as opposed to `latest` below). Can we get a clarifying comment as to why it's set to the value it has?

spanner/src/query_data_with_query_options.php
```diff
@@ -47,7 +47,8 @@ function query_data_with_query_options($instanceId, $databaseId)
         'SELECT VenueId, VenueName, LastUpdateTime FROM Venues',
         [
             'queryOptions' => [
-                'optimizerVersion' => "1"
+                'optimizerVersion' => '1',
+                'optimizerStatisticsPackage' => 'latest'
```
bshaffer, line 51: Also here, even though the value makes no sense, I'd love a little clarify as to why.

bshaffer (changes requested):
I'm thinking that it's explained in the context of cloudsite, but I'd still appreciate a little context or clarification into the values provided

spanner/src/create_client_with_query_options.php
skuruppu: @larkee is on vacation so I tried to push a commit to their repo but didn't have the permissions.

If you're ok with it, I created #1438 to replace this PR.

PR #1438, opened 2021-07-01
feat(spanner): update query option samples with optimizerStatisticsPackage

skuruppu:
Original PR #1425.


bshaffer (approved):

bshaffer:
@skuruppu Looks great, thank you for making those changes! For future reference, it's better if you keep the updates in the previous PR!

PR #1442, opened 2021-07-02
chore: Updating php-cs-fixer config file to support php-cs-fixer 3.0

rsamborski:
Updating php-cs-fixer config file to support php-cs-fixer 3.0

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,4 +1,0 @@
-<?php
-
-print("Please update to php-cs-fixer 3.0" . PHP_EOL);
-exit(1);
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,6 +1,7 @@
 <?php
 
-return PhpCsFixer\Config::create()
+$config = new PhpCsFixer\Config();
+$config
     ->setRules([
         '@PSR2' => true,
         'concat_space' => ['spacing' => 'one'],
@@ -19,3 +20,5 @@ return PhpCsFixer\Config::create()
             ->in(__DIR__)
     )
 ;
+
+return $config;
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,24 +1,4 @@
 <?php
 
-$config = new PhpCsFixer\Config();
-$config
-    ->setRules([
-        '@PSR2' => true,
-        'concat_space' => ['spacing' => 'one'],
-        'no_unused_imports' => true,
-        'method_argument_space' => false,
-        'whitespace_after_comma_in_array' => true,
-        'method_argument_space' => [
-            'keep_multiple_spaces_after_comma' => true
-        ],
-        'return_type_declaration' => [
-            'space_before' => 'none'
-        ],
-    ])
-    ->setFinder(
-        PhpCsFixer\Finder::create()
-            ->in(__DIR__)
-    )
-;
-
-return $config;
+print("Please update to php-cs-fixer 3.0" . PHP_EOL);
+exit(1);
--- a/testing/run_cs_check.sh
+++ b/testing/run_cs_check.sh
@@ -15,6 +15,9 @@
 
 set -ex
 
+PROJECT_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/.."
+DIR="${1:-$PROJECT_ROOT}"
+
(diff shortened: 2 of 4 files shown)


bshaffer (commented):
Thanks for adding this!
I would prefer to be opinionated and support php-cs-fixer 3 only. We can upgrade our testing and composer to reflect this!

rsamborski:
@bshaffer please review again. I am enforcing php-cs-fixer 3.0. For Kokoro I had to extend testing/run_cs_check.sh script to add composer install. Moreover I am keeping the possibility of running php-cs-fixer from local folder first as some people might be using this approach. In case it's not needed we can get rid of lines 30-33.

Sorry for multiple commits, but I was experimenting a bit before finally decided to kick off a local docker for testing - I initially though changes will be more straightforward. I can squash commits at merge to keep the master branch cleaner.

.php_cs.dist
```diff
@@ -1,21 +1,4 @@
...
-            'space_before' => 'none'
-        ],
-    ])
-    ->setFinder(
-        PhpCsFixer\Finder::create()
-            ->in(__DIR__)
-    )
-;
+print("Please update to php-cs-fixer 3.0" . PHP_EOL);
+exit(1);
```
bshaffer, line 4: I'd prefer to just remove this file for simplicity. This is only used by us, so we don't need to add this layer of compatibility.

testing/run_cs_check.sh
```diff
@@ -15,15 +15,24 @@
...
 
+TESTING_DIR="$( dirname "${BASH_SOURCE[0]}")"
+PROJECT_ROOT="$( cd "${TESTING_DIR}" && pwd )/.."
+DIR="${1:-$PROJECT_ROOT}"
+
+# install local version of php-cs-fixer 3.0 from composer.json
+if [ -f "${TESTING_DIR}/composer.json" ]; then
+    # install composer dependencies
+    composer -q install --working-dir="${TESTING_DIR}"
+fi
```
bshaffer, line 26: This could be simplified to simply 

```
composer install -d testing/
```
(see [here](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/.kokoro/system_tests.sh#L60))
```

File: testing/run_cs_check.sh [hunk 3]
```diff
@@ -15,15 +15,24 @@
...
+
 # run php-cs-fixer
 PHP_CS_FIXER="php-cs-fixer"
 if [ -f "vendor/bin/php-cs-fixer" ]; then
     PHP_CS_FIXER="vendor/bin/php-cs-fixer"
 elif [ -f "./php-cs-fixer" ]; then
     PHP_CS_FIXER="./php-cs-fixer"
+elif [ -f "${TESTING_DIR}/vendor/bin/php-cs-fixer" ]; then
+    PHP_CS_FIXER="${TESTING_DIR}/vendor/bin/php-cs-fixer"
```
bshaffer, line 35: This is already handled by the first case (`if [ -f "vendor/bin/php-cs-fixer" ];`)

testing/run_cs_check.sh
rsamborski: Yes, assuming we are in the main folder. Looking at the code I had an impression that there are use cases when users run this script from another folder.

testing/run_cs_check.sh
rsamborski: I though so, but by default Kokoro runs from the root folder (php-docs-samples), so the first case doesn't work. My understanding was that there might be cases when this script is run for a folder inside the repository with some local vendor folder. If we don't have such use cases then I can simplify the whole script by just entering testing folder before execution.

.php_cs.dist
rsamborski: Ok

testing/run_cs_check.sh
bshaffer: This should always be run from the main folder. The checks below are for when `php-cs-fixer` is installed in different ways. That's probably overkill as well.

testing/run_cs_check.sh
bshaffer: I'm all for simplicity! These other checks probably don't need to be there. I'd prefer the script be opinionated. If we want to be extra friendly, we can print an error message like "this script should be run from the project root"

bshaffer (approved):

PR #1443, opened 2021-07-09
chore(tests): ensure test directories use the correct name

bshaffer:
These tests don't pass, but they were accidentally being skipped because the test directory was the wrong name.


bshaffer:
This PR requires https://github.com/googleapis/testing-infra-docker/pull/151 to pass

PR #1448, opened 2021-07-13
Chore: Updated Bigtable 'Create' samples to use the new format

saranshdhingra:


bigtable/src/create_dev_instance.php
```diff
@@ -25,81 +25,81 @@
 // Include Google Cloud dependencies using Composer
 require_once __DIR__ . '/../vendor/autoload.php';
```
bshaffer, line 26: Same as the other PR, let's remove all these!

bigtable/src/create_dev_instance.php
```diff
@@ -25,81 +25,81 @@
...
-            print('Error: ' . $operationResponse->getError()->getMessage());
+/**
+ * Create a development Bigtable instance
+ * @param string $projectId The Google Cloud project ID
+ * @param string $instanceId The ID of the Bigtable instance to be generated
+ * @param string $clusterId The ID of the cluster to be generated
+ * @param string $locationId The Bigtable region ID where you want your instance to reside
+ */
+function create_dev_instance($projectId, $instanceId, $clusterId, $locationId = 'us-east1-b')
+{
```
bshaffer, line 44: Same as the other PR, let's add typehints for everything!
```suggestion
function create_dev_instance(
    string $projectId,
    string $instanceId,
    string $clusterId,
    string $locationId = 'us-east1-b'
): void {
```

bshaffer:
@saranshdhingra we need to update the tests to use `runFunctionSnippet` but otherwise these all look great!

bshaffer (approved):

Issue #1389, opened 2021-06-16
Upgrade all samples to new sample format


bshaffer:
We have a new samples format that greatly reduces the complexity and overhead of maintaining samples by using reflection to invoke the sample method. We need to upgrade all samples to use this format.

**Background**
We have two existing formats, the `symfony/console` format and the "functionless" format.
1. The `symfony/console`format: This format uses the `symfony/console` component to create an executable command file (e.g. [`vision.php`](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/vision/vision.php)) which executes the samples. This requires the overhead of adding every sample to the command file, and adds a layer of abstraction for the users.
2. The "functionless" format: This format uses [inline PHP](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/bigtable/src/create_cluster.php#L32) to execute the sample. This has the advantage of making the sample files executable, but we lose the wrapping function, which is useful for consistency and documentation.

**The New Format**
The new format uses a method in [`testing/sample_helpers.php`](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/testing/sample_helpers.php#L7) to execute the function automatically from the CLI, and also to print usage instructions based on the PHP doc for the function. All that is required is [these two lines](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/asset/src/export_assets.php#L53) at the bottom of each sample file.

**Samples already upgraded**
 - [x] Asset (#1311)
 - [x] Compute
 - [x] Firestore (#1363, #1381)
 - [x] Pubsub (#1356)
 - [x] IAP (#1378)
 - [x] IOT (#1379)
 - [x] Spanner (#1234)
 - [x] Auth (#1375)
 - [x] Storage (#1384)
 - [x] Bigtable (#1496)
 - [x] DLP (#1615)
 - [x] [TextToSpeech](texttospeech) (https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1641)
 - [x] [Video](video) (9 samples) (https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1639)
 - [x] [SecretManager](secretmanager) (15 samples) (https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1647)
 - [x] [Language](language) (13 samples) (https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1650)
[...]

PR #1449, opened 2021-07-14
chore: Changed samples in Bigtable to new format for the 'list samples' - issue #1389

saranshdhingra:


bigtable/src/list_instance.php
```diff
@@ -25,30 +25,30 @@
 // Include Google Cloud dependencies using Composer
 require_once __DIR__ . '/../vendor/autoload.php';
```
bshaffer, line 26: As we discussed, we should remove all these!

bigtable/src/list_instance.php
```diff
@@ -25,30 +25,30 @@
...
-/** Uncomment and populate these variables in your code */
-// $projectId = 'The Google project ID';
-// $instanceId = 'The Bigtable instance ID';
-
-$instanceAdminClient = new BigtableInstanceAdminClient();
+/**
+ * List Bigtable instances in a project
+ * @param string $projectId The Google Cloud project ID
+ */
+function list_instance($projectId)
```
bshaffer, line 35: Let's add typehints for all these
```suggestion
function list_instance(string $projectId): void
```

bshaffer (commented):
Really great work! Just a few suggestions.

bigtable/test/bigtableTest.php
```diff
@@ -109,8 +109,7 @@ public function testCreateDevInstance()
     public function testListInstances()
     {
         $content = self::runSnippet('list_instance', [
-            self::$projectId,
-            self::$instanceId
+            self::$projectId
```
saranshdhingra, line 112: Just wanted to highlight the removal of this as the instanceId wasn't being used by the sample anyway.

bshaffer (approved):

PR #1450, opened 2021-07-15
chore: Bigtable new sample format for write and delete samples

saranshdhingra:


bigtable/test/writeTest.php
```diff
@@ -46,7 +46,7 @@ public static function tearDownAfterClass(): void
 
     public function testWriteSimple()
     {
-        $output = $this->runSnippet('writes/write_simple', [
+        $output = $this->runSnippet('write_simple', [
```
bshaffer, line 49: All of these should be changed to `runFunctionSnippet`
```suggestion
        $output = $this->runFunctionSnippet('write_simple', [
```

bigtable/src/delete_cluster.php
```diff
@@ -22,39 +22,38 @@
...
+/**
+ * Delete a cluster
+ * @param string $projectId The Google Cloud project ID
+ * @param string $instanceId The ID of the Bigtable instance
+ * @param string $clusterId The ID of the cluster to be deleted
+ */
+function delete_cluster(
+    string $projectId,
+    string $instanceId,
+    string  $clusterId
```
bshaffer, line 38:
```suggestion
    string $clusterId
```

saranshdhingra: Somehow `php-cs-fixer` didn't fix this 🤔 . Good catch!

bigtable/test/BigtableTestTrait.php
```diff
@@ -96,7 +96,7 @@ public static function deleteBigtableInstance()
         self::$instanceAdminClient->deleteInstance($instanceName);
     }
 
-    private static function runSnippet($sampleName, $params = [])
+    private static function runFunctionSnippet($sampleName, $params = [])
```
bshaffer, line 99: There is a function `runFunctionSnippet` built into `TestTrait` which should be handling this.
```suggestion
    private static function runSnippet($sampleName, $params = [])
```

bshaffer (approved):

bshaffer (approved):

PR #1451, opened 2021-07-15
chore: Bigtable read_snippets.php split to individual samples in the new format

saranshdhingra:


bigtable/src/read_row.php
```diff
@@ -0,0 +1,77 @@
...
+    $table = $dataClient->table($instanceId, $tableId);
+    
+    $rowkey = "phone#4c410523#20190501";
+    $row = $table->readRow($rowkey);
+
+    print_row($rowkey, $row);
+}
+// [END bigtable_reads_row]
+
+// [START bigtable_reads_print]
```
saranshdhingra, line 52: Please note this is used as a helper and doesn't have it's own sample file.

bigtable/src/read_prefix.php
```diff
@@ -0,0 +1,89 @@
...
+    $end = $prefix;
+    $end[-1] = chr(
+        ord($end[-1]) + 1
+    );
+
+    $rows = $table->readRows([
+        'rowRanges' => [
+            [
+                'startKeyClosed' => $prefix,
+                'endKeyClosed' => $end,
```
saranshdhingra, line 55: I think this is buggy, will analyse and add an issue. 
If I use `phone#4c410523#20190501` as the prefix, then even `phone#4c410523#20190502` (returned 2 rows).
Simply using endKeyOpen did fix it, but I will look into this more.

bigtable/src/read_row_partial.php
```diff
@@ -0,0 +1,77 @@
...
+    string $tableId
+): void {
+    // Connect to an existing table with an existing instance.
+    $dataClient = new BigtableClient([
+        'projectId' => $projectId,
+    ]);
+    $table = $dataClient->table($instanceId, $tableId);
+
+    $rowkey = "phone#4c410523#20190501";
+    $rowFilter = Filter::qualifier()->regex("os_build");
```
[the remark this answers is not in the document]
saranshdhingra: Should we add a rule in cs fixer regarding this?

bigtable/src/read_row_range.php
```diff
@@ -0,0 +1,83 @@
...
+        print_row($key, $row);
+    }
+}
+// [END bigtable_reads_row_range]
+
+// Helper function for printing the row data
+function print_row($key, $row)
+{
+    printf('Reading data for row %s' . PHP_EOL, $key);
+    foreach ((array)$row as $family => $cols) {
```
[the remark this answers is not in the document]
saranshdhingra: Not sure why the cs fixer didn't cover this 🤔

bigtable/src/read_row_partial.php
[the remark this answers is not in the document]
bshaffer: That's a great idea!

.php-cs-fixer.dist.php
```diff
@@ -15,7 +15,7 @@
         ],
         'return_type_declaration' => [
             'space_before' => 'none'
-        ],
+        ]
```
bshaffer, line 18: Nit: revert unneeded change
```suggestion
        ],
```

bigtable/src/read_prefix.php
```diff
@@ -0,0 +1,91 @@
...
+    $dataClient = new BigtableClient([
+        'projectId' => $projectId,
+    ]);
+    $table = $dataClient->table($instanceId, $tableId);
+
+    $prefix = 'phone#';
+    $end = $prefix;
+    $end[-1] = chr(
+        ord($end[-1]) + 1
+    );
```
bshaffer, line 51: This block doesn't look right to me. What is it doing, and why? A few suggestions:

1. add a comment explaining what's happening and why this is the best way to do it
2. find a clearer way to accomplish this
3. use hardcoded characters instead of variables

bshaffer: After reading [the docs](https://cloud.google.com/bigtable/docs/reading-data#ruby_5), this is there to establish the range of the prefix. By incrementing the last character, the filter will match everything between the two, essentially only matching the prefix. So we should add a comment here to explain this, something like:

```
Increment the last character of the prefix so the filter matches everything in between
```

PR #1456, opened 2021-07-22
samples: Bigtable: Added App Profile samples and tests

saranshdhingra:
The tests should pass once https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1450 is merged.


bigtable/src/create_app_profile.php
```diff
@@ -0,0 +1,86 @@
...
+    $appProfile = new AppProfile([
+        'name'=>$appProfileId,
+        'description'=>'Description for this newly created AppProfile'
+    ]);
+
+    // create a new routing policy
+    // allow_transactional_writes refers to Single-Row-Transactions(https://cloud.google.com/bigtable/docs/app-profiles#single-row-transactions)
+    $routingPolicy = new SingleClusterRouting([
+        'cluster_id'=>$clusterId,
+        'allow_transactional_writes'=>false
```
bshaffer, line 58: same here
```suggestion
        'cluster_id' => $clusterId,
        'allow_transactional_writes' => false
```

bigtable/src/create_app_profile.php
```diff
@@ -0,0 +1,86 @@
...
+    // $routingPolicy = new \Google\Cloud\Bigtable\Admin\V2\AppProfile\MultiClusterRoutingUseAny();
+    // $appProfile->setMultiClusterRoutingUseAny($routingPolicy);
+
+    printf("Creating a new AppProfile %s" . PHP_EOL, $appProfileId);
+
+    try {
+        $newAppProfile = $instanceAdminClient->createAppProfile($instanceName, $appProfileId, $appProfile);
+    } catch (ApiException $e) {
+        if ($e->getStatus() === 'ALREADY_EXISTS') {
+            printf("AppProfile %s already exists.", $appProfileId);
```
bshaffer, line 74: This will fall through without a return. 
```suggestion
            return printf("AppProfile %s already exists.", $appProfileId);
```

bigtable/src/create_app_profile.php
```diff
@@ -0,0 +1,86 @@
...
+        } else {
+            throw $e;
+        }
```
bshaffer, line 77: Since we return above, we should use implicit else here:
```suggestion
        }
        throw $e;
```

bigtable/src/delete_app_profile.php
```diff
@@ -0,0 +1,65 @@
...
+        // If $ignoreWarnings is set to false, Bigtable will warn you that all future requests using the AppProfile will fail
+        $instanceAdminClient->deleteAppProfile($appProfileName, $ignoreWarnings);
+        printf("App Profile %s deleted." . PHP_EOL, $appProfileId);
+    } catch (ApiException $e) {
+        if ($e->getStatus() === 'NOT_FOUND') {
+            printf("App Profile %s does not exist." . PHP_EOL, $appProfileId);
+            return;
+        } else {
+            throw $e;
+        }
```
bshaffer, line 58: Implicit else
```suggestion
        }
        throw $e;
```

bigtable/src/get_app_profile.php
```diff
@@ -0,0 +1,74 @@
...
+    printf("Fetching the App Profile %s" . PHP_EOL, $appProfileId);
+    try {
+        $appProfile = $instanceAdminClient->getAppProfile($appProfileName);
+    } catch (ApiException $e) {
+        if ($e->getStatus() === 'NOT_FOUND') {
+            printf("App profile %s does not exist." . PHP_EOL, $appProfileId);
+            return;
+        } else {
+            throw $e;
+        }
```
bshaffer, line 54:
```suggestion
        }
        throw $e;
```

bigtable/src/get_app_profile.php
```diff
@@ -0,0 +1,74 @@
...
+    // Fetch some commonly used metadata
+    echo "Name: " . $appProfile->getName() . PHP_EOL;
+    echo "Etag: " . $appProfile->getEtag() . PHP_EOL;
+    echo "Description: " . $appProfile->getDescription() . PHP_EOL;
+    echo "Routing Policy: " . $appProfile->getRoutingPolicy() . PHP_EOL;
+
+    if ($appProfile->hasSingleClusterRouting()) {
+        echo "Cluster: " . $appProfile->getSingleClusterRouting()->getClusterId() . PHP_EOL;
+        echo "Single-Row Transactions: " . ($appProfile->getSingleClusterRouting()->getAllowTransactionalWrites() ? "Yes" : "No") . PHP_EOL;
+    }
```
bshaffer, line 68: All of these double quotes should be converted to single quotes

bigtable/src/list_app_profiles.php
```diff
@@ -0,0 +1,65 @@
...
+        foreach ($appProfiles->iterateAllElements() as $profile) {
+            // You can fetch any AppProfile metadata from the $profile object(see get_app_profile.php)
+            echo "Name: " . $profile->getName() . PHP_EOL;
+        }
+    } catch (ApiException $e) {
+        if ($e->getStatus() === 'NOT_FOUND') {
+            printf("Instance %s does not exist." . PHP_EOL, $instanceId);
+            return;
+        } else {
+            throw $e;
```
bshaffer, line 57: implicit else

bshaffer (commented):
The main issues I see in this PR are mostly nits, but please check for:

1. Using double quotes when you should have single quotes
2. Lack of implicit else
3. spaces needed around `=>`
4. in `try/catch` make sure the special cases return instead of executing the rest of the sample.

bigtable/src/update_app_profile.php
```diff
@@ -0,0 +1,105 @@
...
+
+    // set the newly created routing policy to our app profile
+    $appProfile->setSingleClusterRouting($routingPolicy);
+
+    // or we could also create a multi cluster routing policy like so:
+    // $routingPolicy = new \Google\Cloud\Bigtable\Admin\V2\AppProfile\MultiClusterRoutingUseAny();
+    // $appProfile->setMultiClusterRoutingUseAny($routingPolicy);
+
+    // returns a string identifier depending on SingleClusterRouting or MultiClusterRoutingUseAny
+    $routingPolicyStr = $appProfile->getRoutingPolicy();
```
saranshdhingra, line 70: This returns `single_cluster_routing` or 'multi_cluster_routing_any`. But TBH, it feels like the field that the user should update is `routing_policy`.

bshaffer: If you take a look at the [API Reference](https://cloud.google.com/bigtable/docs/reference/admin/rpc/google.bigtable.admin.v2#google.bigtable.admin.v2.AppProfile), you'll see there are two fields as part of the `AppProfile` object being updated: `single_cluster_routing` and `multi_cluster_routing_any`. `routing_policy`, even though it's a protobuf getter, is not the field to be updated, it's only the field identifier. So this code is correct.

bshaffer (approved):

PR #1422, opened 2021-06-22
chore: Add a comment over 2 lines in the new snippet format to prevent copy.

saranshdhingra:
Let me know if the text needs a change.

PR #1457, opened 2021-07-29
feat(spanner): add samples for CMMR

larkee:
This PR adds samples for Customer Managed Multi Region instance configurations.

### DIFF
--- a/spanner/src/create_database_with_default_leader.php
+++ b/spanner/src/create_database_with_default_leader.php
@@ -72,5 +72,6 @@ function create_database_with_default_leader($instanceId, $databaseId, $defaultL
 }
 // [END spanner_create_database_with_default_leader]
 
+// The following 2 lines are only needed to run the samples
 require_once __DIR__ . '/../../testing/sample_helpers.php';
 \Google\Cloud\Samples\execute_sample(__FILE__, __NAMESPACE__, $argv);
--- a/spanner/src/create_database_with_default_leader.php
+++ b/spanner/src/create_database_with_default_leader.php
@@ -1,0 +1,76 @@
+<?php
+/**
+ * Copyright 2021 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/spanner/README.md
+ */
+
+namespace Google\Cloud\Samples\Spanner;
+
+// [START spanner_create_database_with_default_leader]
+use Google\Cloud\Spanner\SpannerClient;
+
+/**
+ * Creates a database with a default leader.
+ * Example:
+ * ```
+ * create_database_with_default_leader($instanceId, $databaseId, $defaultLeader);
+ * ```
+ *
+ * @param string $instanceId The Spanner instance ID.
+ * @param string $databaseId The Spanner database ID.
+ * @param string $defaultLeader The leader instance configuration used by default.
+ */
+function create_database_with_default_leader($instanceId, $databaseId, $defaultLeader)
+{
+    $spanner = new SpannerClient();
+    $instance = $spanner->instance($instanceId);
+
+    if (!$instance->exists()) {
+        throw new \LogicException("Instance $instanceId does not exist");
+    }
+
+    $operation = $instance->createDatabase($databaseId, ['statements' => [
+        "CREATE TABLE Singers (
+            SingerId     INT64 NOT NULL,
+            FirstName    STRING(1024),
+            LastName     STRING(1024),
+            SingerInfo   BYTES(MAX)
+        ) PRIMARY KEY (SingerId)",
+        "CREATE TABLE Albums (
+            SingerId     INT64 NOT NULL,
+            AlbumId      INT64 NOT NULL,
+            AlbumTitle   STRING(MAX)
+        ) PRIMARY KEY (SingerId, AlbumId),
+        INTERLEAVE IN PARENT Singers ON DELETE CASCADE",
+        "ALTER DATABASE `$databaseId` SET OPTIONS (
+        version_retention_period = '$defaultLeader')"
+    ]]);
+
+    print('Waiting for operation to complete...' . PHP_EOL);
+    $operation->pollUntilComplete();
+
+    $database = $instance->database($databaseId);
+    printf('Created database %s on instance %s with default leader %s' . PHP_EOL,
+        $databaseId, $instanceId, $database->info()['defaultLeader']);
+}
+// [END spanner_create_database_with_default_leader]
+
+require_once __DIR__ . '/../../testing/sample_helpers.php';
+\Google\Cloud\Samples\execute_sample(__FILE__, __NAMESPACE__, $argv);
--- a/spanner/src/create_database_with_default_leader.php
+++ b/spanner/src/create_database_with_default_leader.php
@@ -60,7 +60,7 @@ function create_database_with_default_leader($instanceId, $databaseId, $defaultL
         ) PRIMARY KEY (SingerId, AlbumId),
         INTERLEAVE IN PARENT Singers ON DELETE CASCADE",
         "ALTER DATABASE `$databaseId` SET OPTIONS (
-        version_retention_period = '$defaultLeader')"
+        default_leader = '$defaultLeader')"
     ]]);
 
     print('Waiting for operation to complete...' . PHP_EOL);
--- a/spanner/src/get_database_ddl.php
+++ b/spanner/src/get_database_ddl.php
@@ -27,7 +27,7 @@ namespace Google\Cloud\Samples\Spanner;
 use Google\Cloud\Spanner\SpannerClient;
 
 /**
- * Updates the default leader of the database.
+ * Gets the database DDL statements.
  * Example:
  * ```
  * get_database_ddl($instanceId, $databaseId);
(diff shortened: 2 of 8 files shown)


larkee:
Test failures are unrelated and will be fixed by #1459

spanner/src/create_database_with_default_leader.php
```diff
@@ -0,0 +1,76 @@
...
+            SingerInfo   BYTES(MAX)
+        ) PRIMARY KEY (SingerId)",
+        "CREATE TABLE Albums (
+            SingerId     INT64 NOT NULL,
+            AlbumId      INT64 NOT NULL,
+            AlbumTitle   STRING(MAX)
+        ) PRIMARY KEY (SingerId, AlbumId),
+        INTERLEAVE IN PARENT Singers ON DELETE CASCADE",
+        "ALTER DATABASE `$databaseId` SET OPTIONS (
+        version_retention_period = '$defaultLeader')"
```
devbww, line 63: s/version_retention_period/default_leader/

larkee: Good catch! Fixed :+1:

spanner/test/spannerTest.php
```diff
@@ -728,6 +758,79 @@ public function testCreateClientWithQueryOptions()
...
+     * @depends testCreateDatabaseWithDefaultLeader
+     */
+    private function testUpdateDatabaseWithDefaultLeader()
+    {
+        $output = $this->runFunctionSnippet('create_database_with_default_leader', [
+            'instance_id' => self::$multiInstanceId,
+            'database_id' => self::$multiDatabaseId,
+            'defaultLeader' => self::$updatedDefaultLeader
+        ]);
+        $this->assertStringContainsString(self::$defaultLeader, $output);
```
hengfengli, line 807: Shouldn't this be `self::$updatedDefaultLeader`?

spanner/test/spannerTest.php
```diff
@@ -728,6 +758,79 @@ public function testCreateClientWithQueryOptions()
...
+    }
+
+    /**
+     * @depends testCreateDatabaseWithDefaultLeader
+     */
+    private function testUpdateDatabaseWithDefaultLeader()
+    {
+        $output = $this->runFunctionSnippet('create_database_with_default_leader', [
```
hengfengli, line 802: The snippet does not match the test name.

spanner/test/spannerTest.php
```diff
@@ -82,6 +100,18 @@ public static function setUpBeforeClass(): void
         self::$encryptedDatabaseId = 'en-test-' . time() . rand();
         self::$backupId = 'backup-' . self::$databaseId;
         self::$instance = $spanner->instance(self::$instanceId);
+
+        self::$multiInstanceId = 'test-' . time() . rand() . 'm';
+        self::$multiDatabaseId = 'test-' . time() . rand() . 'm';
+        self::$instanceConfig = 'nam3';
+        self::$defaultLeader = 'us-east4';
```
hengfengli, line 107: Maybe change this to `us-central1` to keep `defaultLeader` and `updatedDefaultLeader` being different?

hengfengli (approved):
LGTM. Some small issues.

spanner/test/spannerTest.php
larkee: Correct!

spanner/test/spannerTest.php
larkee: Whoops! :sweat_smile:

spanner/test/spannerTest.php
larkee: Good catch!

hengfengli:
Hi @dwsupplee , can we get an approval from you? Thanks.

spanner/src/create_database_with_default_leader.php
```diff
@@ -0,0 +1,76 @@
...
+    print('Waiting for operation to complete...' . PHP_EOL);
+    $operation->pollUntilComplete();
+
+    $database = $instance->database($databaseId);
+    printf('Created database %s on instance %s with default leader %s' . PHP_EOL,
+        $databaseId, $instanceId, $database->info()['defaultLeader']);
+}
+// [END spanner_create_database_with_default_leader]
+
+require_once __DIR__ . '/../../testing/sample_helpers.php';
```
dwsupplee, line 75: https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1422 introduced adding `// The following 2 lines are only needed to run the samples` prior to these lines for clarification - would you be open to following the same pattern here?

spanner/src/get_database_ddl.php
```diff
@@ -0,0 +1,53 @@
...
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/spanner/README.md
+ */
+
+namespace Google\Cloud\Samples\Spanner;
+
+// [START spanner_get_database_ddl]
+use Google\Cloud\Spanner\SpannerClient;
+
+/**
+ * Updates the default leader of the database.
```
dwsupplee, line 30: This description is incorrect.

spanner/src/get_instance_config.php
```diff
@@ -0,0 +1,45 @@
...
+
+// [START spanner_get_instance_config]
+use Google\Cloud\Spanner\SpannerClient;
+
+/**
+ * Gets the leader options for the instance configuration.
+ *
+ * @param string $instanceConfig The name of the instance configuration.
+ */
+function get_instance_config($instanceConfig)
```
dwsupplee, line 34: The body of the method gets the leader options, which seems like a subset of the instance config. What are your thoughts on updating this to be something like `get_instance_config_leader_options` , or more broadly just printing out all the instance config options?

spanner/src/query_information_schema_database_options.php
```diff
@@ -0,0 +1,62 @@
...
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/spanner/README.md
+ */
+
+namespace Google\Cloud\Samples\Spanner;
+
+// [START spanner_query_information_schema_database_options]
+use Google\Cloud\Spanner\SpannerClient;
+
+/**
+ * Updates the default leader of the database.
```
dwsupplee, line 30: This is incorrect.

spanner/src/query_information_schema_database_options.php
```diff
@@ -0,0 +1,62 @@
...
+        "SELECT s.OPTION_NAME, s.OPTION_VALUE
+        FROM INFORMATION_SCHEMA.DATABASE_OPTIONS s
+        WHERE s.OPTION_NAME = 'default_leader'");
+
+    foreach ($results as $row) {
+        $optionName = $row['OPTION_NAME'];
+        $optionValue = $row['OPTION_VALUE'];
+        printf("The $optionName for $databaseId is $optionValue" . PHP_EOL);
+    }
+    if (!$results->stats()['rowCountExact']) {
```
dwsupplee, line 55: I don't have a strong knowledge of the underlying call here, but just wanted to confirm - does `rowCountExact` equate to not having an option for `default_leader`?

spanner/src/query_information_schema_database_options.php
```diff
@@ -0,0 +1,62 @@
...
+function query_information_schema_database_options($instanceId, $databaseId)
+{
+    $spanner = new SpannerClient();
+    $instance = $spanner->instance($instanceId);
+    $database = $instance->database($databaseId);
+
+    $results = $database->execute(
+        "SELECT s.OPTION_NAME, s.OPTION_VALUE
+        FROM INFORMATION_SCHEMA.DATABASE_OPTIONS s
+        WHERE s.OPTION_NAME = 'default_leader'");
```
dwsupplee, line 48: I trust your judgement on what you'd like to do here, but most PHP style guides prefer the closing bracket on a newline: 

```
$results = $database->execute(
    "SELECT s.OPTION_NAME, s.OPTION_VALUE
    FROM INFORMATION_SCHEMA.DATABASE_OPTIONS s
    WHERE s.OPTION_NAME = 'default_leader'"
);
```

spanner/src/query_information_schema_database_options.php
```diff
@@ -0,0 +1,62 @@
...
+
+    foreach ($results as $row) {
+        $optionName = $row['OPTION_NAME'];
```
dwsupplee, line 51: Will these options always be set?

spanner/test/spannerTest.php
```diff
@@ -51,6 +51,24 @@ class spannerTest extends TestCase
     /** @var $instance Instance */
     protected static $instance;
 
+    /** @var string multiInstanceId */
+    protected static $multiInstanceId;
+
+    /** @var $multiInstance Instance */
```
dwsupplee, line 57:
```suggestion
    /** @var Instance $multiInstance */
```

dwsupplee (changes requested):

spanner/src/get_instance_config.php
dwsupplee, line 34: The body of the method gets the leader options, which seems like a subset of the instance config. What are your thoughts on updating this to be something like get_instance_config_leader_options , or more broadly just printing out all the instance config options?

dwsupplee (changes requested):

spanner/src/get_instance_config.php
hengfengli: We prefer to use the current tag. The tag is same for samples across all languages. This is intended to make it more generic to avoid too many similar samples. Currently, we only need to print out the default leader. We may get more requests in the future to print out other fields.

spanner/src/query_information_schema_database_options.php
hengfengli: Yes, as in the SQL query, we specified `s.OPTION_NAME` and `s.OPTION_VALUE`. If the `default_leader` option is not set, then it will not return any rows in `$results`. I don't think this would cause a key error.

spanner/src/create_database_with_default_leader.php
larkee: Done :)

spanner/src/query_information_schema_database_options.php
larkee: Yes. If no rows are returned then the `default_leader` must not be set.

spanner/src/query_information_schema_database_options.php
larkee: Good catch! Updated :+1:

spanner/src/query_information_schema_database_options.php
larkee: Whoops! Fixed :+1:

spanner/test/spannerTest.php
larkee: Done :+1:

hengfengli:
@dwsupplee Could you please take another look? Thanks.

dwsupplee (approved):

PR #1467, opened 2021-08-20
feat(kms): Add samples for new KMS RNG APIs

sethvargo:


kms/src/create_key_mac.php
```diff
@@ -0,0 +1,67 @@
...
+
+if (isset($argv)) {
+    if (count($argv) === 0) {
+        return printf("Usage: php %s PROJECT_ID LOCATION_ID KEY_RING_ID ID\n", basename(__FILE__));
+    }
+
+    require_once __DIR__ . '/../vendor/autoload.php';
+    list($_, $projectId, $locationId, $keyRingId, $id) = $argv;
+    create_key_mac_sample($projectId, $locationId, $keyRingId, $id);
+}
```
bshaffer, line 67: We now have a helper for this. It looks like this:

```php
// The following 2 lines are used to execute this sample on the CLI
require_once __DIR__ . '/../../testing/sample_helpers.php';
\Google\Cloud\Samples\execute_sample(__FILE__, __NAMESPACE__, $argv);
```

If you feel up to it, you can migrate these to use this format instead. If not, we can do it in a separate PR. I'm ambivalent.

sethvargo: If it's okay with you, I'll do all of them at once in the next PR?

sethvargo:
@bshaffer heads up - I don't have permissions to merge on this repo.

bshaffer (approved):

Issue #1465, opened 2021-08-12
Update cs fixer rules


saranshdhingra:
Add the cs fixer rules to add the following:

- [x] Change double quotes to single quotes when no escape characters or variables are used within the string.
- [x] Single space after cast, ex: `(int)$var` to become `(int) $var`
- [ ] Implicit else: If an `else` block contains only a `return` or `throw` statement, then it should be executed implicitly.
- [x] Prevent rogue empty lines.
- [ ] Soft/hard limits on comments
- [x] Space around the `arrow` in array declarations. 
- [ ] Variable naming conventions to use camel case - i.e. `$pageNum `instead of `$page_num`
- [ ] Function naming conventions (underscore to divide words) - to be confirmed if isn't already working as expected
- [ ] Proper indentation inside functions, loops etc.

rsamborski:
* Variable naming conventions to use camel case - i.e. $pageNum instead of $page_num
* Function naming conventions (underscore to divide words) - to be confirmed if isn't already working as expected
* Proper indentation inside functions, loops etc.

saranshdhingra:
Closing this as we are discussing to unify efforts across multiple repos.

PR #1533, opened 2021-09-23
Feat: Added new rules in .php-cs-fixer.dist.php

saranshdhingra:
WIP for #1465 

Following are the test cases considered:
### Input:
```
print("Test double quote string" . PHP_EOL); // -> Only this is changed
print("Test string with variable $a" . PHP_EOL);
print("Test string with 'single quotes'" . PHP_EOL);
print("Test string with \n\t\\\"escape characters" . PHP_EOL);
```
### Input:
```
$a = (int)'89';
$b = (int)   '90';
```
### Output:
```
$a = (int) '89';
$b = (int) '90';
```
### Input:
```
// simple assignment
$a=123;
// aray
$b  =['a'=>'b'];
$c=  ['a'  =>  'b'];
// arithmetic
$d=1+9;
$e = 8*10;
// logical
$f = true||false;
$g = 0 &&false;
// bitwise
$h = 120^20;
// comparison
$i = $a<50;
// unary is not touched
$j = !0;
```
### Output:
```
$a = 123;
$b = ['a' => 'b'];
$c = ['a' => 'b'];
$d = 1 + 9;
$e = 8 * 10;
$f = true || false;
$g = 0 && false;
$h = 120 ^ 20;
$i = $a < 50;
$j = !0;
```
### Input:
```
$a = 3;


$b = 5;



$c = 8;

// not a blank line
$d = 10;
```
### Output:
```
$a = 3;

$b = 5;

$c = 8;

// not a blank line
$d = 10;
```

### Input:
```
/**
 * Description 
 * @param param1
 * @param param2
 */

/**
 * Description
 *
 * 
 * 
 * @param param1
 * @param param2
 * @throws Exception
 * @return void
 */

/**
 * Multiline
 *
 * Description
 * 
 * 
 * @param param1
 * @param param2
 */
```
### Output:
```
/**
 * Description
 * @param param1
 * @param param2
 */

/**
 * Description
 *
 * @param param1
 * @param param2
 * @throws Exception
 * @return void
 */

/**
 * Multiline
 *
 * Description
 *
 * @param param1
 * @param param2
 */
```

### DIFF
--- a/appengine/standard/getting-started/test/CloudSqlTest.php
+++ b/appengine/standard/getting-started/test/CloudSqlTest.php
@@ -144,11 +144,11 @@ class CloudSqlTest extends TestCase
 
         // Clean up.
         $result = $model->delete($breakfastId);
-        $this->assertTrue((bool)$result);
+        $this->assertTrue((bool) $result);
         $this->assertFalse($model->read($breakfastId));
-        $this->assertTrue((bool)$model->read($bellId));
+        $this->assertTrue((bool) $model->read($bellId));
         $result = $model->delete($bellId);
-        $this->assertTrue((bool)$result);
+        $this->assertTrue((bool) $result);
         $this->assertFalse($model->read($bellId));
     }
 }
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -16,6 +16,22 @@ $config
         'return_type_declaration' => [
             'space_before' => 'none'
         ],
+        // only converts simple strings in double quotes to single quotes
+        // ignores strings using variables, escape characters or single quotes inside
+        'single_quote' => true,
+        // there should be a single space b/w the cast and it's operand
+        'cast_spaces' => ['space' => 'single'],
+        // there shouldn't be any trailing whitespace at the end of a non-blank line
+        'no_trailing_whitespace' => true,
+        // there shouldn't be any trailing whitespace at the end of a blank line
+        'no_whitespace_in_blank_line' => true,
+        // there should be a space around binary operators like (=, => etc)
+        'binary_operator_spaces' => ['default' => 'single_space'],
+        // deals with rogue empty blank lines
+        'no_extra_blank_lines' => ['tokens' => ['extra']],
+        // reduces multi blank lines b/w phpdoc description and @param to a single line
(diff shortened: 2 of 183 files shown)


bshaffer:
@saranshdhingra  I know the failures have nothing to do with this PR but we really should take some time to get them resolved.

rsamborski:
LGTM + I agree with Brent's comment regarding the tests. Not sure though if it should be blocker for this PR.

appengine/standard/getting-started/test/CloudSqlTest.php
```diff
@@ -20,22 +20,36 @@
...
+    private static $process;
 
     public function setUp(): void
     {
         $connection = $this->requireEnv('CLOUDSQL_CONNECTION_NAME');
         $dbUser = $this->requireEnv('CLOUDSQL_USER');
         $dbPass = $this->requireEnv('CLOUDSQL_PASSWORD');
         $dbName = getenv('CLOUDSQL_DATABASE_NAME') ?: 'bookshelf';
-        $socket = "/tmp/cloudsql/${connection}";
+        $socketDir = '/tmp/cloudsql';
```
bshaffer, line 36: This should use `DB_SOCKET_DIR`
```suggestion
        $socketDir = $this->requireEnv('DB_SOCKET_DIR');
```

appengine/standard/getting-started/test/CloudSqlTest.php
```diff
@@ -20,22 +20,36 @@
...
+        // create the directory to store the unix socket for cloud_sql_proxy
+        if (!is_dir($socketDir)) {
+            mkdir($socketDir, 0755, true);
+        }
+
+        self::$process = new Process(['cloud_sql_proxy', '-instances=' . $connection, '-dir', $socketDir]);
+        self::$process->start();
+        self::$process->waitUntil(function ($type, $buffer) {
+            return str_contains($buffer, 'Ready for new connections');
+        });
```
bshaffer, line 48: Ideally all of the code to spin up the cloudsqlproxy would be moved to `google/cloud-tools`, as I see it used in several samples.

Definitely not something needed in this PR though!

bshaffer (approved):

Issue #1534, opened 2021-09-28
storage: Change Public Access Prevention "unspecified" to "inherited"


saranshdhingra:
More details of this issue: https://github.com/googleapis/google-cloud-php/issues/4524

This issue should track the work in samples.

PR #1535, opened 2021-09-28
Storage: Changed references from 'unspecified' to 'inherited'

saranshdhingra:
Changed the references of `unspecified` in storage samples and tests to `inherited` as per #1534 

~One test would fail right now as the API hasn't changed the response yet.~

### DIFF
--- a/storage/src/set_public_access_prevention_inherited.php
+++ b/storage/src/set_public_access_prevention_inherited.php
@@ -23,15 +23,15 @@
 
 namespace Google\Cloud\Samples\Storage;
 
-# [START storage_set_public_access_prevention_unspecified]
+# [START storage_set_public_access_prevention_inherited]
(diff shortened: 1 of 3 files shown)


shaffeeullah:
This looks good. Thanks for doing this!

shaffeeullah:
@saranshdhingra this can be merged when it is ready.

saranshdhingra:
Note about the old snippet(`set_public_access_prevention_unspecified`): The default PAP has been changed from `unspecified` to `inherited`, so I definitely had to remove the test for that snippet.
The snippet itself hasn't been removed yet as it is being used in the docs.
Once, the use changes to the new snippets, another PR will remove the old snippet from the samples.
CC: @bshaffer @shaffeeullah

bshaffer (approved):

PR #1548, opened 2021-11-05
Add a quickstart sample for storage transfer service

JesseLovelace:
Creates the storagetransfer directory and adds a quickstart sample for storage transfer service, and a test

### DIFF
--- a/storagetransfer/composer.json
+++ b/storagetransfer/composer.json
@@ -1,0 +1,10 @@
+{
+    "require": {
+    	"google/cloud-storage-transfer": "0.1.1",
+        "paragonie/random_compat": "^9.0.0"
+    },
+    "require-dev": {
+    	"google/cloud-storage": "^1.20.1",
+        "guzzlehttp/guzzle": "^7.0"
+    }
+}
--- a/storagetransfer/composer.json
+++ b/storagetransfer/composer.json
@@ -4,7 +4,6 @@
         "paragonie/random_compat": "^9.0.0"
     },
     "require-dev": {
-    	"google/cloud-storage": "^1.20.1",
-        "guzzlehttp/guzzle": "^7.0"
+        "google/cloud-storage": "^1.20.1"
     }
 }
--- a/storagetransfer/src/quickstart.php
+++ b/storagetransfer/src/quickstart.php
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 namespace Google\Cloud\Samples\StorageTransfer;
 
 # [START storagetransfer_quickstart]
@@ -23,7 +24,6 @@ use Google\Cloud\StorageTransfer\V1\TransferJob\Status;
 use Google\Cloud\StorageTransfer\V1\TransferSpec;
 use Google\Cloud\StorageTransfer\V1\GcsData;
 
-
 /**
  * Creates and runs a tranfser job between two GCS buckets
  *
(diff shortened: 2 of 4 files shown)


storagetransfer/src/quickstart.php
```diff
@@ -0,0 +1,58 @@
...
+    $transferJob = new TransferJob([
+        'project_id' => $projectId,
+        'transfer_spec' => new TransferSpec([
+            'gcs_data_sink' => new GcsData(['bucket_name' => $sourceGcsBucketName]),
+            'gcs_data_source' => new GcsData(['bucket_name' => $sourceGcsBucketName])
+        ]),
+        'status' => Status::ENABLED
+    ]);
+
+    $client = new StorageTransferServiceClient();
```
frankyn, line 48: Nit: I think you're leaning to create client after TransferJob creation, so it's created right before it's needed. This may need to be updated in other language quickstarts.

frankyn (commented):
Want to confirm that tests are ran because I see a code coverage failure for PHP8, but will leave it as a question for the PHP experts.
4fMNjboxUdxHV5L

JesseLovelace:
I know that the tests ran because they failed at first due to the STS API not being enabled, then passed after I enabled it and re-ran

JesseLovelace:
@bshaffer would appreciate a review!

storagetransfer/composer.json
```diff
@@ -0,0 +1,10 @@
+{
+    "require": {
+    	"google/cloud-storage-transfer": "0.1.1",
+        "paragonie/random_compat": "^9.0.0"
+    },
+    "require-dev": {
+    	"google/cloud-storage": "^1.20.1",
+        "guzzlehttp/guzzle": "^7.0"
+    }
```
bshaffer, line 9: The require-dev dependency for guzzle shouldn't be necessary because it's in the test deps already

storagetransfer/src/quickstart.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @param string $projectId Your Google Cloud project ID.
+ * @param string $sourceGcsBucketName The name of the GCS bucket to transfer objects from.
+ * @param string $sinkGcsBucketName The name of the GCS bucket to transfer objects to.
+ */
+function quickstart($projectId, $sourceGcsBucketName, $sinkGcsBucketName)
+{
+    // $project = 'my-project-id';
+    // $sourceGcsBucketName = 'my-source-bucket';
+    // $sinkGcsBucketName = 'my-sink-bucket';
```
bshaffer, line 38: We don't want these vars here, they're covered in the doc block for the function parameters

storagetransfer/test/StorageTransferTest.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        self::checkProjectEnvVars();
+        self::$storage = new StorageClient();
+        self::$sts = new StorageTransferServiceClient();
+        self::$sourceBucket = self::$storage->createBucket(
+            sprintf('php-source-bucket-%s', time())
+        );
+        self::$sinkBucket = self::$storage->createBucket(
+            sprintf('php-sink-bucket-%s', time())
+        );
```
bshaffer, line 46: This is apparently resulting in a race condition (see the PHP 8.0 failures). Try something like this:

```suggestion
        $uniqueBucketId = time() . rand();
        self::$sourceBucket = self::$storage->createBucket(
            sprintf('php-source-bucket-%s', $uniqueBucketId)
        );
        self::$sinkBucket = self::$storage->createBucket(
            sprintf('php-sink-bucket-%s', $uniqueBucketId)
        );
```

bshaffer (commented):
These tests are failing with the following error in PHP 8.0:

```
1) Google\Cloud\Samples\StorageTransfer\StorageTransferTest::testQuickstart
Google\Cloud\Core\Exception\ConflictException: {"error":{"code":409,"message":"You already own this bucket. Please select another name.","errors":[{"message":"You already own this bucket. Please select another name.","domain":"global","reason":"conflict"}]}} in /tmpfs/src/github/php-docs-samples/storagetransfer/vendor/google/cloud-core/src/RequestWrapper.php:368
```

storagetransfer/src/quickstart.php
JesseLovelace: This is actually on purpose, the doc block provides a description of what the parameters are, while these comments provide an example of how to actually set them. This is how we do all of our PHP samples in storage, for example see [here](https://cloud.google.com/storage/docs/access-control/using-iam-permissions#storage_add_bucket_conditional_iam_binding-php) or any of the other PHP samples on this page.

bshaffer: ok seems fine to me

bshaffer (approved):

PR #1553, opened 2021-11-18
refactor(functions): declarative function signature for helloworld_http

anniefu:
Update to the new [declarative function signature](https://github.com/GoogleCloudPlatform/functions-framework-php/pull/116)  style that was introduced in Functions Framework v1.1.0

### DIFF
--- a/functions/helloworld_http/composer.json
+++ b/functions/helloworld_http/composer.json
@@ -1,5 +1,6 @@
 {
     "require": {
+        "php": ">= 7.4",
         "google/cloud-functions-framework": "^1.1"
     },
     "scripts": {
--- a/functions/helloworld_http/composer.json
+++ b/functions/helloworld_http/composer.json
@@ -1,6 +1,6 @@
 {
     "require": {
-        "google/cloud-functions-framework": "^1.0"
+        "google/cloud-functions-framework": "^1.1"
     },
     "scripts": {
         "start": [
--- a/functions/helloworld_http/index.php
+++ b/functions/helloworld_http/index.php
@@ -16,7 +16,14 @@ limitations under the License.
 <!-- [START functions_helloworld_http] ->
 <?php
 
+use Google\CloudFunctions\FunctionsFramework;
 use Psr\Http\Message\ServerRequestInterface;
+
+// Register the function with Functions Framework.
+// This enables omitting the `FUNCTIONS_SIGNATURE_TYPE=http` environment
+// variable when deploying. The `FUNCTION_TARGET` environment variable should
+// match the first parameter.
+FunctionsFramework::http('helloHttp', 'helloHttp');
 
 function helloHttp(ServerRequestInterface $request): string
 {
--- a/functions/helloworld_http/SampleUnitTest.php
+++ b/functions/helloworld_http/SampleUnitTest.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Copyright 2020 Google LLC.
  *
--- a/functions/helloworld_pubsub/composer.json
+++ b/functions/helloworld_pubsub/composer.json
@@ -1,12 +1,12 @@
 {
     "require": {
-        "php": ">= 7.4",
-        "google/cloud-functions-framework": "^1.0.0",
+        "cloudevents/sdk-php": "^1.0",
+        "google/cloud-functions-framework": "^1.1"
     },
     "scripts": {
         "start": [
             "Composer\\Config::disableProcessTimeout",
-            "FUNCTION_SIGNATURE_TYPE=cloudevent FUNCTION_TARGET=helloworldPubsub php -S localhost:${PORT:-8080} vendor/bin/router.php"
+            "FUNCTION_TARGET=helloworldPubsub php -S localhost:${PORT:-8080} vendor/bin/router.php"
         ]
     },
     "require-dev": {
--- a/functions/helloworld_pubsub/composer.json
+++ b/functions/helloworld_pubsub/composer.json
@@ -1,6 +1,7 @@
 {
     "require": {
-        "google/cloud-functions-framework": "^1.0.0"
+        "php": ">= 7.4",
+        "google/cloud-functions-framework": "^1.0.0",
     },
     "scripts": {
         "start": [


functions/helloworld_http/index.php
```diff
@@ -16,8 +16,15 @@
...
 <?php
 
+use Google\CloudFunctions\FunctionsFramework;
 use Psr\Http\Message\ServerRequestInterface;
 
+// Register the function with Functions Framework.
+// This enables omitting the `FUNCTIONS_SIGNATURE_TYPE=http` environment
+// variable when deploying. The `FUNCTION_TARGET` environment variable should
+// match the first parameter.
+FunctionsFramework::http('helloHttp', 'helloHttp');
```
bshaffer, line 26: I think it would be more idiomatic to use an anonymous function here, unless this is more inline with the other language samples and for some reason you prefer them to be named?

anniefu: I did it this way cause it's easier to do if you already have an existing function that works today and are switching over to declarative function signatures. I'm fine with any anonymous function too. Which use case do you think is more important?

BTW, is it possible to disable php 7.3 validation for the functions/ dir? CloudEvents SDK and FF PHP now require 7.4, and GCF has only ever supported PHP 7.4, so we were ok with that.

anniefu: Ohh, also the anonymous function can't be tested.

bshaffer: We could test them by using PHP reflection:

```php
$reflectionProperty = (new ReflectionClass(Invoker::class))->getProperty('registeredFunctions');
$reflectionProperty->setAccessible(true);
$function = $reflectionProperty->getValue()['functionName'];

$function($request); // invoke the function
```

anniefu: I don't think this would be an improvement over the current unit testing strategy of just calling `helloHttp` in a unit test?

You shouldn't need to understand anything about the functions framework's implementation to test your function in a unit test. And we provide instructions for how to test it as a web server locally as a separate concern.

bshaffer: @anniefu Yes, we should definitely disable PHP 7.3... the easiest way to do this is just to add `"php": "7.4"` in the "requires" section of `composer.json`. See [here](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/cloud_sql/sqlserver/pdo/composer.json#L9) for an example.

bshaffer: > You shouldn't need to understand anything about the functions framework's implementation to test your function in a unit test. And we provide instructions for how to test it as a web server locally as a separate concern.

I don't disagree. This is a good argument for adding a way to `get` registered functions... Because we do have samples which show users how to test their functions. I remember you were against adding something like this to `FunctionsFramework`, but that seems like the correct place for it IMHO.

The thing is, it would be more idiomatic in PHP to pass anonymous functions to the registry. Declaring functions in global scope as we have here is something the community has been moving away from. So while I agree they should be easy to test (since we have samples specifically for testing), I don't agree that we should keep them in global scope.

anniefu: Hmm, could we leave this sample update for now and iterate on the FF/samples as we go? 

Since the current Functions paradigm is declaring a global function and users may already have tests written like this, I think adding the one FunctionsFramework registration line is a meaningful, purely additive update that doesn't require users to modify their existing function or testing code.

We should continue iterating in the FF on the balance between global scoping and convenience to the function author. Functions are a bit unique because we don't expect any additional application code surrounding the function, so globals may not be as problematic.

I'm a little hesitant to tie writing and testing the functions too heavily to FF library. I think it's a nice user story to be able to focus on "write a function that handles an http request and unit test it like any other function" and as an afterthought "ok now I'll hook it up to FF". Maybe adding a testing-only FunctionsFramework package as we go is a good compromise?

functions/helloworld_http/composer.json
```diff
@@ -1,6 +1,7 @@
 {
     "require": {
-        "google/cloud-functions-framework": "^1.0"
+        "php": ">= 7.4",
```
anniefu, line 3: Starting a new thread specifically for the PHP 7.3 Kokoro tests. I think the check needs to be disabled on the GitHub side somehow? I'm not sure if you can specifically exclude a directory from a GitHub check...

bshaffer: The tests are set up in Kokoro, and one job loops through all the sample directories and executes their tests. Because of this, to disable testing of an individual sample, the job which executes all the tests needs to be aware of which versions are valid. Adding a minimum PHP version to `composer.json` is the way we chose to implement this logic in the samples, and I think it's a good place for it.

We've discussed refactoring the tests so they each get an individual kokoro job. I think this would be an improvement in general, but even then the logic of which samples are valid for which PHP versions needs to be _somewhere_, and I think `composer.json` is still the best place.

functions/helloworld_http/index.php
bshaffer: As for adding a separate package for testing, I'm against this, as it would be a lot of overhead to provide an additional package for what would ultimately amount to just a few lines of code.

anniefu: It is just a few lines, but the testing utility would be more about hiding the internals of FF. There's no reason for people to know about the `Invoker` class really. I could also see the testing package growing over time.

I'm also ok with adding a getter utility to the FF to make testing cleaner. Is that a change you would want to send out a PR for? Otherwise, I can take a swing at it and add you as a reviewer.

functions/helloworld_http/composer.json
anniefu: Hmm, it looks like the PHP 7.3 tests are still running despite this update to the PR? Do I need to first update just the `composer.json` in a separate PR?

functions/helloworld_http/index.php
bshaffer: I just added https://github.com/GoogleCloudPlatform/functions-framework-php/issues/118 to address this. In the meantime, I think what you have here works.

functions/helloworld_http/composer.json
bshaffer: Addressed in https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1558

bshaffer (approved):

PR #1555, opened 2021-11-18
refactor(functions): declarative function signature for helloworld_storage

anniefu:
Update to the new [declarative function signature](https://github.com/GoogleCloudPlatform/functions-framework-php/pull/116)  style that was introduced in Functions Framework v1.1.0


bshaffer (approved):

bshaffer (changes requested):
Looks like we forgot to update `composer.json` to use the latest version of the functions framework (`^1.1.0`)

bshaffer (approved):

PR #1568, opened 2021-12-06
feat(spanner): Added sample for copyBackup

saranshdhingra:


spanner/src/copy_backup.php
```diff
@@ -0,0 +1,72 @@
...
+    $ready = ($newBackup->state() == Backup::STATE_READY);
+
+    if ($ready) {
+        print('Backup is ready!' . PHP_EOL);
+        $info = $newBackup->info();
+        printf(
+            'Backup %s of size %d bytes was copied at %s from the source backup %s' . PHP_EOL,
+            basename($info['name']), $info['sizeBytes'], $info['createTime'], $sourceBackupId);
+    } else {
+        print('Backup is not ready!' . PHP_EOL);
```
bshaffer, line 65: This is not very helpful. Do we know why this would be? If it's a race condition maybe we could add a sleep, and if not we should throw an exception or print an error with the state like this:

```php
printf('Unexpected state: %s', Backup::name($newBackup->state()));
```

spanner/test/spannerBackupTest.php
```diff
@@ -163,6 +163,21 @@ public function testListBackupOperations()
...
+    public function testCopyBackup()
+    {
+        $newBackupId = 'copy-' . self::$backupId;
+
+        $output = $this->runFunctionSnippet('copy_backup', [
+            self::$instanceId,
+            $newBackupId,
+            self::$backupId
+        ]);
+        $this->assertStringContainsString($newBackupId, $output);
```
bshaffer, line 178: Id prefer a little more specific check:

```php
$regex = '/Backup %s of size \d+ bytes was copied at \w+ from the source backup %s/';
$this->assertRegExp(sprintf($regex, self::$backupId, $newBackupId), $output);
```

spanner/src/copy_backup.php
saranshdhingra: I modelled it using the `create_backup` sample. But, I agree, I'll try to find out the reason behind it and update both samples accordingly.

spanner/src/copy_backup.php
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Copy a backup from another source backup.
+ * Example:
+ * ```
+ * copy_backup($instanceId, $backupId, $sourceBackupId);
+ * ```
+ *
+ * @param string $instanceId The Spanner instance ID.
+ * @param string $backupId The Spanner backup ID.
+ * @param string $sourceBackupId The Spanner backup ID of the source.
```
asthamohta, line 39: nit: could you change this is add that we need the "full path" over here

spanner/src/copy_backup.php
saranshdhingra: I am not yet sure why this state ever occurs, but the `create_backup` snippet of all other languages(except node) don't have an `else` block.

spanner/src/copy_backup.php
saranshdhingra: Updated the parameters to contain the source and destination instances. IMO this is better than asking the user to input the path.

bshaffer (commented):
Looks great! I'll hold off approval until it's ready to merge and the tests pass.

spanner/src/update_backup.php
```diff
@@ -25,6 +25,7 @@
 
 // [START spanner_update_backup]
 use Google\Cloud\Spanner\SpannerClient;
+use \DateTime;
```
bshaffer, line 28: leading slash is redundant for `use` statements
```suggestion
use DateTime;
```

spanner/src/update_backup.php
```diff
@@ -40,10 +41,14 @@ function update_backup($instanceId, $backupId)
...
 
-    // Expire time must be within 366 days of the create time of the backup.
-    $newTimestamp = new \DateTime('+30 days');
-    $backup->updateExpireTime($newTimestamp);
+    // The new expire time can't be greater than maxExpireTime for the backup.
+    $newExpireTime = new DateTime('+30 days');
+    $maxExpireTime = new DateTime($backup->info()['maxExpireTime']);
+    $newExpireTime = $newExpireTime > $maxExpireTime ? $maxExpireTime : $newExpireTime;
+
+    $backup->updateExpireTime($newExpireTime);
```
bshaffer, line 51: The php [max function](https://www.php.net/manual/en/function.max.php) works with datetimes as well:
```suggestion
    $backup->updateExpireTime(max($newExpireTime, $maxExpireTime));
```

bshaffer (approved):

spanner/src/update_backup.php
saranshdhingra: Used the suggestion, but just to clarify, in this case we need `min`.

saranshdhingra:
This should be ready to merge once https://github.com/googleapis/google-cloud-php/pull/4732 is merged and released.

PR #1591, opened 2022-02-04
fix: functions command

bshaffer:
Composer switched recently to require anything in `bin` to be an executable binary file. The way we we use `vendor/bin/router.php` is as a PHP file (by executing it as part of the PHP script). as a result, this has been failing with a PHP fatal error for our users.

By making the path to `router.php` absolute, we ensure compatibility for all our users on any version of composer.


bshaffer:
GCF tests seem to be flakey

grayside (approved):

PR #1592, opened 2022-02-10
Bigquery: Added sample to get large dataset with legacy SQL

saranshdhingra:
Added sample to get large dataset with legacy SQL in Bigquery


bigquery/api/src/query_legacy_large_results.php
```diff
@@ -0,0 +1,52 @@
...
+
+// Include Google Cloud dependendencies using Composer
+require_once __DIR__ . '/../vendor/autoload.php';
+
+if (count($argv) != 2) {
+    return printf("Usage: php %s PROJECT_ID\n", __FILE__);
+}
+list($_, $projectId) = $argv;
+
+// [START bigquery_query_legacy_large_results]
```
bajajneha27, line 32: Correct me if I'm wrong, I thought the region tag for this sample is [bigquery_query_legacy ](https://github.com/googleapis/python-bigquery/blob/35627d145a41d57768f19d4392ef235928e00f72/samples/client_query_legacy_sql.py#L18)

bshaffer: Python _also_ has a `large_results` sample, but it looks different than this, as it uses the `allow_large_results` option:

https://github.com/googleapis/python-bigquery/blob/679e8edc27f24a312243f5587e1c71798902fce8/samples/client_query_destination_table_legacy.py#L18

So something is missing here, and I believe it's that you need to call `->allowLargeResults()` on the `$jobConfig` object.

PR #1596, opened 2022-02-23
chore: stricter static analysis

bshaffer:


bshaffer:
@saranshdhingra I'm getting an error on the `unpack` call [here](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/bigtable/src/filter_composing_condition.php#L50) because `exactMatch` expects a string and not an array, and the second argument to `unpack` expects a string and not an integer. I think there is a mistake in this sample.

rsamborski (commented):
compute folder reviewed - LGTM. Nevertheless I see failing tests in other parts, so I assume it's still WIP

vishwarajanand:
@bshaffer  if this is not being worked upon, maybe we can close this. The current PR will need significant effort to get merged due to conflicts.

bshaffer:
@vishwarajanand a lot of work was done to add this, and it's a very important tool that will ensure quality samples for our users. We should work on getting this reviewed and merged rather than just closing it.

bshaffer:
@vishwarajanand I just resolved the merge conflicts in under 4 minutes, so "significant efforts" was a bit of a misjudgement.

bshaffer:
I've noticed something strange - the static analysis tests are skipping every directory even though they should be running. I'll have to look into this.

vishwarajanand (approved):
Lgtm

PR #1600, opened 2022-03-02
Implementing sample firestore_data_get_as_custom_type

vishwarajanand:
## Background

PHP has not yet implemented any sample to show how a class can be fetched using PHP code. This sample will cover that.

## Change Summary

1. Added a new custom type which encapsulates the underlying class -> `City`.
2. Added `from_map` and `to_map` in City class.
3. Added `testGetCustomType` which can be executed and validated.

## Tests

1. The custom type class `City` and the custom type get both seem to run fine by the following command:

```
➜  firestore git:(get_as_custom_type) ✗ 
➜  firestore git:(get_as_custom_type) ✗ XDEBUG_MODE=coverage  ../testing/vendor/bin/phpunit -c phpunit.xml.dist test
PHPUnit 8.5.23 by Sebastian Bergmann and contributors.

....................S...............S...................S.        58 / 58 (100%)

Time: 2.92 minutes, Memory: 16.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 58, Assertions: 123, Skipped: 3.

Generating code coverage report in Clover XML format ... done [114 ms]
➜  firestore git:(get_as_custom_type) ✗ 
➜  firestore git:(get_as_custom_type) ✗ 

```

> Test run succeeded on local.


firestore/src/City.php
```diff
@@ -0,0 +1,92 @@
...
+            'name' => $this->name,
+            'state' => $this->state,
+            'country' => $this->country,
+            'capital' => $this->capital,
+            'population' => $this->population,
+            'regions' => $this->regions,
+        ];
+
+        return $dest;
+        # [END_EXCLUDE]
```
bshaffer, line 77: what is this? Will it extend the code here from the sample? if so, why not exclude the whole function?

firestore/src/data_get_as_custom_type.php
```diff
@@ -0,0 +1,57 @@
...
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/firestore/README.md
+ */
+
+namespace Google\Cloud\Samples\Firestore;
+
+use Google\Cloud\Firestore\FirestoreClient;
+
+include 'City.php';
```
bshaffer, line 28: This is relative to the user's CWD, so make it absolute with `__DIR__`:
```suggestion
include __DIR__ . '/City.php';
```

firestore/src/City.php
vishwarajanand: Similar to [Python equivalent](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/firestore/cloud-client/snippets.py#L136), this placeholder class function depicts how to do get a class object queries from Firestore by implementing `from_map` and `to_map` functions.

vishwarajanand: Commenting the whole function could confuse developers on how a map got translated to a class object.

bshaffer: That makes sense to me (although I'd prefer to just keep the function).

Can we put a comment outside the exclude to make this clear, something like...

```
// implementation of this function is excluded for brevity
```

or 

```
// implementation of fromArray goes here
```

firestore/src/City.php
```diff
@@ -0,0 +1,92 @@
...
+    public $population;
+    public $regions;
+
+    public function __construct(
+        $name,
+        $state,
+        $country,
+        $capital = false,
+        $population = 0,
+        $regions = []
```
bshaffer, line 42: we should be using typehints for everything. Right now it's really hard to know what the types are, and which ones can be nullable. 
```suggestion
        string $name,
        string $state,
        string $country,
        bool $capital = false,
        int $population = 0,
        array $regions = []
```

firestore/src/City.php
```diff
@@ -0,0 +1,92 @@
...
+    ) {
+        $this->name = $name;
+        $this->state = $state;
+        $this->country = $country;
+        $this->capital = $capital;
+        $this->population = $population;
+        $this->regions = $regions;
+    }
+
+    public static function from_map($source)
```
bshaffer, line 52: three things here:
1. names should be camel cased
2. maps don't exist in PHP, so this should be `array`, not `map`
3. we need return types and argument type hints
```suggestion
    public static function fromArray(array $source): void
```

firestore/src/City.php
```diff
@@ -0,0 +1,92 @@
...
+        $city = new City($source['name'], $source['state'], $source['country']);
+        $city->capital = $source['capital'] ?? null;
+        $city->population = $source['population'] ?? null;
+        $city->regions = $source['regions'] ?? null;
+
+        return $city;
+        # [END_EXCLUDE]
+    }
+
+    public function to_map()
```
bshaffer, line 64:
```suggestion
    public function toArray(): array
```

firestore/src/City.php
```diff
@@ -0,0 +1,92 @@
...
+
+    public function __toString()
+    {
+        return ("Custom Type data(\
+        name={$this->name}, \
+        country={$this->country}, \
+        population={$this->population}, \
+        capital={$this->capital}, \
+        regions={$this->regions}\
+       )");
```
bshaffer, line 88: We usually prefer `sprintf` for something like this, to make it more readable. Also, `regions` and `capital`  require special handling because they are not strings (`population` is okay because PHP will properly cast it to a string):
```suggestion
        return sprintf(<<<EOF
Custom Type data(
    name=%s,
    country=%s,
    population=%s,
    capital=%s,
    regions=%s
)
EOF,
            $this->name,
            $this->country,
            $this->population,
            $this->captital ? 'true' : 'false', 
            implode(", ", $this->regions)
        );
```

firestore/src/City.php
```diff
@@ -0,0 +1,92 @@
...
+            'capital' => $this->capital,
+            'population' => $this->population,
+            'regions' => $this->regions,
+        ];
+
+        return $dest;
+        # [END_EXCLUDE]
+    }
+
+    public function __toString()
```
bshaffer, line 80: I'm a little confused why we'd expose the `__toString()` method but exclude the others. The logic here is not very interesting, so if we're excluding functions for brevity then this should be excluded. I'm guessing it makes sense in the context of the sample?

firestore/test/firestoreTest.php
```diff
@@ -381,6 +381,21 @@ public function testRetrieveCreateExamples()
...
+     * @depends testRetrieveCreateExamples
+     */
+    public function testGetCustomType()
+    {
+        $output = $this->runFirestoreSnippet('data_get_as_custom_type');
+        $this->assertStringContainsString('Document data:', $output);
+        $this->assertStringContainsString('Google\Cloud\Samples\Firestore\City', $output);
+        $this->assertStringContainsString('[population] => 860000', $output);
+        $this->assertStringContainsString('[state] => CA', $output);
+        $this->assertStringContainsString('[capital] =>', $output);
```
bshaffer, line 394: make this assert a value, and add a test for `regions`

firestore/src/City.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->regions = $regions;
+    }
+
+    public static function from_map($source)
+    {
+        # [START_EXCLUDE]
+        $city = new City($source['name'], $source['state'], $source['country']);
+        $city->capital = $source['capital'] ?? null;
+        $city->population = $source['population'] ?? null;
+        $city->regions = $source['regions'] ?? null;
```
bshaffer, line 58: this doesn't make sense the way you're doing this... they're available in the constructor properties so we should be consistent and set them there. Also, you're using nulls instead of the types / defaults for `capital`, `population`, and `region`. This is bad because 1. you're overriding the defaults you have in the constructor, and 2. you're adding `null` types but not adding proper handling of those types in the `toString` function. This is what I think you want:
```suggestion
        $city = new City(
            $source['name'],
            $source['state'],
            $source['country'],
            $source['capital'] ?? false,
            $source['population'] ?? 0,
            $source['regions'] ?? []
        );
```

This isn't great because our default values are essentially being defined twice, but that's better than setting them to `null`. Another option would be: 
```suggestion
        $city = new City($source['name'], $source['state'], $source['country']);
        if (isset($source['capital'])) {
            $city->capital = $source['capital'];
        }
        if (isset($source['population'])) {
            $city->population = $source['population'];
        }
        if (isset($source['regions'])) {
            $city->regions = $source['regions'];
        }
```

firestore/src/City.php
```diff
@@ -0,0 +1,92 @@
...
+
+# [START firestore_data_custom_type_definition]
+class City
+{
+    public $name;
+    public $state;
+    public $country;
+    public $capital;
+    public $population;
+    public $regions;
```
bshaffer, line 34: These should have types as well
```suggestion
    public string $name;
    public string $state;
    public string $country;
    public bool $capital;
    public int $population;
    public array $regions;
```

bshaffer (changes requested):

firestore/src/City.php
vishwarajanand: Added `public static function fromArray(array $source): City`

firestore/src/City.php
vishwarajanand: I wasnt used to using sprintf.. it seems v cluttered personally to me. 😅 
But I edited it since thats what we follow at Google.

firestore/src/City.php
vishwarajanand: This method is actually used by tests only. I tried to mimic the python sample which also uses a similar approach, [ref link here](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/firestore/cloud-client/snippets.py#L156)

firestore/test/firestoreTest.php
vishwarajanand: This firestore document ([source link here](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/firestore/src/data_get_dataset.php#L42)) didnt have region field hence I had originally skipped it. Now I put it back like this:

```
        $this->assertStringContainsString('[regions] => Array', $output);
```

bshaffer:
1. The linter failed... take a look at the kokoro job to see why.
2. My suggestion to add property types was a bad one, as we are still running tests on PHP 7.3, and that syntax was added in 7.4.  So we should remove it for now (my fault).

firestore/src/data_get_as_custom_type.php
```diff
@@ -0,0 +1,57 @@
...
+        'projectId' => $projectId,
+    ]);
+    # [START firestore_data_get_as_custom_type]
+    $docRef = $db->collection('samples/php/cities')->document('SF');
+    $snapshot = $docRef->snapshot();
+    $city = City::fromArray($snapshot->data());
+
+    if ($snapshot->exists()) {
+        printf('Document data:' . PHP_EOL);
+        print_r($city);
```
bshaffer, line 48: If we added a toString function, why are we printing the object?

Im ok doing it this way, but then we should remove the toString funtion on the object because it isn't being used

vishwarajanand: It took a while to pinpoint that I missed the `string` cast here. I intended to use `__toString`because printing object will have to test the classpath like`Google\Cloud\Samples\Firestore\City` which might change later causing the tests to be flaky.

firestore/src/data_get_as_custom_type.php
```diff
@@ -45,7 +45,7 @@ function data_get_as_custom_type(string $projectId): void
 
     if ($snapshot->exists()) {
         printf('Document data:' . PHP_EOL);
-        print_r($city);
+        print_r((string) $city);
```
bshaffer, line 48: NIT: Since it's already a string, let's just make this `print` instead of `print_r`
```suggestion
        print((string) $city);
```

firestore/src/data_get_as_custom_type.php
bshaffer: That makes sense! I've added one final nit and then this'll be good to go!

bshaffer (approved):

PR #1614, opened 2022-04-07
feat(spanner): Added Samples for Postgres Dialect

saranshdhingra:


saranshdhingra:
This should be merged after https://github.com/googleapis/google-cloud-php/pull/5165 is merged and released.

bshaffer (commented):
These look great but, shouldn't the files and functions be without "spanner", since we are already in the spanner namespace it is a bit redundant, and it will also be more consistent with the region tags (e.g. `pg_spanner_add_column` => `pg_add_column`)

saranshdhingra:
> These look great but, shouldn't the files and functions be without "spanner", since we are already in the spanner namespace it is a bit redundant, and it will also be more consistent with the region tags (e.g. `pg_spanner_add_column` => `pg_add_column`)

Hey @bshaffer 
Thanks for the catch.

I actually followed the go [PR](https://github.com/hengfengli/google-cloud-go/pull/6/files) for the sample names. But I see the point as these things should make sense for our repo too.

Do you want me to remove the `spanner` from the file names or make it the same as the region tags like `spanner_postgresql_add_column`?

bshaffer:
@saranshdhingra I'd be happy with just removing spanner from the filename and function name. This will be most like our other samples also!

spanner/src/pg_cast_data_type.php
```diff
@@ -0,0 +1,61 @@
...
+ */
+function pg_cast_data_type(string $instanceId, string $databaseId): void
+{
+    $spanner = new SpannerClient();
+    $instance = $spanner->instance($instanceId);
+    $database = $instance->database($databaseId);
+
+    $sql = "select 1::varchar as str, '2'::int as int, 3::decimal as dec, " .
+    "'4'::bytea as bytes, 5::float as float, 'true'::bool as bool, " .
+    "'2021-11-03T09:35:01UTC'::timestamptz as timestamp";
```
bshaffer, line 43: It's a nit, but multiline string styles should stay consistent

spanner/src/pg_batch_dml.php
```diff
@@ -0,0 +1,84 @@
...
+ * @param string $databaseId The Spanner database ID.
+ */
+function pg_batch_dml(string $instanceId, string $databaseId): void
+{
+    $spanner = new SpannerClient();
+    $instance = $spanner->instance($instanceId);
+    $database = $instance->database($databaseId);
+
+    $sql = 'INSERT INTO Singers (SingerId, FirstName, LastName)'
+    . ' VALUES ($1, $2, $3)';
```
bshaffer, line 44: here is another example

spanner/src/pg_case_sensitivity.php
```diff
@@ -0,0 +1,65 @@
...
+
+    $operation = $database->updateDdl(
+        sprintf('CREATE TABLE %s (
+            -- SingerId will be folded to "singerid"
+            SingerId  bigint NOT NULL PRIMARY KEY,
+            -- FirstName and LastName are double-quoted and will therefore retain their
+            -- mixed case and are case-sensitive. This means that any statement that
+            -- references any of these columns must use double quotes.
+            "FirstName" varchar(1024) NOT NULL,
+            "LastName"  varchar(1024) NOT NULL
```
bshaffer, line 51: Here's a third example

spanner/src/pg_information_schema.php
```diff
@@ -0,0 +1,79 @@
...
+
+    // Get all the user tables in the database. PostgreSQL uses the `public` schema for user
+    // tables. The table_catalog is equal to the database name.
+
+    $results = $database->execute('SELECT table_catalog, table_schema, table_name,'
+        . ' user_defined_type_catalog,'
+        . ' user_defined_type_schema,'
+        . ' user_defined_type_name'
+        . ' FROM INFORMATION_SCHEMA.tables'
+        . ' WHERE table_schema=\'public\'');
```
bshaffer, line 66: Here's a fourth example.

In this case, I'd prefer to see the statement created in a variable first and then executed. Or similar to your other SQL statements, simply defined in one multiline string.

spanner/test/spannerTest.php
```diff
@@ -22,16 +22,21 @@
...
+ * @retryDelayMethod exponentialBackoff
+ */
 class spannerTest extends TestCase
 {
     use TestTrait {
         TestTrait::runFunctionSnippet as traitRunFunctionSnippet;
     }
 
-    use RetryTrait, EventuallyConsistentTestTrait;
+    use RetryTrait, EventuallyConsistentTestTrait, ExponentialBackoffTrait;
```
bshaffer, line 39: You don't need to add the `ExponentialBackoffTrait` for this. The Exponential Backoff logic in this case is handled in t he `RetryTrait`

bshaffer (approved):
Minor nits, otherwise LGTM

PR #1618, opened 2022-04-21
feat(Storage): Dual-Region Sample

danielbankhead:
Closes https://github.com/GoogleCloudPlatform/php-docs-samples/issues/1617

### DIFF
--- a/storage/src/create_bucket_dual_region.php
+++ b/storage/src/create_bucket_dual_region.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Copyright 2021 Google Inc.
+ * Copyright 2022 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


storage/src/create_bucket_dual_region.php
```diff
@@ -0,0 +1,53 @@
...
+    // $bucketName = 'my-bucket';
+    // $location1 = 'US-EAST1';
+    // $location2 = 'US-WEST1';
+
+    $storage = new StorageClient();
+    $bucket = $storage->createBucket($bucketName, [
+        'location' => "${location1}+${location2}",
+    ]);
+
+    printf("Created dual-region bucket '%s' in %s+%s", $bucket->name(), $location1, $location2);
```
bshaffer, line 47: nit: do we want to surround the region with single quotes as we did the bucket name?
```suggestion
    printf("Created dual-region bucket '%s' in '%s+%s'", $bucket->name(), $location1, $location2);
```

storage/src/create_bucket_turbo_replication.php
```diff
@@ -31,7 +31,7 @@
  * The bucket must be a dual-region bucket for this setting to take effect.
  *
  * @param string $bucketName The name of your Cloud Storage bucket.
- * @param string $location The Dual Region location where you want your bucket to reside. (Read more at https://cloud.google.com/storage/docs/locations#location-dr)
+ * @param string $location The Dual-Region location where you want your bucket to reside. (Read more at https://cloud.google.com/storage/docs/locations#location-dr)
```
bshaffer, line 34: I'd like to see an example of the dual-region format here as well
```suggestion
 * @param string $location The Dual-Region location where you want your bucket to reside (e.g. "US-CENTRAL1+US-WEST1").
                                           Read more at https://cloud.google.com/storage/docs/locations#location-dr
```

bshaffer (commented):
This looks great! Some very minor suggestions only!

bshaffer (approved):

PR #1621, opened 2022-05-05
fix(deps): Updating google/cloud-compute to v1.0.2

rsamborski:
Updating google/cloud-compute to v1.0.2


bshaffer (approved):
FYI we don't typically update `composer.json` for patch version bumps

PR #1640, opened 2022-05-26
chore: upgrade monitoring to new sample format

bshaffer:


bshaffer:
The error coming from the REST-only tests ("Object of class Google\Cloud\Monitoring\V3\TimeInterval could not be converted to string") will be fixed in https://github.com/googleapis/gax-php/pull/430

monitoring/test/alertsTest.php
```diff
@@ -19,25 +19,24 @@
...
 use PHPUnit\Framework\TestCase;
 use PHPUnitRetry\RetryTrait;
 
 class alertsTest extends TestCase
 {
-    use ExecuteCommandTrait;
     use TestTrait;
     use RetryTrait;
 
-    private static $commandFile = __DIR__ . '/../alerts.php';
```
vishwarajanand, line 33: You need to remove it from `monitoring/phpunit.xml.dist` too

vishwarajanand: same for `monitoring.php`

vishwarajanand (approved):
LGTM, just minor nits

monitoring/test/alertsTest.php
bshaffer: good catch, done!

PR #1641, opened 2022-05-26
chore: upgrade texttospeech to new sample format

bshaffer:


texttospeech/src/synthesize_ssml.php
```diff
@@ -37,29 +31,36 @@
...
 
-/** Uncomment and populate these variables in your code */
-// $ssml = 'SSML to synthesize';
-
-// create client object
-$client = new TextToSpeechClient();
+/**
+ * @param string $ssml SSML to synthesize
+ */
+function synthesize_ssml(string $ssml)
```
saranshdhingra, line 37: Just a small question about no return type, I thought we should be adding those in the samples too 🤔

saranshdhingra (approved):
LGTM

texttospeech/src/synthesize_ssml.php
bshaffer: good idea! I missed that... will add them :)

PR #1643, opened 2022-05-26
chore: update logging to new samples format

bshaffer:


bshaffer:
moving snippet tags in cl/486749608

PR #1646, opened 2022-06-02
Create update_secret_with_alias.php

nataliesalvati:
Update samples for Access By Alias Launch


bshaffer:
This looks good but it needs a test!

nataliesalvati:
Think I've updated the test properly.

secretmanager/test/secretmanagerTest.php
```diff
@@ -288,3 +288,16 @@ public function testUpdateSecret()
         $this->assertStringContainsString('Updated secret', $output);
     }
 }
+public function testUpdateSecretWithAlias()
```
bshaffer, line 291: fix CS

```suggestion

    public function testUpdateSecretWithAlias()
```

secretmanager/src/update_secret_with_alias.php
```diff
@@ -0,0 +1,63 @@
...
+// Import the Secret Manager client library.
+use Google\Cloud\SecretManager\V1\Secret;
+use Google\Cloud\SecretManager\V1\SecretManagerServiceClient;
+use Google\Protobuf\FieldMask;
+
+/**
+ * @param string $projectId Your Google Cloud Project ID (e.g. 'my-project')
+ * @param string $secretId  Your secret ID (e.g. 'my-secret')
+ */
+function update_secret(string $projectId, string $secretId): void
```
bshaffer, line 38:
```suggestion
function update_secret_with_alias(string $projectId, string $secretId): void
```

bshaffer (approved):

PR #1651, opened 2022-06-14
Mark Laravel tutorial as deprecated

bshaffer:


ryokomy:
I'm one of those who had trouble in developing a Laravel application with CloudLogging.

This example works with Monolog v1.
But we have to use Monolog v2 with Laravel v8.
Then this tutorial doesn't work.

In those cases, the following documents are your go-to ones.
https://cloud.google.com/appengine/docs/standard/php-gen2/writing-application-logs 
https://cloud.google.com/logging/docs/setup/php#creating_a_psr-3_logger

For more info, I found that
- `psrBatchLogger` doesn't work
- formatters don't work

I'm happy if this will help you.

PR #1655, opened 2022-07-07
samples: create BigQuery subscription

kamalaboulhosn:


.kokoro/secrets-example.sh
```diff
@@ -102,6 +102,7 @@ export REDIS_PORT=
 # PubSub
 export GOOGLE_PUBSUB_SUBSCRIPTION=php-example-subscription
 export GOOGLE_PUBSUB_TOPIC=php-example-topic
+export GOOGLE_PUBSUB_BIGQUERY_TABLE=$GOOGLE_PROJECT_ID.pubsub_test_dataset.pubsub_test_table
```
saranshdhingra, line 105: @bshaffer Small query on how this will be populated?
The `$GOOGLE_PROJECT_ID` is a string if I understand correctly.

Apart from this LGTM.

kamalaboulhosn: One has to manually create the dataset and table. Having run the examples, it seems the same is true for php-example-topic and php-example-subscription

bshaffer: Let's just make this empty - since it's an example, it can just be removed

saranshdhingra (approved):
LGTM

pubsub/api/src/create_bigquery_subscription.php
```diff
@@ -0,0 +1,54 @@
...
+ * @param string $tableName  The BigQuery table to which to write.
+ */
+function create_bigquery_subscription($projectId, $topicName, $subscriptionName, $table)
+{
+    $pubsub = new PubSubClient([
+        'projectId' => $projectId,
+    ]);
+    $topic = $pubsub->topic($topicName);
+    $subscription = $topic->subscription($subscriptionName);
+    $config = new BigQueryConfig(['table' => $table]);
```
saranshdhingra, line 45: @kamalaboulhosn I had to do a minor change here. Just FYI.

pubsub/api/test/pubsubTest.php
```diff
@@ -270,6 +270,32 @@ public function testCreateAndDeletePushSubscription()
         $this->assertRegExp(sprintf('/%s/', $subscription), $output);
     }
 
+    public function testCreateAndDeleteBigQuerySubscription()
+    {
+        $topic = $this->requireEnv('GOOGLE_PUBSUB_TOPIC');
+        $subscription = 'test-subscription-' . rand();
+        $projectId = $this->requireEnv('GOOGLE_PROJECT_ID');
+        $table = $projectId . '.' . $this->requireEnv('GOOGLE_PUBSUB_BIGQUERY_TABLE');
```
saranshdhingra, line 278: Had to make minor changes here as we use alternate projects and therefore the project changes after the secrets have been assigned.

PR #1664, opened 2022-07-28
docs(secretmanager): Added sample for creating Secret with UserManaged replication

kapishps:
Added sample for creating Secret with UserManaged replication

Fixes b/222386112

### DIFF
--- a/secretmanager/src/create_secret_with_user_managed_replication.php
+++ b/secretmanager/src/create_secret_with_user_managed_replication.php
@@ -37,7 +37,7 @@ use Google\Cloud\SecretManager\V1\SecretManagerServiceClient;
  * @param string $secretId  Your secret ID (e.g. 'my-secret')
  * @param array  $locations Replication locations (e.g. array('us-east1', 'us-east4'))
  */
-function create_secret_with_user_managed_replication(string $projectId, string $secretId, ...$locations): void
+function create_secret_with_user_managed_replication(string $projectId, string $secretId, array $locations): void
 {
     // Create the Secret Manager client.
     $client = new SecretManagerServiceClient();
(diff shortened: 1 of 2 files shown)


secretmanager/src/create_secret_with_user_managed_replication.php
```diff
@@ -0,0 +1,72 @@
...
+        array_push($replicas, $replica);
+    }
+
+    $secret = new Secret([
+        'replication' => new Replication([
+            'user_managed' => new UserManaged(),
+        ]),
+    ]);
+
+    $secret->getReplication()->getUserManaged()->setReplicas($replicas);
```
bshaffer, line 61: Why not be consistent here?
```suggestion
    $secret = new Secret([
        'replication' => new Replication([
            'user_managed' => new UserManaged([
                'replicas' => $replicas
            ]),
        ]),
    ]);
```

secretmanager/src/create_secret_with_user_managed_replication.php
```diff
@@ -0,0 +1,72 @@
...
+
+    // Build the resource name of the parent project.
+    $parent = $client->projectName($projectId);
+
+    $replicas = array();
+    foreach ($locations as $location) {
+        $replica = new Replica();
+        $replica->setLocation($location);
+        array_push($replicas, $replica);
+    }
```
bshaffer, line 53: In PHP it's more idiomatic to use short array syntax and push syntax:
```suggestion
    $replicas = [];
    foreach ($locations as $location) {
        $replicas[] = new Replica(['location' => $location]);
    }
```

bshaffer (approved):

secretmanager/src/create_secret_with_user_managed_replication.php
kapishps: Thanks @bshaffer, I'm very new to php. I'll keep this in mind for later. :)

PR #1672, opened 2022-08-10
feat(PubSub): Added samples for Exactly Once Delivery

saranshdhingra:


pubsub/api/src/subscribe_exactly_once_delivery.php
```diff
@@ -0,0 +1,66 @@
...
+    $subscription = $pubsub->subscription($subscriptionId);
+    $messages = $subscription->pull();
+
+    foreach ($messages as $message) {
+        // When exactly once delivery is enabled on the subscription,
+        // the message is guaranteed to not be delivered again if the ack succeeds.
+        // Passing the `returnFailures` flag retries any temporary failures received
+        // while acking the msg and also returns any permanently failed msgs.
+        // Passing this flag on a subscription with exactly once delivery disabled
+        // will have no effect.
```
maheshgattani, line 52: Can we clarify "no effect"? Does that mean $failedMsgs will always return empty? If so, we should use that language.

maheshgattani (approved):

dwsupplee (approved):
These look great, nice work @saranshdhingra.

PR #1673, opened 2022-08-17
feat(Spanner): Add a default for versionTime in create_backup sample.

### LINKED ISSUE #1549: Spanner: Create Backup sample should have latest timestamp as a default.
Currently, we can pass `now` for the parameter `$versionTime` to achieve the backup until the current timestamp. The simplest way to make it default can be to use the string `now` as a default for `$versionTime`.

This however still means that we pass a `DateTime` object to Spanner. Perhaps we can also use `null` as a default, in which case we don't pass the `versionTime` parameter to the `$backup->create()` method, thereby letting Spanner take the call which will be equivalent to unchecking `Create backup from an earlier point in time` in the Cloud Console UI.

saranshdhingra:
The PR also adds a link for the user to understand different possible values for the `versionTime` argument.

Fixes #1549


spanner/src/create_backup.php
```diff
@@ -37,7 +37,8 @@
  * @param string $instanceId The Spanner instance ID.
  * @param string $databaseId The Spanner database ID.
  * @param string $backupId The Spanner backup ID.
- * @param string $versionTime The version of the database to backup.
+ * @param string $versionTime The version of the database to backup. Read more
+ * at https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances.backups#Backup.FIELDS.version_time
  */
 function create_backup(string $instanceId, string $databaseId, string $backupId, string $versionTime): void
```
vishwarajanand, line 43: The actual change got lost in merges, re-adding that now.

```suggestion
function create_backup(string $instanceId, string $databaseId, string $backupId, string $versionTime = '-1hour'): void
```

vishwarajanand (approved):

PR #1678, opened 2022-08-25
fix(functions): add 2nd-gen cloudevent tags

ace-n:


iennae (approved):

iennae:
@GoogleCloudPlatform/php-admins can y'all please review? This is required to fix sample browser generation for gen 1/2 Cloud Functions samples since PHP doesn't require duplicate samples for the different versions but in order for sample browser to show the sample in both places needs 2 region tags. 

Thanks.

PR #1680, opened 2022-08-29
feat: Laravel on Cloud Run

glasnt:
Code and README-based tutorial for Laravel on Cloud Run

Includes some region tags as highlighted in "Understanding the code". 

(View to possibly migrate README tutorial to DevSite, but if that happens I'll submit another PR to redirect readers to that version)


glasnt:
Unsure why linting is failing. 

```

Files that were not fixed due to errors reported during linting before fixing:
   1) /tmpfs/src/github/php-docs-samples/run/laravel/app/Providers/RouteServiceProvider.php

```

saranshdhingra:
@glasnt I believe that is happening due to the use of `?->` operator in the `RouteServiceProvider.php` file. This was introduced in PHP 8, while the linter is running an image containing PHP 7.4 [see this](https://github.com/benmatselby/sublime-phpcs/issues/140#issuecomment-132271390).

When I ran the fixer on your PR locally with PHP 7.3 in verbose mode I got the "Invalid file syntax" error.
On PHP 8.1 there were no changes.

Now, since this `RouteServiceProvider.php` is provided by the framework itself, we can either,
1) Ignore this and force merge.
2) Downgrade the Laravel version to be the one compatible to PHP 7.4.
3) Add an exclusion in the config file for php-cs-fixer.

@bshaffer Does this look right?

glasnt:
tests all pass now 🥳

run/laravel/README.md
```diff
@@ -0,0 +1,333 @@
...
+    gcloud secrets create laravel_settings --data-file .env
+    ```
+
+### Configure access to the secret
+
+* Allow Cloud Run access to the secret: 
+
+    ```bash
+    export PROJECTNUM=$(gcloud projects describe ${PROJECT_ID} --format='value(projectNumber)')
+    gcloud secrets add-iam-policy-binding laravel_settings \
```
bshaffer, line 207: nit: I like how all the other code lines are single commands, so what if we keep that paradigm going by splitting this up into two separate code blocks. Or, even better, we could move this up to the top block where we export all the variables together (right underneath where we export `PROJECT_ID`).

run/laravel/README.md
```diff
@@ -0,0 +1,333 @@
...
+    ```bash
+    gcloud artifacts repositories create containers \
+        --repository-format=docker \
+        --location=${REGION}
+    ```
+
+* Determine the registry name for future operations: 
+
+    ```bash
+    ${REGION}-docker.pkg.dev/${PROJECT_ID}/containers/
```
bshaffer, line 165: nit: since all the other code-fences are commands, I'd like to keep this one the same, to avoid confusion. Maybe something like
```suggestion
    export REGISTRY_NAME=${REGION}-docker.pkg.dev/${PROJECT_ID}/containers/
```

And then we can use it in the `gcloud beta jobs create` and `gcloud run deploy` commands later. If you'd prefer not to set an env var for later (I'm ambivalent), let's at least change this to be `echo`:

```suggestion
    echo ${REGION}-docker.pkg.dev/${PROJECT_ID}/containers/
```

run/laravel/README.md
```diff
@@ -0,0 +1,333 @@
...
+    ```bash
+    cp .env.example .env
+    ```
+
+* Update the values in `.env` with your values. 
+    
+    ⚠️ Replace `${}` with your values, don't use the literals.
+Get these values with e.g. `echo ${DATABASE_NAME}`
+    * DB_CONNECTION: `mysql`
+    * DB_SOCKET:  `/cloudsql/${PROJECT_ID}:${REGION}:${INSTANCE_NAME}`
```
bshaffer, line 181: nit, but in `.env.example`, `DB_SOCKET` was at the bottom. I think we should keep this consistent, whichever position is preferred.

run/laravel/README.md
```diff
@@ -0,0 +1,333 @@
...
+    * DB_DATABASE: `${DATABASE_NAME}`
+    * DB_USERNAME: `${DATABASE_USERNAME}`
+    * DB_PASSWORD: `${DATABASE_PASSWORD}`
+    * ASSET_BUCKET: `${ASSET_BUCKET}`
```
bshaffer, line 185: This is missing `ASSET_URL`, but I'm guessing that's because it's calculated from `ASSET_BUCKET`. Maybe add a comment in the `.env.example`  or here to clarify that point.

run/laravel/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+* Apply database migrations:
+
+    ```
+    gcloud beta run jobs create migrate-$(date +"%s") --execute-now --wait \
+        --image=${REGION}-docker.pkg.dev/${PROJECT_ID}/containers/laravel \
+        --set-cloudsql-instances ${PROJECT_ID}:${REGION}:${INSTANCE_NAME} \
+        --set-secrets /config/.env=laravel_settings:latest \
+        --command launcher \
+        --args "php artisan migrate"
```
bshaffer, line 238: This command as-is prompts the user for their `region`, and since we already have determined that above in our environment variables, let's go ahead and add ` --region $REGION` to the command so it executes without a prompt.

run/laravel/README.md
```diff
@@ -0,0 +1,333 @@
...
+### Deploy the service to Cloud Run
+
+1. Deploy the service from the previously created image, specifying the database connection and secret configuration: 
+
+    ```bash
+    gcloud run deploy laravel \
+        --image ${REGION}-docker.pkg.dev/${PROJECT_ID}/containers/laravel \
+        --set-cloudsql-instances ${PROJECT_ID}:${REGION}:${INSTANCE_NAME} \
+        --set-secrets /config/.env=laravel_settings:latest \
+        --allow-unauthenticated
```
bshaffer, line 271: Same here - this command prompts for a region, but if we add `--region $REGION` to the command, users won't have to deal with the prompt

run/laravel/README.md
```diff
@@ -0,0 +1,333 @@
...
+* Clone a copy of the code into your local machine; 
+
+    ```bash
+    git clone https://github.com/GoogleCloudPlatform/php-docs-samples.git
+    cd php-docs-samples/run/laravel/
+    ```
+
+## Confirm your PHP setup
+
+* Check you have PHP 8.1 installed (or [install it](https://www.php.net/manual/en/install.php)):
```
bshaffer, line 64: PHP 8.2 has been released, and also looking at `composer.json`, the version of Laravel we use allows PHP 8.0.2 and above. I suggest rewording this to be "PHP 8.0.2 or higher", as the version of PHP you're running locally doesn't matter.  In fact, since the tutorial never actually mentions running PHP in any way, we could exclude this requirement all together (along with `composer`), or at least put a note like "*if you'd like to run laravel locally*" so users know it's not required in this tutorial.

run/laravel/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Confirm deployment success
+
+
+
+1. Go to the Service URL to view the website.
+
+1. Confirm the information in the lower right of the Laravel welcome screen. 
+
+   * You should see a variation of "Laravel v9.26.1 (PHP v8.1.9)" (the exact version of Laravel and PHP may change)
```
bshaffer, line 282: I see `PHP 7.2.0`, which was just released this week. This is great, but is there a way to pin to a PHP version? I Most users wouldn't want their major version of PHP to update unexpectedly. If so I think that's a good idea (and we should pin to PHP 8.2, since it's the latest and greated), but we wouldn't want to require PHP 8.2 locally (since it's recently released and we don't want to make users upgrade their local version if they don't have to).

I made notes about requiring `PHP 8.0.2+` locally (and making a note that even this isn't _actually required_ to complete the tutorial) in the prerequisites above.  I think the best solution here would be to pin to PHP 8.2 for deployment but "require" PHP 8.0.2+ locally.

run/laravel/README.md
```diff
@@ -0,0 +1,333 @@
...
+While the initial provisioning and deployment steps were complex, making updates is a simpler process. 
+
+1. Build the container (to capture any new application changes), then update the service to use this new container image:
+
+    ```bash
+    gcloud builds submit \
+        --pack image=us-central1-docker.pkg.dev/${PROJECT_ID}/containers/laravel
+
+    gcloud run services update laravel \
+        --image us-central1-docker.pkg.dev/${PROJECT_ID}/containers/laravel
```
bshaffer, line 302: Here also! This command prompts for a region, but if we add --region $REGION to the command, users won't have to deal with the prompt

run/laravel/README.md
```diff
@@ -0,0 +1,333 @@
...
+* Check you have `composer` installed (or [install it](https://getcomposer.org/download/)): 
+
+    ```bash
+    composer --version
+    ```
+
+* Install the PHP dependencies: 
+
+    ```bash
+    composer install
```
bshaffer, line 79: for what it's worth, this command is not required to complete the tutorial, since this step is also performed during the deploy stage. 

What this does achieve though, would be pinning PHP 8.0 (or whatever version the user is running locally) compatible dependencies and deploying them to a PHP  8.2 runtime, which is not a great idea. This is because when the user runs `composer install` (and so generated a `composer.lock` file), this will pin _those_ dependencies for when the server deploys PHP 8.2.

bshaffer (commented):
Great tutorial!! I was able to get from beginning to end without any issues! I have a few comments related to minor tutorial based things, and only one major issue, which has to do with PHP versions (both required locally and deployed to).  Please let me know if you have any questions!

run/laravel/README.md
glasnt: The current reason to have php+composer is the `php artisan key:generate` command later to generate the `APP_KEY` in "Configuring the application". 

I can remove the `composer` dependency by getting the user to generate their own `APP_KEY` to put into the `.env` file with pure PHP: 

```shell
php -r "echo 'APP_KEY=base64:'.base64_encode(random_bytes(32));"
```

The user needs node+npm on their system to do the static asset compilation (because these aren't production deps and aren't in the container, the `gcloud run jobs` method can't be used). Given this dependency, the user could use node to generate the key instead: 

```shell
node -e "require('crypto').randomBytes(32,function(o,e){console.log('APP_KEY=base64:'+e.toString('base64'))});"
```

Or, without node or php, if we can presume `openssl` on a user's machine (or direct them to run in Cloud Shell): 

```shell
echo "APP_KEY=base64:$(openssl rand -base64 32)"
```

To run the application locally there needs to be some more information around changing the database configurations, but that is also possible. 

This would require: 

 * Changing the `DB_` configurations in `.env` to: `DB_CONNECTION=sqlite`, `DB_DATABASE=/path/to/database.sql`
 * Running `php artisan migrate`
 * Updating docs to emphasise that the local DB is different to the prod DB, don't exist data migration, etc.

run/laravel/README.md
glasnt: We can pin the PHP version in composer.json (which buildpacks should use ([code](https://github.com/GoogleCloudPlatform/buildpacks/blob/main/pkg/php/php.go#L250)), but I don't know when/if PHP 8.2 will be supported with Buildpacks, so we're limited there.

In this confirmation step, as long as I mention "PHP v8..." and "Laravel v9..." I think this will satisfy the checks without pinning

run/laravel/README.md
glasnt: See above on removing composer from the local stack. 

We can specify our minimum php in `composer.json` with [caret version ranges](https://getcomposer.org/doc/articles/versions.md#caret-version-range-)

Again, we're limited on what buildpacks has available, so as long as we specify an okay minimum we should be fine.

run/laravel/README.md
glasnt: Oh hey, it is supported! 🥳 

```
[builder] Installing PHP Runtime v8.2.0.
```

run/laravel/app/Http/helpers.php
```diff
@@ -0,0 +1,26 @@
...
+            return 'Unknown';
+        }
+
+        $metadata = new Google\Cloud\Core\Compute\Metadata();
+        $metadataValue = $metadata->get($metadataKey);
+
+        return $metadataValue;
+    }
+    // [END cloudrun_laravel_get_metadata]
+}
```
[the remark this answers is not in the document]
glasnt: Looks like I don't end up calling this code at all, and included all the logic in routes/web.php. This file has been removed.

run/laravel/README.md
bshaffer: Ahh right, I missed the key generation step! My bad. I think keeping composer and PHP locally for the `php artisan key:generate` step is okay with me, but I do think we should make the version restriction as wide as possible (e.g matching the laravel required version of PHP 8.0+) rather than saying we require 8.1 or 8.2!!

run/laravel/README.md
```diff
@@ -61,6 +61,8 @@ This tutorial uses the following billable components of Google Cloud:
 
 ## Confirm your PHP setup
 
+You will need PHP on your local system in order to generate run `php artisan` commands later.
```
bshaffer, line 64:
```suggestion
You will need PHP on your local system in order to run `php artisan` commands later.
```

run/laravel/Procfile
```diff
@@ -0,0 +1,3 @@
+web: pid1 --nginxBinaryPath nginx --nginxConfigPath /layers/google.php.webconfig/webconfig/nginx.conf --serverConfigPath /layers/google.php.webconfig/webconfig/nginxserver.conf --nginxErrLogFilePath /var/log/nginx.log --customAppCmd "php-fpm -R --nodaemonize --fpm-config /layers/google.php.webconfig/webconfig/php-fpm.conf" --pid1LogFilePath /var/log/pid1.log --mimeTypesPath /layers/google.utils.nginx/nginx/conf/mime.types --customAppSocket /layers/google.php.webconfig/webconfig/app.sock
```
glasnt, line 1: Just confirmed this setup vs the current buildpack web config, and it's still 'up to date', and the logs end up in Cloud Run successfully.

glasnt:
One noop comment, I'm happy with this being merged when you are.

bshaffer (approved):

PR #1685, opened 2022-09-04
feat(storage): sample for autoclass

vishwarajanand:
Addresses: https://github.com/GoogleCloudPlatform/php-docs-samples/issues/1682
(couldn't test in local, since feature not enabled in my gcp proj)

Design doc: [go/php-storage-autoclass](http://goto.google.com/php-storage-autoclass)
PR for python: https://github.com/googleapis/python-storage/pull/791


storage/src/create_bucket_autoclass.php
```diff
@@ -0,0 +1,57 @@
...
+ * Create a new bucket with autoclass enabled.
+ *
+ * @param string $bucketName The name of your Cloud Storage bucket.
+ */
+function create_bucket_autoclass($bucketName)
+{
+    // $bucketName = 'my-bucket';
+
+    $storage = new StorageClient();
+    $location = 'ASIA';
```
vishwarajanand, line 40: May need to change this, since the feature is geo restricted in Oct. Will check and confirm.

storage/src/create_bucket_autoclass.php
```diff
@@ -0,0 +1,58 @@
...
+# [START storage_create_autoclass]
+
+use Google\Cloud\Storage\StorageClient;
+
+/**
+ * Create a new bucket with autoclass enabled.
+ *
+ * @param string $bucketName The name of your Cloud Storage bucket.
+ */
+function create_bucket_autoclass($bucketName)
```
saranshdhingra, line 35:
```suggestion
function create_bucket_autoclass(string $bucketName): void
{
```

storage/src/create_bucket_autoclass.php
```diff
@@ -0,0 +1,58 @@
...
+{
+    // $bucketName = 'my-bucket';
```
saranshdhingra, line 37: Super nit, but I think when the arguments don't have a default, let's not include these comments.

Even if the user uncomments these, they still won't be able to run these samples without modifying the code to either remove the argument from the function definition, or add a default.

storage/src/get_bucket_autoclass.php
```diff
@@ -0,0 +1,62 @@
...
+# [START storage_get_autoclass]
+
+use Google\Cloud\Storage\StorageClient;
+
+/**
+ * Print a bucket autoclass configuration.
+ *
+ * @param string $bucketName The name of your Cloud Storage bucket.
+ */
+function get_bucket_autoclass($bucketName)
```
saranshdhingra, line 35: Func definition changes like sample 1.

storage/src/set_bucket_autoclass.php
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Updates an existing bucket with provided autoclass toggle.
+ *
+ * @param string $bucketName The name of your Cloud Storage bucket.
+ * @param bool $toggle If true, enables Autoclass. If false, disables Autoclas.
+ */
+function set_bucket_autoclass($bucketName, $toggle)
+{
+    // $bucketName = 'my-bucket';
+    // $toggle = false;
```
saranshdhingra, line 39: Remove this if not needed, or add a default in the function definition.
If you end up adding a default, then probably `$autoclassStatus` or `$status` would be better suited?

storage/test/storageTest.php
```diff
@@ -757,6 +757,62 @@ public function testChangeDefaultStorageClass()
         );
     }
 
+    public function testCreateBucketWithAutoclass()
+    {
+        $bucketName = uniqid('samples-create-autoclass-');
+
+        $output = self::runFunctionSnippet('create_bucket_autoclass', [
```
saranshdhingra, line 764: If there are indeed 3 samples, I think they should be called separately in separate tests.

Also, to verify `create_ bucket_autoclass` we don't really need to call `get_bucket_autoclass`.

The test should only be the existence of `Created bucket %s in %s with autoclass enabled` in the `$output`.

storage/src/create_bucket_autoclass.php
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/storage/README.md
+ */
+
+namespace Google\Cloud\Samples\Storage;
+
+# [START storage_create_autoclass]
```
saranshdhingra, line 26: I didn't find this region tag in the internal doc or the github issue attached. Do we need this sample?

saranshdhingra (changes requested):

storage/src/create_bucket_autoclass.php
bshaffer: +1

storage/test/storageTest.php
vishwarajanand: @cojenco confirmed with tech writing team that create sample isn't needed. So I removed them.

storage/src/create_bucket_autoclass.php
vishwarajanand: Removed now after confirmation from tech writing team.

storage/src/create_bucket_autoclass.php
vishwarajanand: Nide suggestion, I am curious why don't we follow this convention across samples?
For now, I am inclined to keep it this way. Because adding a default value will be different from other samples, so it might look odd one out on cdc page. Also it will print an exception for users who re-run the sample, without providing the actual way to run it.

But I am open to discuss and convert all samples to the default param values if needed.

storage/src/get_bucket_autoclass.php
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/storage/README.md
+ */
+
+namespace Google\Cloud\Samples\Storage;
+
+# [START storage_get_autoclass]
+
```
bshaffer, line 27: nit: remove extra newline
```suggestion
```

storage/src/get_bucket_autoclass.php
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Print a bucket autoclass configuration.
+ *
+ * @param string $bucketName The name of your Cloud Storage bucket.
+ */
+function get_bucket_autoclass(string $bucketName): void
+{
+    // $bucketName = 'my-bucket';
+
```
bshaffer, line 38: I would rather put. this in the parameter description, for example:

```
 * @param string $bucketName The name of your Cloud Storage bucket (e.g. 'my-bucket')
```

But I am not super opinionated on it!

storage/src/set_bucket_autoclass.php
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/storage/README.md
+ */
+
+namespace Google\Cloud\Samples\Storage;
+
+# [START storage_set_autoclass]
+
```
bshaffer, line 27:
```suggestion
```

storage/test/storageTest.php
```diff
@@ -757,6 +757,49 @@ public function testChangeDefaultStorageClass()
...
+            $bucket->name(),
+            true
+        ]);
+        $bucket->delete();
+
+        $this->assertStringContainsString(
+            sprintf(
+                'Updated bucket %s with autoclass enabled',
+                $bucket->name(),
+                true
```
bshaffer, line 797: I think the `true` is here by accident
```suggestion
                $bucket->name()
```

storage/src/set_bucket_autoclass.php
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Updates an existing bucket with provided autoclass toggle.
+ *
+ * @param string $bucketName The name of your Cloud Storage bucket.
+ * @param bool $autoclassStatus If true, enables Autoclass. Disables otherwise.
+ */
+function set_bucket_autoclass(string $bucketName, bool $autoclassStatus): void
+{
+    // $bucketName = 'my-bucket';
+    // $autoclassStatus = true;
```
cojenco, line 39: Could we instead **disable** Autoclass in this sample with `$autoclassStatus = false`?
As previously discussed, users can disable Autoclass on an existing bucket upon our 1st launch in Oct (unfortunately not the same case for enabling the feature on an existing bucket...) 

For a better user experience, I'd suggest to set bucket autoclass to false in this sample. Otherwise, looking good. Thanks @vishwarajanand for adding these samples.

vishwarajanand: I changed the comment to false and also in tests, I reflected this change.

storage/src/get_bucket_autoclass.php
vishwarajanand: Very similar to Saransh's comment https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1685#discussion_r974079050, do you think we should remove comments like this from all samples to instead have them as default params or as `(e.g. '<value>')`.

I am hesitating because all other samples are written with such comments and a PHP user will be confused while reading samples on cgc page. Thoughts?

storage/src/create_bucket_autoclass.php
saranshdhingra: Just to clarify, wasn't asking to put a default, I was just saying that uncommenting this line doesn't help for the user.

Regarding why other samples have it, we did have an old format for these samples before, where we didn't used to wrap everything in a function. Probably left overs from them.

I'd say let's not include this for any new samples we do and let's create a task to remove these from the existing samples as well.

vishwarajanand: Added a task to clear samples with commented values. b/249236170

vishwarajanand:
Update: To be merged in the first week of Nov.

storage/src/set_bucket_autoclass.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+namespace Google\Cloud\Samples\Storage;
+
+# [START storage_set_autoclass]
+use Google\Cloud\Storage\StorageClient;
+
+/**
+ * Updates an existing bucket with provided autoclass toggle.
+ *
```
cojenco, line 31: Could we add a comment here to emphasize that currently v1 Autoclass can only be enabled upon bucket creation? Otherwise, LGTM

```
# Note: Only patch requests that disable autoclass are currently supported.
# To enable autoclass, you must set it at bucket creation time.
```

cojenco (approved):
Thanks for working on the samples. Deferring to @saranshdhingra for further review

storage/src/set_bucket_autoclass.php
vishwarajanand: Done :-)

saranshdhingra (approved):

PR #1691, opened 2022-09-27
feat(Spanner): Run backup tests only when requested explicitly

saranshdhingra:
This aims to run the Backup tests only when requested. That will be recognised by the presence of a label `spanner:run-backup-tests` on the PR.

If the label isn't present, the backup tests will be skipped.

Otherwise they will be run.

This will heavily help in the frequent quota errors we are seeing in spanner related PRs.

### DIFF
--- a/spanner/test/spannerBackupTest.php
+++ b/spanner/test/spannerBackupTest.php
@@ -72,6 +72,9 @@ class spannerBackupTest extends TestCase
         if (!extension_loaded('grpc')) {
             self::markTestSkipped('Must enable grpc extension.');
         }
+        if ('true' !== getenv('GOOGLE_SPANNER_RUN_BACKUP_TESTS')) {
+            self::markTestSkipped('Skipping backup tests.');
+        }
         self::$instanceId = self::requireEnv('GOOGLE_SPANNER_INSTANCE_ID');
 
         $spanner = new SpannerClient([
--- a/testing/run_test_suite.sh
+++ b/testing/run_test_suite.sh
@@ -85,12 +85,19 @@ FILES_CHANGED=$(git diff --name-only HEAD $(git merge-base HEAD master))
 if [ -z "$PULL_REQUEST_NUMBER" ]; then
     RUN_ALL_TESTS=1
 else
+    labels=$(curl "https://api.github.com/repos/GoogleCloudPlatform/php-docs-samples/issues/$PULL_REQUEST_NUMBER/labels")
+    
     # Check to see if the repo includes the "kokoro:run-all" label
-    if curl "https://api.github.com/repos/GoogleCloudPlatform/php-docs-samples/issues/$PULL_REQUEST_NUMBER/labels" \
-        | grep -q "kokoro:run-all"; then
+    if  grep -q "kokoro:run-all" <<< $labels; then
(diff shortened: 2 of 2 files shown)


testing/run_test_suite.sh
```diff
@@ -107,6 +118,11 @@ if ! type $TESTCMD > /dev/null; then
   exit 1
 fi
 
+# If we intend to run the backup tests in Spanner, we pass the env variable
+if [ $GOOGLE_SPANNER_RUN_BACKUP_TESTS == 1 ]; then
+    TESTCMD="export GOOGLE_SPANNER_RUN_BACKUP_TESTS=true && $TESTCMD"
+fi
+
```
bshaffer: We don't need these lines either, just run `export GOOGLE_SPANNER_RUN_BACKUP_TESTS=1` above and this will work as expected (I've made a code suggestion)
```suggestion
```

testing/run_test_suite.sh
```diff
@@ -80,19 +80,30 @@ FAILED_FLAKY_FILE=${TMP_REPORT_DIR}/failed_flaky
...
         RUN_ALL_TESTS=1
     else
         RUN_ALL_TESTS=0
     fi
 
+    # Check to see if the repo includes the "spanner:run-backup-tests" label
+    if grep -q "spanner:run-backup-tests" <<< $labels; then
+        GOOGLE_SPANNER_RUN_BACKUP_TESTS=1
+    else
+        GOOGLE_SPANNER_RUN_BACKUP_TESTS=0
```
bshaffer: You can remove the `else`... no need to export the variable if we don't want to run the tests. Also, use `export` so that the `TESTCMD` will read the envvar
```suggestion
        export GOOGLE_SPANNER_RUN_BACKUP_TESTS=1
```

spanner/test/spannerBackupTest.php
```diff
@@ -72,6 +72,9 @@ public static function setUpBeforeClass(): void
         if (!extension_loaded('grpc')) {
             self::markTestSkipped('Must enable grpc extension.');
         }
+        if (!getenv('GOOGLE_SPANNER_RUN_BACKUP_TESTS')) {
```
bshaffer, line 75: nit: let's make this explicit (because `GOOGLE_SPANNER_RUN_BACKUP_TESTS=false` would still run the tests, for example). Also, our other env vars like this are explicit to either be "true" or "1":
```suggestion
        if ('true' !== getenv('GOOGLE_SPANNER_RUN_BACKUP_TESTS')) {
```

testing/run_test_suite.sh
```diff
@@ -80,19 +80,29 @@ FAILED_FLAKY_FILE=${TMP_REPORT_DIR}/failed_flaky
 # (will be empty if running from "master").
 FILES_CHANGED=$(git diff --name-only HEAD $(git merge-base HEAD master))
 
+# Spanner Backup tests will be skipped if this env variable is unset
+GOOGLE_SPANNER_RUN_BACKUP_TESTS=0
```
bshaffer, line 84: I have another comment on this, but the diff moved so I'll say it again. There's no need to add this.
```suggestion
```

bshaffer (approved):

PR #1692, opened 2022-09-28
[Storage][Sample] print_bucket_default_acl

yash30201:
Wrote sample and unit test for fetching and printing bucket's default ACL.

### DIFF
--- a/storage/src/print_bucket_default_acl.php
+++ b/storage/src/print_bucket_default_acl.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * Copyright 2022 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/storage/README.md
+ */
+
+namespace Google\Cloud\Samples\Storage;
+
+# [START storage_print_bucket_default_acl]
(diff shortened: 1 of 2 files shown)


storage/src/print_bucket_default_acl.php
```diff
@@ -0,0 +1,48 @@
...
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/master/storage/README.md
+ */
+
+namespace Google\Cloud\Samples\Storage;
+
+# [START storage_print_bucket_default_acl]
+use Google\Cloud\Storage\StorageClient;
+
+/**
+ * Print all entities and roles for a bucket's Default ACL.
```
saranshdhingra, line 30:
```suggestion
 * Print the default object ACL for a bucket.
 * Read more at https://cloud.google.com/storage/docs/access-control/create-manage-lists#defaultobjects
```

saranshdhingra (dismissed):
Small change, otherwise LGTM.

saranshdhingra:
@yash30201 
So, the [doc](https://cloud.google.com/storage/docs/access-control/create-manage-lists#defaultobjects) has a sample for C# to "view" the default ACL for the object.
While the other languages have a sample to "add" the default ACL.

Before merging please contact the Storage owners to understand why that is the case.

storage/src/print_bucket_default_acl.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @param string $bucketName The name of your Cloud Storage bucket.
+ */
+function print_bucket_default_acl($bucketName)
```
saranshdhingra, line 34:
```suggestion
function print_bucket_default_acl(string $bucketName): void {
```

Use [this](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/storage/src/create_bucket_notifications.php#L37-L40) as a reference. Same for the other 2 PRs.

yash30201:
> @yash30201 So, the [doc](https://cloud.google.com/storage/docs/access-control/create-manage-lists#defaultobjects) has a sample for C# to "view" the default ACL for the object. While the other languages have a sample to "add" the default ACL.
> 
> Before merging please contact the Storage owners to understand why that is the case.

Thanks for pointing this out. I previously did discover this and found that this sample is for adding more roles to the default ACL so that this modified ACL becomes the new default one. It serves a different purpose than just printing the default ACL and isn't related to this sample. Moreover, I think that the error on that C# sample is wrong region tag(as that sample seems to be just doing the work of this sample instead of that one). I also discovered some issues like this in PHP samples, plan to rectify them later.

PR #1709, opened 2022-10-12
feat(Bigquery): Append column samples

yash30201:
Wrote and tested two samples - 
* Add a column using a load job
* Add a column using a query job


vishwarajanand (approved):
LGTM

PR #1715, opened 2022-10-26
chore(storage): Remove commented example value for arguments in Storage samples

ajupazhamayil:
This PR is related to the comment https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1685#discussion_r974079050


ajupazhamayil:
Total number of files in the src of Storage = 110
Total number of files changed = 106

Four unchanged files are:

1. generate_encryption_key.php
2. list_buckets.php
3. print_bucket_default_acl.php
4. print_bucket_website_configuration.php

ajupazhamayil:
This PR used a script in the repository: https://github.com/ajupazhamayil/function_default_comment_remover

storage/src/remove_bucket_conditional_iam_binding.php
```diff
@@ -40,12 +40,6 @@
  */
 function remove_bucket_conditional_iam_binding(string $bucketName, string $role, string $title, string $description, string $expression): void
 {
-    // $bucketName = 'my-bucket';
-    // $role = 'roles/storage.objectViewer';
-    // $title = 'Title';
-    // $description = 'Condition Description';
-    // $expression = 'resource.name.startsWith("projects/_/buckets/bucket-name/objects/prefix-a-")';
```
bshaffer, line 47: Personally I think these are helpful as they are not hurting anything. I know the previous maintainers felt strongly about having example values supplied _somewhere_, so if not here, we should put them somewhere else.

storage/src/add_bucket_acl.php
```diff
@@ -35,10 +35,6 @@
  */
 function add_bucket_acl(string $bucketName, string $entity, string $role): void
 {
-    // $bucketName = 'my-bucket';
```
vishwarajanand, line 38: The docblock before the function is supposed to capture the suggested value at the end. Such as:
`(e.g. 'my-bucket')`

vishwarajanand (changes requested):
Need to change the docblocks accordingly.

storage/src/add_bucket_acl.php
ajupazhamayil: Will do this way!

storage/src/remove_bucket_conditional_iam_binding.php
ajupazhamayil: @bshaffer Thank you for the suggestion, As @vishwarajanand suggested, we can add it to the parameter comment section itself. Example:

```
 /*
 * @param string $projectId The ID of your Google Cloud Platform project (e.g. 'my-project-id').
 * @param string $accessId Access ID for an inactive HMAC key (e.g. 'GOOG0234230X00').
 */
```

ajupazhamayil:
Great, will change according to the comments, moved the PR to draft, will make it open after the changes, thank you all!

storage/src/remove_bucket_conditional_iam_binding.php
```diff
@@ -32,20 +32,14 @@
...
- * @param string $bucketName The name of your Cloud Storage bucket.
- * @param string $role the role that will be given to members in this binding.
- * @param string $title The title of the condition.
- * @param string $description The description of the condition.
- * @param string $expression Te condition specified in CEL expression language.
+ * @param string $bucketName The name of your Cloud Storage bucket. (e.g. 'my-bucket')
+ * @param string $role the role that will be given to members in this binding. (e.g. 'roles/storage.objectViewer')
+ * @param string $title The title of the condition. (e.g. 'Title')
+ * @param string $description The description of the condition. (e.g. 'Condition Description')
+ * @param string $expression Te condition specified in CEL expression language. (e.g. 'resource.name.startsWith("projects/_/buckets/bucket-name/objects/prefix-a-")')
```
vishwarajanand, line 39: Too long, break into two sentences?

vishwarajanand: Usually we try to break sentences once it's way more than 80 chars.

storage/src/rotate_encryption_key.php
```diff
@@ -29,25 +29,20 @@
...
- * @param string $objectName The name of your Cloud Storage object.
- * @param string $oldBase64EncryptionKey The Base64 encoded AES-256 encryption
+ * @param string $bucketName The name of your Cloud Storage bucket. (e.g. 'my-bucket')
+ * @param string $objectName The name of your Cloud Storage object. (e.g. 'my-object')
+ * @param string $oldBase64EncryptionKey The Base64 encoded AES-256 encryption (e.g. 'TIbv/fjexq+VmtXzAlc63J4z5kFmWJ6NdAPQulQBT7g=')
  *     key originally used to encrypt the object. See the documentation on
  *     Customer-Supplied Encryption keys for more info:
  *     https://cloud.google.com/storage/docs/encryption/using-customer-supplied-keys
- * @param string $newBase64EncryptionKey The new base64 encoded encryption key.
+ * @param string $newBase64EncryptionKey The new base64 encoded encryption key. (e.g. '0mMWhFvQOdS4AmxRpo8SJxXn5MjFhbz7DkKBUdUIef8=')
```
vishwarajanand, line 38: This may also need to be broken into multiple lines like this:

```suggestion
 * @param string $newBase64EncryptionKey The new base64 encoded encryption key.
          (e.g. '0mMWhFvQOdS4AmxRpo8SJxXn5MjFhbz7DkKBUdUIef8=')
```

storage/src/remove_bucket_conditional_iam_binding.php
vishwarajanand: May need a change here: [ref](https://github.com/ajupazhamayil/function_param_example_comment_remover/blob/main/function_param_example_comment_remover.py#L52)

vishwarajanand (approved):
LGTM

PR #1717, opened 2022-11-04
Use binary-only images for installing Composer

Wirone:
See: https://blog.codito.dev/2022/11/composer-binary-only-docker-images/


bshaffer (approved):

bshaffer:
This looks good but you will need to sign the [Google CLA](https://cla.developers.google.com/) before we can merge this

Wirone:
@bshaffer I already did this, right after initial fail in CI 🙂 Check was refreshed and is green.

Issue #1704, opened 2022-10-06
Bigquery: Update samples format.


saranshdhingra:
Update the samples to use the new format like in other products.
[Reference](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/bigtable/src/create_table.php)

vishwarajanand:
May refer to https://github.com/ajupazhamayil/function_param_example_comment_remover/

yash30201:
Fixed #1720

PR #1720, opened 2022-11-10
fix(#1704): Refractored all samples as functions

yash30201:
Refractored all samples as functions to stay consistent with other products.

Fixed [#1704]

### DIFF
--- a/bigquery/api/src/stream_row.php
+++ b/bigquery/api/src/stream_row.php
@@ -23,45 +23,49 @@
 
 namespace Google\Cloud\Samples\BigQuery;
 
-// Include Google Cloud dependendencies using Composer
-require_once __DIR__ . '/../vendor/autoload.php';
-
-if (count($argv) < 4 || count($argv) > 5) {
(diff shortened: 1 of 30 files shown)


bigquery/api/src/stream_row.php
```diff
@@ -23,45 +23,49 @@
...
-]);
-$dataset = $bigQuery->dataset($datasetId);
-$table = $dataset->table($tableId);
+/**
+ * Stream data into bigquery
+ *
+ * @param string $projectId The project Id of your Google Cloud Project.
+ * @param string $datasetId The BigQuery dataset ID.
+ * @param string $tableId The BigQuery table ID.
+ * @param array $data, Json encoded data For eg,
```
bshaffer, line 35: typo
```suggestion
 * @param array $data Json encoded data For eg,
```

bshaffer (approved):
Wow, this is great work! I'm so glad you took the time to do this. The new format looks way better. 

One tiny typo, otherwise this is approved.

PR #1722, opened 2022-11-14
nit: formatting fixes to trigger firebase rtdb

vishwarajanand:
This PR triggers tests in firebase rtdb.


vishwarajanand:
Teh build was successful and tests were triggered, so no action is need on this PR.

PR #1746, opened 2022-12-12
feat(analyticsdata): add samples for batch and pivot reports

anweshan:


bshaffer (changes requested):
This does not seem to be ready for review. Please let us know when this is updated with the same changes as we've discussed with the other samples!

bshaffer (approved):

PR #1759, opened 2022-12-29
feat: [LiveStream] add samples and tests

irataxy:
b:241866780

Part 1 of 3: Inputs (see https://cloud.google.com/livestream/docs/how-to/create-input-endpoints)

### DIFF
--- a/media/livestream/src/update_input.php
+++ b/media/livestream/src/update_input.php
@@ -56,7 +56,7 @@ function update_input(
     $input->setPreprocessingConfig($config);
 
     $updateMask = new FieldMask([
-        'paths' => ['preprocessing_configx']
+        'paths' => ['preprocessing_config']
     ]);
 
     try {
--- a/media/livestream/src/update_input.php
+++ b/media/livestream/src/update_input.php
@@ -46,14 +46,14 @@ function update_input(
     $livestreamClient = new LivestreamServiceClient();
 
     $formattedName = $livestreamClient->inputName($callingProjectId, $location, $inputId);
-    $input = new Input();
-    $input->setName($formattedName);
     $crop = (new PreprocessingConfig\Crop())
         ->setTopPixels(5)
         ->setBottomPixels(5);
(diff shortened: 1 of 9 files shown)


media/livestream/src/update_input.php
```diff
@@ -0,0 +1,82 @@
...
+        if ($operationResponse->operationSucceeded()) {
+            $result = $operationResponse->getResult();
+            // Print results
+            printf('Updated input: %s' . PHP_EOL, $result->getName());
+        } else {
+            $error = $operationResponse->getError();
+            // handleError($error)
+        }
+    } finally {
+        $livestreamClient->close();
```
bshaffer, line 75: PHP requests are typically short-lived, so we usually don't include the `try` / `finally`/ call to `close()` in our samples. But if you'd prefer to keep this in, that's okay also.

media/livestream/src/update_input.php
```diff
@@ -0,0 +1,82 @@
...
+
+    $formattedName = $livestreamClient->inputName($callingProjectId, $location, $inputId);
+    $input = new Input();
+    $input->setName($formattedName);
+    $crop = (new PreprocessingConfig\Crop())
+        ->setTopPixels(5)
+        ->setBottomPixels(5);
+    $config = (new PreprocessingConfig())
+        ->setCrop($crop);
+    $input->setPreprocessingConfig($config);
```
bshaffer, line 56:
```suggestion
    $crop = (new PreprocessingConfig\Crop())
        ->setTopPixels(5)
        ->setBottomPixels(5);
    $config = (new PreprocessingConfig())
        ->setCrop($crop);
    $input = (new Input())
        ->setName($formattedName)
        ->setPreprocessingConfig($config);
```

media/livestream/src/update_input.php
irataxy: Removed

bshaffer (approved):

PR #1765, opened 2023-01-10
feat: [LiveStream] add Channel event samples and tests

irataxy:
b:241866780

Part 3 of 3: Channel events (see https://cloud.google.com/livestream/docs/how-to/create-channel-events)

### DIFF
--- a/media/livestream/composer.json
+++ b/media/livestream/composer.json
@@ -2,6 +2,7 @@
     "name": "google/live-stream-sample",
     "type": "project",
     "require": {
-        "google/cloud-video-live-stream": "^0.2.4"
+        "google/cloud-video-live-stream": "^0.2.4",
+        "grpc/grpc": "^1.38"
     }
 }
--- a/media/livestream/src/create_channel_event.php
+++ b/media/livestream/src/create_channel_event.php
@@ -1,0 +1,69 @@
+<?php
+
+/**
+ * Copyright 2022 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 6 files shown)


media/livestream/composer.json
```diff
@@ -2,6 +2,7 @@
     "name": "google/live-stream-sample",
     "type": "project",
     "require": {
-        "google/cloud-video-live-stream": "^0.2.4"
+        "google/cloud-video-live-stream": "^0.2.4",
+        "grpc/grpc": "^1.38"
```
bshaffer, line 6: This already comes installed as part of [google/gax](https://github.com/googleapis/gax-php/blob/main/composer.json#L12). Is the reason that LiveStream needs something in that specific version (1.38)? If so, we should make that the minimum version in the livestream client.

If this is here because the RPCs require grpc to be enabled, then you need to require the grpc extension, like [Spanner does](https://github.com/googleapis/google-cloud-php/blob/main/Spanner/composer.json#L7)

media/livestream/src/create_channel_event.php
```diff
@@ -0,0 +1,69 @@
...
+
+    $parent = $livestreamClient->channelName($callingProjectId, $location, $channelId);
+
+    $eventAdBreak = new Event\AdBreakTask([
+        'duration' => new Duration(['seconds' => 30])
+    ]);
+    $event = new Event([
+        'ad_break' => $eventAdBreak,
+        'execute_now' => true
+    ]);
```
bshaffer, line 58: this is subjective (again), but to be consistent we can use the fluid interfaces for the first level of message creation:
```suggestion
    $eventAdBreak = (new Event\AdBreakTask())
        ->setDuration(new Duration(['seconds' => 30]));
    $event = (new Event())
        ->setAdBreak($eventAdBreak)
        ->setExecuteNow(true);
```
You can ignore this suggestion though, if you prefer the other way.

bshaffer (approved):

PR #1767, opened 2023-01-17
feat: [DocumentAI] add quickstart sample

holtskinner:
NOTE: The test project will need the following API enabled

- `documentai.googleapis.com`

And a Document OCR Processor will need to be created in the project

https://cloud.google.com/document-ai/docs/create-processor

And the Processor ID should replace the `aaaaaaaaa` in the quickstartTest.php file


.gitignore
```diff
@@ -11,4 +11,5 @@ credentials.*
 .php-cs-fixer.cache
 .vscode/
 .kokoro/secrets.sh
-.phpunit.result.cache
+.phpunit.result.cache]
```
bshaffer, line 14: this is a typo... please fix it!

documentai/quickstart.php
```diff
@@ -0,0 +1,59 @@
...
+$name = $client->processorName($projectId, $location, $processor);
+
+# Make Processing Request
+$response = $client->processDocument($name, [
+    'raw_document' => $rawDocument
+]);
+
+# Print Document Text
+printf('Document Text');
+printf($response->document->text);
```
bshaffer, line 57: nit: this is more idiomatic:

```suggestion
printf('Document Text: %s', $response->document->text);
```

documentai/quickstart.php
```diff
@@ -0,0 +1,59 @@
...
+
+# Load File Contents into RawDocument
+$rawDocument = new RawDocument([
+    'content' => $contents,
+    'mime_type' => 'application/pdf'
+]);
+
+$projectId = 'YOUR_PROJECT_ID'; # Your Google Cloud Platform project ID
+$location = 'us'; # Your Processor Location
+$processor = 'YOUR_PROCESSOR_ID'; # Your Processor ID
```
bshaffer, line 45: We put placeholders at the top  (in this case, just below the `use` imports) so that they're easy to see when looking at the quickstart

documentai/test/quickstartTest.php
```diff
@@ -0,0 +1,43 @@
...
+            $contents
+        );
+        file_put_contents($file, $contents);
+
+        // Invoke quickstart.php
+        ob_start();
+        $output = ob_get_clean();
+
+        // Make sure it looks correct
+        $this->expectOutputString('Document');
```
bshaffer, line 41: Let's actually check for text that's in the PDF.

bshaffer (changes requested):

.gitignore
holtskinner: Fixed, thanks!

documentai/quickstart.php
holtskinner: Ok, I thought that would make sense, but I didn't see that in other samples. I'll change it up

documentai/test/quickstartTest.php
holtskinner: Ok

documentai/phpunit.xml.dist
```diff
@@ -0,0 +1,38 @@
...
+        <testsuite name="PHP documentai test">
+            <directory>test</directory>
+        </testsuite>
+    </testsuites>
+    <logging>
+        <log type="coverage-clover" target="build/logs/clover.xml"/>
+    </logging>
+    <filter>
+        <whitelist>
+            <directory suffix=".php">./src</directory>
```
bshaffer, line 28: nit but, `src` doesn't exist. If we think more samples are coming, then this is okay

bshaffer (approved):
looks good!

documentai/phpunit.xml.dist
holtskinner: It's possible, leaving options open

holtskinner:
@bshaffer Thanks for the review! Can you make the configuration changes to the kokoro project as listed in https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1767#issue-1536697245 
?

PR #1768, opened 2023-01-20
fix extracted docs page for Cloud Run

justinmahood:
make sure <?php directive gets included in extracted docs page

### DIFF
--- a/run/helloworld/index.php
+++ b/run/helloworld/index.php
@@ -1,4 +1,3 @@
-<?php
 /**
  * Copyright 2020 Google LLC.
  *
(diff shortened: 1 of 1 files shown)


bshaffer:
@justinmahood it's best practice in the PHP community to not include the open `<?php` tags in sample code, as this would become redundant across all our samples. However, in this case, if it's in the context where we say "copy all of this into your own file", I think you're right that we should include it. Is that the case here?

justinmahood:
@bshaffer yes.    I made the change made in response to feedback that a Cloud Run quickstart was broken.

The contents of this file are surfaced in the [Cloud Run PHP Quickstart](https://cloud.google.com/run/docs/quickstarts/build-and-deploy/deploy-php-service#writing).  
Screenshot 2023-01-23 10 33 24 AM

PR #1770, opened 2023-01-24
chore: add PHP codesniffer and fix errors/warnings

bshaffer:
Add linter for PHPCS, which checks for line lengths and found multiple other whitespace issues.

PHP CS Fixer (the existing checker) does not do a good job with whitespace (and in some cases, doesn't do the job at all). For this reason, it's better to run both linters.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   styles:
+    name: PHP CS Fixer
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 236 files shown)


rsamborski (changes requested):
Approved for compute and security products. Before merging please update [CONTRIBUTING.md](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/main/CONTRIBUTING.md#style) to explain how to run the sniffer locally before submitting changes.

.github/workflows/lint.yml
```diff
@@ -18,7 +19,26 @@ jobs:
...
-  staticanalysis:
+  php_code_sniffer:
+    name: PHP Code Sniffer
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '7.4'
```
bshaffer, line 30:
```suggestion
        php-version: '8.1'
```

PR #1774, opened 2023-02-07
feat(Bigquery): Undelete table sample.

yash30201:
This sample showcases how to restore an accidently deleted table within the [time travel window](https://cloud.google.com/bigquery/docs/managing-tables#undeletetable).


bigquery/api/src/undelete_table.php
```diff
@@ -0,0 +1,71 @@
...
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/main/bigquery/api/README.md
+ */
+
+namespace Google\Cloud\Samples\BigQuery;
+
+# [START bigquery_undelete_table]
+use Google\Cloud\BigQuery\BigQueryClient;
+
+/**
+ * Restored a table from its snapshot for with deleted base table.
```
vishwarajanand, line 31:
```suggestion
 * Restore a deleted table from its snapshot.
```

bigquery/api/src/undelete_table.php
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @param string $projectId The project Id of your Google Cloud Project.
+ * @param string $datasetId The BigQuery dataset ID.
+ * @param string $snapshotId Snapshot ID of a snapshot with deleted base table
+ *  to be restored.
```
vishwarajanand, line 36: Other language samples, [ref](https://github.com/googleapis/nodejs-bigquery/blob/main/samples/undeleteTable.js#L53) take in a table id instead and form a snapshot id from the current timestamp. Any specific reason to not follow this pattern?
If none, then pls change otherwise it would look off on the docs page.

vishwarajanand:
```suggestion
 * @param string $tableId ID of the deleted table to be restored.
```

bigquery/api/src/undelete_table.php
```diff
@@ -0,0 +1,71 @@
...
+    string $projectId,
+    string $datasetId,
+    string $snapshotId,
+    string $restoredTableId
+): void {
+    $bigQuery = new BigQueryClient([
+        'projectId' => $projectId,
+    ]);
+    $dataset = $bigQuery->dataset($datasetId);
+    $snapshot = $dataset->table($snapshotId);
```
vishwarajanand, line 49: Something like this?
```suggestion
    // Construct the restore-from table ID at current timestamp
    $snapshotEpoch = date_create()->format('Uv');
    $snapshotId = "{$tableId}@{$snapshotEpoch}";
    $snapshot = $dataset->table($snapshotId);
```

yash30201: Wow, that seems a really nice way to do this! I'll incorporate this.

bigquery/api/src/undelete_table.php
yash30201: > Other language samples, [ref](https://github.com/googleapis/nodejs-bigquery/blob/main/samples/undeleteTable.js#L53) take in a table id instead and form a snapshot id from the current timestamp. Any specific reason to not follow this pattern? If none, then pls change otherwise it would look off on the docs page.

Not a really solid reason, it's just that the region tag seemed to describe only the process of restoring a table from it's snapshot. So went for it.

But yeah it does seems right to maintain uniformity with other laguages and describe the whole process of creation, deletion and then restoring. Will do this.

yash30201: It seems we need a sample for undeleting a table withing the [time travel window](https://cloud.google.com/bigquery/docs/managing-tables#undeletetable), thus we don't need to create a snapshot exclusively.

bigquery/api/src/undelete_table.php
yash30201: Thanks, applied in commit [8463095](https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1774/commits/84630956b7a6f5c17a73325e954c6c186a19a898)

bigquery/api/src/undelete_table.php
yash30201: Applied the changes in commit [8463095](https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1774/commits/84630956b7a6f5c17a73325e954c6c186a19a898) and correspondingly updating the description of the PR.

bigquery/api/src/undelete_table.php
yash30201: Thanks, applied in commit [8463095](https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1774/commits/84630956b7a6f5c17a73325e954c6c186a19a898)

bigquery/api/test/bigqueryTest.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
vishwarajanand, line 2: remove

bigquery/api/test/bigqueryTest.php
```diff
@@ -33,17 +34,18 @@ class FunctionsTest extends TestCase
...
 
+    private static $client;
     private static $datasetId;
     private static $dataset;
 
     public static function setUpBeforeClass(): void
     {
         self::$projectId = self::requireEnv('GOOGLE_PROJECT_ID');
-        $client = new BigQueryClient([
+        self::$client = new BigQueryClient([
```
vishwarajanand, line 44: This change is probably not needed and also the `private` keyword above.

vishwarajanand:
A side comment, do you think adding a `break` after `foundValue` is set in `verifyTable` makes sense?

https://github.com/GoogleCloudPlatform/php-docs-samples/blob/02086fb9b97dcefd63e4beef348a00a4fe743f1a/bigquery/api/test/bigqueryTest.php#L437-L444

vishwarajanand:
LGTM, just fix the few remaining nits and we are good to go.

yash30201:
> A side comment, do you think adding a `break` after `foundValue` is set in `verifyTable` makes sense?
> 
> https://github.com/GoogleCloudPlatform/php-docs-samples/blob/02086fb9b97dcefd63e4beef348a00a4fe743f1a/bigquery/api/test/bigqueryTest.php#L437-L444

@vishwarajanand, As we need to get the `numRows` also, so breaking would give incorrent `numRows` value thus the assertion would fail. Thus putting a `break` isn't feasible there.

bigquery/api/test/bigqueryTest.php
yash30201: Thanks for pointing this out. Fixed in commit [4e66b98](https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1774/commits/4e66b98882855596cef23b8d404ea457e50b425d)

bigquery/api/test/bigqueryTest.php
yash30201: Thanks for pointing this out. Fixed in commit [4e66b98](https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1774/commits/4e66b98882855596cef23b8d404ea457e50b425d)

vishwarajanand: missed pushing any commit?

vishwarajanand:
@yash30201 I was suggesting something like below which would preserve the `$numRows` value:

```
 foreach ($table->rows([]) as $row) { 
     foreach ($row as $column => $value) { 
         if ($value == $expectedValue) { 
             $foundValue = true;  
             break;  // here
         } 
     } 
     $numRows++; 
 } 
```

yash30201:
> @yash30201 I was suggesting something like below which would preserve the `$numRows` value:
> 
> ```
>  foreach ($table->rows([]) as $row) { 
>      foreach ($row as $column => $value) { 
>          if ($value == $expectedValue) { 
>              $foundValue = true;  
>              break;  // here
>          } 
>      } 
>      $numRows++; 
>  } 
> ```

@vishwarajanand , Oh you meant breaking outer loop, then should we do this as this? Because apart from matching row, all other rows would be traversed completely and thus putting a single break won't make much performance difference.
```php
 foreach ($table->rows([]) as $row) { 
     if (!$foundValue) {
         foreach ($row as $column => $value) { 
             if ($value == $expectedValue) { 
                 $foundValue = true;  
                 break;  // here
             } 
         }
     } 
     $numRows++; 
 }
```

vishwarajanand (approved):
LGTM

PR #1781, opened 2023-03-03
feat[BigqueryStorage]: Added quickstart sample

yash30201:
This sample showcases a basic quickstart for using the Bigquery Storage API.

The corresponsing published samples are [here](https://cloud.google.com/bigquery/docs/samples/bigquerystorage-quickstart)

Added `rg/avro-php v3.0` which required php >= 7.3 and as within a couple of weeks we are dropping support for php <= 7.3, thus this would work fine.


bigquerystorage/composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": {
+        "google/cloud-bigquery-storage": "^1.2"
```
yash30201, line 3: This is the latest stable version of cloud bigquery storage, do we go ahead with this?

bigquerystorage/phpunit.xml.dist
```diff
@@ -0,0 +1,34 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<phpunit bootstrap="../testing/bootstrap.php">
+    <testsuites>
+        <testsuite name="Google BigQuery Storage Quickstart Tests">
```
yash30201, line 19: This name is just to describe the test configuration, right? Or is there any convention for it?

bigquerystorage/phpunit.xml.dist
```diff
@@ -0,0 +1,34 @@
...
+            <directory>test</directory>
+        </testsuite>
+    </testsuites>
+    <logging>
+        <log type="coverage-clover" target="build/logs/clover.xml"/>
+    </logging>
+    <filter>
+        <whitelist>
+            <file>quickstart.php</file>
```
yash30201, line 28: It's directly specified to the `quickstart.php` file because we are required to create a basic quickstart for this product just like `bigquerydatatranfer API`.

yash30201:
I was confused on how to write this sample between a functional format or a standalone script format and then choose the script style as this felt inline as a quickstart.

bigquerystorage/composer.json
vishwarajanand: Confirmed the [release version to 1.2.11](https://github.com/googleapis/google-cloud-php-bigquery-storage/releases/tag/v1.2.11)

bigquerystorage/quickstart.php
```diff
@@ -0,0 +1,113 @@
...
+
+// Set a snapshot time if it's been specified.
+if ($snapshotMillis != null) {
+    $timestamp = new Timestamp();
+    $timestamp->setSeconds($snapshotMillis / 1000);
+    $timestamp->setNanos((int) ($snapshotMillis % 1000) * 1000000);
+    $tableModifier = new TableModifiers();
+    $tableModifier->setSnapshotTime($timestamp);
+    $readSession->setTableModifiers($tableModifier);
+}
```
vishwarajanand, line 62: This code-snippet will never execute, you needed to pass it as an argument.

bigquerystorage/test/quickstartTest.php
```diff
@@ -0,0 +1,45 @@
...
+class quickstartTest extends TestCase
+{
+    public function testQuickstart()
+    {
+        if (!$projectId = getenv('GOOGLE_PROJECT_ID')) {
+            $this->markTestSkipped('GOOGLE_PROJECT_ID must be set.');
+        }
+
+        $file = sys_get_temp_dir() . '/bigquerystorage_quickstart.php';
+        $contents = file_get_contents(__DIR__ . '/../quickstart.php');
```
vishwarajanand, line 29: Why are we copying the contents to a temp script?

bigquerystorage/test/quickstartTest.php
```diff
@@ -0,0 +1,45 @@
...
+        $contents = str_replace(
+            ['YOUR_PROJECT_ID', '__DIR__'],
+            [$projectId, sprintf('"%s/.."', __DIR__)],
+            $contents
+        );
```
vishwarajanand, line 34: Consider also doing this for `$snapshotMillis`, something like below:

```suggestion
        $milliseconds = floor(microtime(true) * 1000);
        $contents = str_replace(
            ['YOUR_PROJECT_ID', 'YOUR_SNAPSHOT_MILLIS', '__DIR__'],
            [$projectId, $milliseconds, sprintf('"%s/.."', __DIR__)],
            $contents
        );
```

bigquerystorage/quickstart.php
vishwarajanand: In python ([code pointer link](https://github.com/googleapis/python-bigquery-storage/blob/ad439bc83e52e753805ceca6e35a5f307e72a0c7/samples/quickstart/quickstart_test.py#L27-L36)), we are testing against **(A)** not setting the snapshotMillis at all _AND_ **(B)** passing a value of current time in milliseconds minus 5000.
Maybe replicate that here as well in two different test methods?

bigquerystorage/test/quickstartTest.php
vishwarajanand: Understood, this is needed to replace the values of variables.

vishwarajanand (changes requested):
Pls re-structure the code to make use of snapshit millis as an argument and the avro package.

bigquerystorage/test/quickstartTest.php
```diff
@@ -39,7 +43,12 @@ public function testQuickstart()
...
         $result = ob_get_clean();
 
-        // Make sure it looks correct
-        $this->assertStringContainsString('Total rows: 130809', $result);
+        // Assertion for without snapshot millis
+        $expected = sprintf('Got 6482 unique names in states: WA');
+        $this->assertStringContainsString($expected, $result);
+
+        // Assertion for with snapshot millis
+        $expected = sprintf('Got 6482 unique names in states: WA');
```
vishwarajanand, line 51: Both these assertions are same and might result in test pass when only one of the invocations give correct output. 
Pls split to two test methods (cleanest) _OR_ including the `$file` again without timestamp and just compare its output.

bigquerystorage/quickstart.php
```diff
@@ -51,17 +51,31 @@
...
+    if ($snapshotMillis != null) {
+        $timestamp = new Timestamp();
+        $timestamp->setSeconds($snapshotMillis / 1000);
+        $timestamp->setNanos((int) ($snapshotMillis % 1000) * 1000000);
+        $tableModifier = new TableModifiers();
+        $tableModifier->setSnapshotTime($timestamp);
+        $readSession->setTableModifiers($tableModifier);
+        $stream = getRowsStream($client, $project, $readSession);
+        deserializeRows($stream);
+    }
```
vishwarajanand, line 72: There is a duplication of code. It would be better to call `getRowsStream` just once. 
Previous design was neater, to call `setSnapshotTime` when timestamp is provided. Previously we were not setting the timestamp value.

bigquerystorage/quickstart.php
```diff
@@ -0,0 +1,106 @@
...
+use Google\Cloud\BigQuery\Storage\V1\DataFormat;
+use Google\Cloud\BigQuery\Storage\V1\ReadSession;
+use Google\Cloud\BigQuery\Storage\V1\ReadSession\TableModifiers;
+use Google\Cloud\BigQuery\Storage\V1\ReadSession\TableReadOptions;
+use Google\Protobuf\Timestamp;
+
+// Instantiates the client and sets the project
+$client = new BigQueryReadClient();
+$project = $client->projectName('YOUR_PROJECT_ID');
+$snapshotMillis = 'YOUR_SNAPSHOT_MILLIS';
```
bshaffer, line 32: Why not just supply a default value here? It seems strange to require this since it's not actually required for the sample to work (like `projectId` is)

bshaffer: I've been told that this is probably referenced in the devsite docs because other languages have the same requirement, so LGTM!

PR #1785, opened 2023-03-13
fix(Spanner): Making DML returning samples consistent with other languages

yash30201:
Refractoring the samples for DML returning samples in spanner so that they become consistent according to other languages.

Changes:
* Added a new derivative column(`FullName = FirstName + ' ' + LastName`) as required 
* Used derivative column for returning 
* Fixed updation and deletion returning sample tests to make them independent of other samples so that samples can become consistent according to other languages.


vishwarajanand (approved):
Lgtm

PR #1789, opened 2023-03-18
chore(docs): add composer install instructions

bshaffer:
in case anyone wants some extra help, add minimum instructions on enhancing the docker container to use composer


glasnt:
Depends on #1941 to get deployment tests green

glasnt (approved):

PR #1811, opened 2023-04-14
feat(Pubsub): Add samples for ordering keys

ajupazhamayil:


ajupazhamayil:
This PR merge is blocked by issue https://github.com/GoogleCloudPlatform/php-docs-samples/issues/1812

meredithslota:
#1812 is fixed, but tests are failing.

ajupazhamayil:
@meredithslota The tests are passing now.

pubsub/api/src/publish_with_ordering_keys.php
```diff
@@ -0,0 +1,52 @@
...
+ * @param string $projectId  The Google project ID.
+ * @param string $topicName  The Pub/Sub topic name.
+ */
+function publish_with_ordering_keys($projectId, $topicName)
+{
+    $pubsub = new PubSubClient([
+        'projectId' => $projectId,
+    ]);
+
+    $topic = $pubsub->topic($topicName);
```
vishwarajanand, line 42: Not very well versed with PubSub, but [Python Sample](https://github.com/googleapis/python-pubsub/blob/15d6cdf5ca2eeb6b2fcccc92fc1c69fef2ad07c9/samples/snippets/publisher.py#L320-L326) needs an ordering key in here. 

Pls check whether you need to call `$pubsub->createTopic($topicName, ['enableMessageOrdering' => true])` instead?

pubsub/api/src/enable_subscription_ordering.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+function enable_subscription_ordering($projectId, $topicName, $subscriptionName)
+{
+    $pubsub = new PubSubClient([
+        'projectId' => $projectId,
+    ]);
+    $topic = $pubsub->topic($topicName);
+    $subscription = $topic->subscription($subscriptionName);
+
+    $subscription->create(['enableMessageOrdering' => true]);
```
vishwarajanand, line 44: Does creating a subscription directly from client make more sense? Like this:
`$pubsub->createTopic($topicName, ['enableMessageOrdering' => true])`

vishwarajanand (commented):
Pls check whether the way to pass ordering flag is appropriate

pubsub/api/src/enable_subscription_ordering.php
ajupazhamayil: I think create topic is to create a topic, we need to look for subscribing with ordering. Python has different classes for publisher and subscriber but PHP doesn't. So we used this approach of giving the option at the `subscription->create` function instead of a `SubscriberClient` class (which we don't have).

PR #1829, opened 2023-05-05
feat: update secretmanager to new (beta) v2gapic surface

bshaffer:
Depends on https://github.com/googleapis/google-cloud-php/pull/6136

updates secretmanager to new V2 surface

### DIFF
--- a/secretmanager/src/iam_grant_access.php
+++ b/secretmanager/src/iam_grant_access.php
@@ -27,10 +27,12 @@ namespace Google\Cloud\Samples\SecretManager;
 
 // [START secretmanager_iam_grant_access]
 // Import the Secret Manager client library.
-use Google\Cloud\SecretManager\V1\SecretManagerServiceClient;
+use Google\Cloud\SecretManager\V1\Client\SecretManagerServiceClient;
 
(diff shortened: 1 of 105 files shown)


secretmanager/src/iam_grant_access.php
```diff
@@ -46,7 +48,7 @@ function iam_grant_access(string $projectId, string $secretId, string $member):
     $name = $client->secretName($projectId, $secretId);
 
     // Get the current IAM policy.
-    $policy = $client->getIamPolicy($name);
+    $policy = $client->getIamPolicy((new GetIamPolicyRequest)->setResource($name));
```
bshaffer, line 51: We can't use the `build` methods on these objects _yet_ because we haven't shipped the V2surfaces yet.

bshaffer: This will be added in https://github.com/googleapis/google-cloud-php/pull/6273

bshaffer: Actually, it won't be, as these classes are part of https://github.com/googleapis/common-protos-php and not the IAM library. I will open up a feature request for this

PR #1854, opened 2023-06-01
Response streaming

janell-chen:


functions/response_streaming/index.php
```diff
@@ -0,0 +1,34 @@
...
+    $queryResults = $job->queryResults();
+
+    $i = 0;
+    foreach ($queryResults as $row) {
+        foreach ($row as $column => $value) {
+            printf('%s' . PHP_EOL, json_encode($value));
+            flush();
+        }
+    }
+}
```
josephlewis42, line 34: Looks like this needs a newline by the linter

functions/response_streaming/phpunit.xml.dist
```diff
@@ -0,0 +1,34 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<phpunit bootstrap="../../testing/bootstrap.php" convertWarningsToExceptions="false">
+    <testsuites>
+        <testsuite name="Cloud Functions HTTP Methods Test Suite">
+            <directory>test</directory>
```
josephlewis42, line 20: I think this directory needs to exist and have tests for the other test one to pass. If you're doing mocks in other languages, it looks like this might be the way: https://github.com/GoogleCloudPlatform/php-docs-samples/blob/29c9bb31c91660d3f710555d5f8b77f4dd52f0e5/compute/logging/test/loggingTest.php

functions/response_streaming/index.php
```diff
@@ -0,0 +1,57 @@
...
+    $backoff->execute(function () use ($job) {
+        $job->reload();
+        if (!$job->isComplete()) {
+            throw new Exception('Job has not yet completed', 500);
+        }
+    });
+    $queryResults = $job->queryResults();
+
+    // Stream out large payload by iterating rows and flushing output.
+    $i = 0;
```
josephlewis42, line 48: looks like this is unused?

functions/response_streaming/index.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    // Provide the Cloud Project ID by setting env var.
+    $projectId = getenv('GOOGLE_PROJECT_ID');
+    $datasetId = 'my_new_dataset_' . time();
+    // Example large payload from BigQuery's public dataset.
+    $query = 'SELECT abstract FROM `bigquery-public-data.breathe.bioasq` LIMIT 1000';
+
+    $bigQuery = new BigQueryClient([
+        'projectId' => $projectId,
+        'datasetId' => $datasetId,
```
josephlewis42, line 33: Looks like this might not be necessary? https://cloud.google.com/php/docs/reference/cloud-bigquery/latest/BigQueryClient#_Google_Cloud_BigQuery_BigQueryClient____construct__

functions/response_streaming/index.php
garethgeorge, line 48: nit: unused?

functions/response_streaming/index.php
```diff
@@ -0,0 +1,57 @@
...
+    ]);
+    $jobConfig = $bigQuery->query($query);
+    $job = $bigQuery->startQuery($jobConfig);
```
josephlewis42, line 36: Would runquery be better here than creating a config and doing the backoff since this example is more to demonstrate the streaming than the semantics of calling BQ? https://cloud.google.com/php/docs/reference/cloud-bigquery/latest/BigQueryClient#_Google_Cloud_BigQuery_BigQueryClient__runQuery__

josephlewis42 (approved):
Couple of comments, but LGTM from my side

functions/response_streaming/index.php
janell-chen: I think the project id is required when running tests, unless I'm missing some env var that the tests use by default. 
I'm getting: GoogleException: No project ID was provided, and we were unable to detect a default project ID.

muncus (approved):

PR #1860, opened 2023-06-28
chore: show composer install logs

yash30201:
This PR is to inspect what versions of product libraries are getting installed while running system tests.

### DIFF
--- a/bigtable/src/create_app_profile.php
+++ b/bigtable/src/create_app_profile.php
@@ -43,6 +43,7 @@ function create_app_profile(
     string $clusterId,
     string $appProfileId
 ): void {
+    $instanceAdminClient = 'Hello';
     $instanceAdminClient = new BigtableInstanceAdminClient();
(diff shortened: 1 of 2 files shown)


bigtable/src/create_app_profile.php
```diff
@@ -43,6 +43,7 @@ function create_app_profile(
     string $clusterId,
     string $appProfileId
 ): void {
+    $instanceAdminClient = 'Hello';
```
bshaffer, line 46: ?

bshaffer (changes requested):

bigtable/src/create_app_profile.php
yash30201: This PR is just a test PR to enable composer logs while they install packages for gcp products. Earlier we had a `-q` flag.

I've verified that after the gax version upgrade, we've solved the issue of authentication scopes not passing properly so I'll close this PR now.

yash30201:
This is obselete, hence closing this now.

Issue #1763, opened 2023-01-06
feat: upgrade or remove Endpoints getting started sample


bshaffer:
The [endpoints/getting-started](https://github.com/GoogleCloudPlatform/php-docs-samples/issues/endpoints/getting-started) sample app is outdated, and needs to be either upgraded to the new format or removed. 

I have started upgrading the sample in https://github.com/GoogleCloudPlatform/php-docs-samples/commit/f36e623812240582b26ebbcab6593e3f3a58930b, but I'm not sure this works, or is even still a part of our cloud site samples.

vishwarajanand:
Thanks for taking this up @bshaffer. 
I checked it, there are only 3 tags in `/endpoints` swagger, esp and configuration. All of them are not tracked with a product but are used on cloud site.

vishwarajanand:
I will check whether the endpoints samples are showing up in dashboards. If they do, goal is to remove them.

bshaffer:
> but are used on cloud site.

That's good to know, and says to me we should keep them / update them if possible!

vishwarajanand:
Fixed by #1862

PR #1861, opened 2023-06-30
fix: upgrade endpoints samples to new format

vishwarajanand:
Fixes: https://github.com/GoogleCloudPlatform/php-docs-samples/issues/1763

Major contribution from Brent's commit: https://github.com/GoogleCloudPlatform/php-docs-samples/commit/f36e623812240582b26ebbcab6593e3f3a58930b


bshaffer:
This is great! But in the future, it would be better to use `git cherry-pick` on my commit and then make changes from there, so as to preserve the changes I made.

vishwarajanand:
Closing this PR, since #1862 is merged.

PR #1862, opened 2023-06-30
Upgrade endpoints samples

bshaffer:
#1861, but with proper commits separated


vishwarajanand (approved):
I created a local server in a terminal and executed `php src/make_request.php http://localhost:8080 APIKEY`  as described in README.

PR #1863, opened 2023-07-04
chore: phpunit migrate configuration

vishwarajanand:
PhpUnit throws the following error / suggestion, hence upgrading all it's configs by using `--migrate-configuration`:
```
Warning:       Your XML configuration validates against a deprecated schema.
Suggestion:    Migrate your XML configuration using "--migrate-configuration"!
```

Here's the script to do this migration:

Command to migrate PhpUnit configs:
```
php-docs-samples > find * -name 'phpunit.xml*' -not -path '*vendor/*' | \
xargs -I {} bash -c "echo \"## Running for '{}'\" && testing/vendor/bin/phpunit  -c {} --migrate-configuration && git add {}"
```

Log from script run: https://paste.googleplex.com/5904304555687936

### DIFF
--- a/analyticsdata/phpunit.xml.dist
+++ b/analyticsdata/phpunit.xml.dist
@@ -14,24 +14,25 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-<phpunit bootstrap="../testing/bootstrap.php">
-    <testsuites>
-        <testsuite name="Google Analytics Data API Tests">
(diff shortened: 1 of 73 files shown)


vishwarajanand:
A lot of phpunit files failed to migrate due to invalid `whitelist` tag and removing this from phpunit's whitelist migrates them:

`<file>index.php</file>`

I'll enquire more abt it

PR #1866, opened 2023-07-11
feat(functions): create new samples for typed function signature

garethgeorge:
Creates new samples for GCF python strongly typed function signature. Samples added are

 * greeting - a typed hello world (accepting JSON)


garethgeorge:
CC @bshaffer for familiarity with the typed functions project (if you have review bandwidth)

functions/typed_greeting/test/TestCasesTrait.php
```diff
@@ -0,0 +1,36 @@
...
+        return [
+            [
+                'label' => 'Default',
+                'first_name' => 'Jane',
+                'last_name' => 'Doe',
+                'expected_message' => 'Hello Jane Doe!',
+            ],
+        ];
+    }
+}
```
bshaffer, line 36: Just for simplicity I'd just put this function in the `UnitTest.php` file for simplification

bshaffer (changes requested):
One suggestion - can simplify tests into into one file by copying the method over. Traits are typically for code reuse, so if you're only including the trait in one test, it doesn't really make sense. 

If you think you'll be a `SystemTest` or `DeployTest` soon (like we have in the other function samples), then we can keep them!

functions/typed_greeting/test/TestCasesTrait.php
garethgeorge: Thanks @bshaffer! Makes sense, I don't think we'll be needing more tests out of this sample.

grayside (approved):

functions/typed_greeting/test/UnitTest.php
```diff
@@ -0,0 +1,69 @@
...
+
+declare(strict_types=1);
+
+namespace Google\Cloud\Samples\Functions\TypedGreeting\Test;
+
+use PHPUnit\Framework\TestCase;
+use GreetingRequest;
+use GreetingResponse;
+
+require_once __DIR__ . '/TestCasesTrait.php';
```
bshaffer, line 26:
```suggestion
```

bshaffer (approved):

PR #1887, opened 2023-07-20
chore: upgrade speech samples to new surface

bshaffer:


bshaffer:
These need to be upgraded to Speech V2 in order to take advantage of the new client

PR #1892, opened 2023-07-24
feat(Transcoder): add batch mode sample and test

irataxy:
b:288111390

Please merge on approval.


media/transcoder/test/transcoderTest.php
```diff
@@ -224,6 +226,31 @@ public function testJobFromPreset()
...
+            self::$outputUriForPresetBatchMode,
+            self::$preset
+        ]);
+
+        $this->assertMatchesRegularExpression(sprintf('%s', self::$jobIdRegex), $output);
+
+        $jobId = explode('/', $output);
+        $jobId = trim($jobId[(count($jobId) - 1)]);
+
+        sleep(30);
```
bshaffer, line 244: `assertJobStateSucceeded` already uses an exponential backoff, so 30 seconds seems a bit steep here. I feel like no sleep, or maybe at most 5-10 seconds, should be sufficient

irataxy: Adjusted to 10 sec.

bshaffer (approved):

PR #1901, opened 2023-07-28
feat: initial commit of run mc php nginx sample

pattishin:
**Issue:** To support documentation in https://cloud.google.com/run/docs/tutorials (php sample angle)

This uses new cloud build trigger `hello-php-nginx-sample-pr` and adds new e2e test.

(Moved from https://github.com/GoogleCloudPlatform/cloud-run-samples/pull/199)


pattishin:
Note: Adding PR to `php-tools` to add `--source` flag to gcloud submit command in `CloudRun` class stub. (https://github.com/GoogleCloudPlatform/php-tools/pull/128)

run/multi-container/hello-php-nginx-sample/Makefile
```diff
@@ -0,0 +1,19 @@
+PROJECT_NAME  ?= ${PROJECT_ID}
+REPO_NAME     ?= ${REPO_NAME}
+REGISTRY      ?= ${REGION}-docker.pkg.dev
+MC_SERVICE_NAME ?= ${MC_SERVICE_NAME}
```
JulienBreux, line 4:
```suggestion
PROJECT_NAME    ?= ${PROJECT_ID}
REPO_NAME       ?= ${REPO_NAME}
REGISTRY        ?= ${REGION}-docker.pkg.dev
MC_SERVICE_NAME ?= ${MC_SERVICE_NAME}
```

run/multi-container/hello-php-nginx-sample/nginx/nginx.conf
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START cloudrun_mc_hello_php_nginx_serving_conf]
+server {
+    # Server at localhost
+    server_name localhost;
```
JulienBreux, line 18:
```suggestion
    server_name _;
```
Related to this post https://cloud.google.com/blog/products/serverless/cloud-run-now-supports-multi-container-deployments we don't name the server :)

run/multi-container/hello-php-nginx-sample/nginx/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# [START cloudrun_mc_hello_php_nginx_serving_dockerfile]
+
+# Context: ./
+
+FROM nginx:alpine
```
grayside, line 5: question: We often prefer to track major version where available. Should this follow `nginx:1-alpine`?

run/multi-container/hello-php-nginx-sample/nginx/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# [START cloudrun_mc_hello_php_nginx_serving_dockerfile]
+
+# Context: ./
```
grayside, line 3: suggestion: I like this idea of clarifying the expected usage of the Dockerfile! If we are going to include this, we should expand this comment to minimize assumption the reader understands docker context.

run/multi-container/hello-php-nginx-sample/nginx/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# [START cloudrun_mc_hello_php_nginx_serving_dockerfile]
```
grayside, line 1: suggestion: In case we may want to have other samples showing an nginx proxy in front of an application, could we remove "PHP" from the region tag? We generally try to avoid including language in the region tag name unless the sample is related to language-specific functionality.

grayside, line 1: suggestion: This sample is already in a PHP context, maybe we should change this to `hello-mc-nginx` or similar?

run/multi-container/hello-php-nginx-sample/nginx/nginx.conf
```diff
@@ -0,0 +1,49 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START cloudrun_mc_hello_php_nginx_serving_conf]
+server {
+    # Server at localhost
+    server_name _;
+    # Listen at port 8080
+    listen 8080;
```
grayside, line 20: suggestion: Given this is hard-coded to 8080, it may be useful to comment that if port is specified, it must be `8080`.

https://cloud.google.com/run/docs/configuring/services/containers#configure-port

run/multi-container/hello-php-nginx-sample/php-app/opcache.ini
```diff
@@ -0,0 +1,9 @@
+[opcache]
+opcache.enable=1
+opcache.revalidate_freq=0
+opcache.validate_timestamps=0
+opcache.max_accelerated_files=10000
+opcache.memory_consumption=192
+opcache.max_wasted_percentage=10
+opcache.interned_strings_buffer=16
+opcache.fast_shutdown=1
```
grayside, line 9: nit: `opcache.fast_shutdown` has been removed as of PHP 7.2. https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.fast-shutdown

```suggestion
```

run/multi-container/hello-php-nginx-sample/php-app/opcache.ini
```diff
@@ -0,0 +1,9 @@
+[opcache]
+opcache.enable=1
+opcache.revalidate_freq=0
+opcache.validate_timestamps=0
```
grayside, line 4: suggestion: `revalidate_freq=0` means it will check timestamps for updated code on every request, but only if validate_timestamps isn't disabled. 

Including this only makes sense if folks know to set `opcache.validate_timestamps=1` in the event they are doing active development using advanced techniques such as volume mounts to have live code changes served by a running container.

I would either remove `revalidate_freq` or add a comment explaining this usage.

run/multi-container/hello-php-nginx-sample/service.yaml
```diff
@@ -0,0 +1,56 @@
...
+spec:
+  template:
+    metadata:
+      annotations:
+        run.googleapis.com/execution-environment: gen2
+        # Defines container startup order within multi-container service.
+        # Below requires nginx container to spin up before php container,
+        # which depends on nginx.
+        # https://cloud.google.com/run/docs/configuring/containers#container-ordering
+        run.googleapis.com/container-dependencies: '{"hellophp":["nginx"]}'
```
grayside, line 35: blocker: This looks backwards to me based on the documentation: We should start the PHP server before the proxy to ensure the backing service is ready before nginx starts.

run/multi-container/hello-php-nginx-sample/service.yaml
```diff
@@ -0,0 +1,56 @@
...
+              memory: 256Mi
+        - name: hellophp
+          image: "REGION-docker.pkg.dev/PROJECT_ID/REPO_NAME/php"
+          env:
+            - name: PORT
+              value: "9000"
+          resources:
+            limits:
+              cpu: 1000m
+              memory: 512Mi
```
grayside, line 55: suggestion: One of the key intermediate things we may want to gain from this sample is showing how to configure the php-fpm container. Using the configuration I recommended above, each PHP-FPM process is limited to 128MB.

512 - 192MB (Opcache) = 320MB, 320/128 = 2.5. This configuration supports ~2.5 concurrency.

I think it would be valuable to show how to introduce a php configuration override and set the php-fpm memory limit somewhat lower, something like 40MB or 64MB.

You might consider this a bit advanced for the sample, but tailing php settings to the application is broadly required.

run/multi-container/hello-php-nginx-sample/php-app/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# [START cloudrun_mc_hello_php_nginx_sidecar_dockerfile]
+
+# Context: ../
+
+FROM php:8-fpm-alpine
+
+RUN docker-php-ext-install opcache
```
grayside, line 7: suggestion: This Dockerfile is leaving default PHP.ini settings active. From https://hub.docker.com/_/php section on configuration, we should more proactively guide toward recommended configurations.

```suggestion
# Use the default production configuration
RUN mv "$PHP_INI_DIR/php.ini-production" "$PHP_INI_DIR/php.ini"

RUN docker-php-ext-install opcache
```

run/multi-container/hello-php-nginx-sample/service.yaml
```diff
@@ -0,0 +1,56 @@
...
+        - name: nginx
+          image: "REGION-docker.pkg.dev/PROJECT_ID/REPO_NAME/nginx"
+          ports:
+            - name: http1
+              containerPort: 8080
+          resources:
+            limits:
+              cpu: 500m
+              memory: 256Mi
+        - name: hellophp
```
grayside, line 47: suggestion: As noted in my comment on container memory and PHP-FPM configuration below, there is a set concurrency this container will support based on memory limits. We should probably specify a more supportable concurrency than the default.

grayside (dismissed):
Thanks for this PR, I'm really excited to see a PHP-FPM multi-container sample for Cloud Run, having spent some time in 2019 trying to make an all-in-one container without it becoming overwhelming. I've left many suggestion that are aimed at making this example more polished and helpful as a container onramp for PHP developers.

The only blocker I have is a consideration that the container dependency order seems reversed.

run/multi-container/hello-php-nginx-sample/service.yaml
grayside: Follow-up: I thought further on this, and realized we made a different decision about memory configuration in the Cloud Run Helloworld: https://github.com/GoogleCloudPlatform/php-docs-samples/blob/main/run/helloworld/Dockerfile

Specifically, we turned off the PHP process memory limit, and are relying on Cloud Run to handle memory. In the context of PHP-FPM, this may lead to increased importance on concurrency and verifying the php-fpm child configuration is effective: e.g., a static count matching the concurrency limit.

grayside (commented):
While there are a lot of my comments here, I think they are things that are either optional or can be handled in a follow-up. I did not have a chance to review the test implementation in detail, which is why I'm not adding LGTM. I am away next week, I am open to someone else completing the review and approving, otherwise I will be back!

run/multi-container/hello-php-nginx-sample/Makefile
```diff
@@ -0,0 +1,19 @@
+PROJECT_NAME    ?= ${PROJECT_ID}
```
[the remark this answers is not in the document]
bshaffer: I would definitely suggest using composer scripts - this is something we've done in other samples:

https://github.com/GoogleCloudPlatform/php-docs-samples/blob/ac4062aa1d2fb3844e4edcb4b29577a77bb051a5/functions/helloworld_get/composer.json#L5-L10

This would allow the user to run the script just by running `composer start`

run/multi-container/hello-php-nginx-sample/Makefile
```diff
@@ -0,0 +1,19 @@
...
+login: ## [project] Log-in to registry
+	gcloud auth login
+	gcloud artifacts repositories create ${REPO_NAME} --repository-format=docker
+
+build: ## [project] Build images to Artifact Registry
+	gcloud builds submit --tag=${REGION}-docker.pkg.dev/${PROJECT_ID}/${REPO_NAME}/nginx ./nginx
+	gcloud builds submit --tag=${REGION}-docker.pkg.dev/${PROJECT_ID}/${REPO_NAME}/php ./php-app
+
+deploy: ## [project] Deploy to Cloud Run MC service
+	sed -i -e s/MC_SERVICE_NAME/${MC_SERVICE_NAME}/g -e s/REGION/${REGION}/g -e s/REPO_NAME/${REPO_NAME} service.yaml
```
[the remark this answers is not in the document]
pattishin: Removed Makefile at this time - for the `sed` command, I specified to users to replace them manually for now. Test still uses `sed`.

run/multi-container/hello-php-nginx-sample/Makefile
```diff
@@ -0,0 +1,19 @@
+PROJECT_NAME    ?= ${PROJECT_ID}
+REPO_NAME       ?= ${REPO_NAME}
+REGISTRY        ?= ${REGION}-docker.pkg.dev
+MC_SERVICE_NAME ?= ${MC_SERVICE_NAME}
+
+login: ## [project] Log-in to registry
+	gcloud auth login
+	gcloud artifacts repositories create ${REPO_NAME} --repository-format=docker
```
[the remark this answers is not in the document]
pattishin: Great point. Removed Makefile at this time, but will note for future 👍🏼 . Ty!

run/multi-container/hello-php-nginx-sample/Makefile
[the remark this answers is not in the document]
pattishin: Added a small `composer.json` file and removed makefile so users will rely on readme. 👍🏼

run/multi-container/hello-php-nginx-sample/service.yaml
pattishin: I opted for 64MB and added a doc link for users to read more about Cloud Run memory limit. Would you say that more information (maybe in readme) around thinking through php memory settings is needed for users to be successful?

pattishin:
Ty for the feedback! Whenever you have a moment @bshaffer @grayside  - ready for eyes once again 👀

run/multi-container/hello-php-nginx-sample/nginx/Dockerfile
grayside: Pointing to the comment above, we're not calling it context, docker/container tooling calls it context. Maybe we should show an example of how to build the container?

run/multi-container/hello-php-nginx-sample/service.yaml
grayside: Per my follow-up above, we should make memory limit of the PHP process unlimited, and set the concurrency level to match the number of fpm workers and align that with the memory limit given to the Cloud Run service.

I think some documentation in the README or even better in the Cloud Run docs pointing to https://cloud.google.com/run/docs/tips/general#optimize_concurrency would be good.

grayside (approved):
LGTM with further comments.

PR #1911, opened 2023-08-23
fix(Datastore): migrate insert sample to individual file

vishwarajanand:
This PR migrates a single Datstore sample to individual file format. 
ref: [b/138631070](http://b/138631070)

## Changes

1. __Removed autoload from `composer.json` to load new samples (existing tests will fail till then, so removed now to migrate samples one by one)__
2. Removed a particular sample from `/datastore/api/src/functions/concepts.php`
3. Added the sample (with license information & removing the return statements) to its own file `/datastore/api/src/<function_name>.php`
4. Added a print statement with an information specific to the sample, say some id, description string, etc
5. Modified the sample test function in `/datastore/api/test/ConceptsTest.php` to assert on the sample's print statement.

## Tests
1. Ensure we have a GCP with Datastore enabled.
2. Run tests:
```
➜  php-docs-samples git:(datastore_samples_migration) cd datastore/api                                                           
➜  api git:(datastore_samples_migration) ../../testing/vendor/bin/phpunit -c phpunit.xml.dist --filter="testInsert"                     
PHPUnit 9.6.11 by Sebastian Bergmann and contributors.

Warning:       XDEBUG_MODE=coverage or xdebug.mode=coverage has to be set

.                                                                   1 / 1 (100%)

Time: 00:00.309, Memory: 10.00 MB

OK (1 test, 1 assertion)
➜  api git:(datastore_samples_migration)
```

### DIFF
--- a/datastore/api/composer.json
+++ b/datastore/api/composer.json
@@ -1,11 +1,5 @@
 {
     "require": {
         "google/cloud-datastore": "^1.2"
-    },
-    "autoload": {
-        "psr-4": { "Google\\Cloud\\Samples\\Datastore\\": "src" },
-        "files": [
-            "src/functions/concepts.php"
-        ]
     }
 }
--- a/datastore/api/composer.json
+++ b/datastore/api/composer.json
@@ -1,5 +1,11 @@
 {
     "require": {
         "google/cloud-datastore": "^1.2"
+    },
+    "autoload": {
+        "psr-4": { "Google\\Cloud\\Samples\\Datastore\\": "src" },
+        "files": [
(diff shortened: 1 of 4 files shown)


bshaffer:
Looks good but two thoughts: 

> Removed autoload from composer.json to load new samples (existing tests will fail till then, so removed now to migrate samples one by one

I see no changes to `composer.json`, so what is this referring to?

Also, we should do these all at once!

vishwarajanand:
@bshaffer sure thing, @shivgautam will take care of it.

shivgautam:
Fixed as part of #1968

PR #1913, opened 2023-08-25
fix: updates cloud run mc php nginx sample

pattishin:
**Updates include:**
* Adding note in readme about lack of explicit settings defining php-fpm workers
* Add `opcache.ini` content into Dockerfile
* Swap from `Mi` memory unit to `M`

### DIFF
--- a/run/multi-container/hello-php-nginx-sample/README.md
+++ b/run/multi-container/hello-php-nginx-sample/README.md
@@ -51,7 +51,7 @@ gcloud builds submit --tag=${REGION}-docker.pkg.dev/${PROJECT_ID}/${REPO_NAME}/p
 
 4. Configure the service with the appropriate memory limit.
 
-You will see as you read through `service.yaml`, that the memory limit has been explicitly set to `320Mi`. This leaves ~128M for PHP processes after allocating 192M for opcache. 
+You will see as you read through `service.yaml`, that the memory limit has been explicitly set to `335M`. This leaves ~143M for PHP processes after allocating 192M for opcache. 
 See how we got [here](https://cloud.google.com/run/docs/configuring/services/memory-limits#optimizing) and read more about how to [optimize for concurrency](https://cloud.google.com/run/docs/tips/general#optimize_concurrency).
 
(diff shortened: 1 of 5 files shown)


run/multi-container/hello-php-nginx-sample/php-app/cloudrun.ini
```diff
@@ -0,0 +1,2 @@
+; Configure Cloud Run memory
```
grayside, line 1: Need to copy this into Dockerfile.

I'm also comfortable if we want to put the opcache settings in this file and rename the file being copied in the existing Dockerfile line.

grayside (changes requested):

run/multi-container/hello-php-nginx-sample/README.md
```diff
@@ -51,10 +51,11 @@ gcloud builds submit --tag=${REGION}-docker.pkg.dev/${PROJECT_ID}/${REPO_NAME}/p
 
 4. Configure the service with the appropriate memory limit.
 
-You will see as you read through `service.yaml`, that the memory limit has been explicitly
-set to `320Mi` due to container concurrency of `1` with a single `fpm` worker.
+You will see as you read through `service.yaml`, that the memory limit has been explicitly set to `320Mi` due to container concurrency of `1`.
```
grayside, line 54: I missed these numbers on previous read-through, and I'm missing some context for why this is true.

320Mi - 192M = ~128, and this app definitely does not need 128M per process.

Let me suggestion a change based on the understanding I have on what we're trying to say:

```suggestion
You will see as you read through `service.yaml`, that the memory limit has been explicitly set to `320Mi`. This leaves ~128M for PHP processes after allocating 192M for opcache. 
```

Note that we are using Mi and M as memory units together, if we can use M in the service.yaml it would make the math simpler and more accurate because I've not done conversions on 320Mi to M in this math.

grayside (approved):
Looking nice! I saw another consideration in the README on the nuances of memory. I don't think it blocks this improvement but I think just a bit more polish there would be a good idea.

grayside (approved):

PR #1917, opened 2023-09-15
fix: refactor error reporting code sample

minherz:
align code sample for error reporting with code in other languages (b/275730159):

- use the same error message across samples in all languages
- use correct env var for capturing project Id in the managed environment

### DIFF
--- a/error_reporting/quickstart.php
+++ b/error_reporting/quickstart.php
@@ -11,7 +11,7 @@ use Google\Cloud\Core\Report\SimpleMetadataProvider;
 
 // These variables are set by the App Engine environment. To test locally,
 // ensure these are set or manually change their values.
-$projectId = getenv('GCLOUD_PROJECT') ?: 'YOUR_PROJECT_ID';
+$projectId = getenv('GOOGLE_CLOUD_PROJECT') ?: 'YOUR_PROJECT_ID';
 $service = getenv('GAE_SERVICE') ?: 'error_reporting_quickstart';
 $version = getenv('GAE_VERSION') ?: 'test';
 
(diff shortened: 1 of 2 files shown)


bshaffer:
It's worth mentioning that both `GCLOUD_PROJECT` and `GOOGLE_CLOUD_PROJECT` env vars are supported in the client library.

bshaffer:
I'm also curious what you mean by "use the fixed error message" and "use the client library implementation for capturing the error context"? All I can see is the example error message changed, which is pretty arbitrary IMHO.

minherz:
@bshaffer regarding your [comment][1], I am not sure what do you mean by this. `GOOGLE_CLOUD_PROJECT` is a "standard" environment variable supported by auth library and CLI and it is implicitly defined in all managed environments. The code was refactored to use it instead of the legacy variable (`GCLOUD_PROJECT).

Where do you propose to place the mentioning? We should mind that the "support" might be misleading here because the auth library (not the error reporting client library) _uses_ this variable to auto-discover the project ID. 

[1]: https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1917#issuecomment-1721904506

minherz:
@bshaffer regarding your [other comment][2], the provided bug (you can see it in the internal buganizer system) should provide sufficient explanations. Please, let me know if you still have questions.

[2]: https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1917#issuecomment-1721907416

bshaffer:
If you are just making the error message consistent with the other language samples, that makes sense to me. But the first two bullet points you've provided do not seem to have anything to do with this PR. I suggest changing the PR description to accurately describe what this PR is doing.

bshaffer (approved):

PR #1927, opened 2023-10-13
fix: deprecate assertRegexp usage over assertMatchesRegularExpression

vishwarajanand:
We are seeing warnings like below which will be fixed by this PR:
```
There was 1 warning:

1) Google\Cloud\Samples\PubSub\PubSubTest::testPublishAndSubscribeWithOrderingKeys
assertRegExp() is deprecated and will be removed in PHPUnit 10. Refactor your code to use assertMatchesRegularExpression() instead.
```

ref: https://github.com/sebastianbergmann/phpunit/issues/4086

### DIFF
--- a/appengine/flexible/metadata/test/DeployTest.php
+++ b/appengine/flexible/metadata/test/DeployTest.php
@@ -31,7 +31,7 @@ class DeployTest extends TestCase
             '200',
             $resp->getStatusCode(),
             'Top page status code should be 200');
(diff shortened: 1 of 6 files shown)


yash30201 (approved):

bshaffer:
Very minor nit here, but to "deprecate" something means to mark it as no longer recommended for use. So this PR does not deprecate anything, but rather updates the methods being used because they _were deprecated_ by PHPUnit.

PR #1936, opened 2023-11-13
fix: update message reflecting functionality of error_log

glasnt:
`error_log` flows through to Cloud logging, update message and comment. 

Fixes b/286058340


glasnt:
I think this PR is failing due to php-tools issues with identity, trying to access private functions.

glasnt:
Looking into this more: it's not the deployment that's failing (as I previously guessed); it's the SystemTest that's failing, which uses `CloudFunctionLocalTestTrait` which tries to run the function locally. 

See logs in https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1943 where there's no changes, and the SystemTest fails. 

I'm not able to replicate this specific error locally (though I am getting a different error kokoro isn't getting)

glasnt:
After updating php-tools for the linked PR (or alternative solution), tests for this should succeed. 

Keeping in draft until then.

glasnt:
Ready for review ✨

iennae (approved):
Thank you. Thank you. Thank you.

PR #1939, opened 2023-12-04
feat(Spanner): Add directed read samples

ajupazhamayil:
This is dependent on https://github.com/googleapis/google-cloud-php/pull/6078
Once the dependent PR merged, need to update the `composer.json` file to match the `google-cloud-php` latest version.

### DIFF
--- a/spanner/src/directed_read.php
+++ b/spanner/src/directed_read.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Copyright 2023 Google LLC.
+ * Copyright 2024 Google LLC.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spanner/src/directed_read.php
+++ b/spanner/src/directed_read.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * Copyright 2023 Google LLC.
+ *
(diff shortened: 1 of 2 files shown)


spanner/src/directed_read.php
```diff
@@ -0,0 +1,88 @@
+<?php
+/**
+ * Copyright 2023 Google LLC.
```
vishwarajanand, line 3: 😼 
```suggestion
 * Copyright 2024 Google LLC.
```

spanner/src/directed_read.php
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/main/spanner/README.md
+ */
+
+namespace Google\Cloud\Samples\Spanner;
+
+// [START spanner_directed_read]
```
vishwarajanand, line 26: Are you sure this is the correct tag? Only C++ seems to have implemented this!

vishwarajanand (approved):
Lgtm, approving to unblock.
Pls check the `region_tag` and run tests once before merge.

spanner/src/directed_read.php
ajupazhamayil: Yes! Its mentioned in the shared design doc.

Issue #1962, opened 2024-02-08
feat: Update Cloud Functions samples to use non-EOL PHP


bshaffer:
Some of our Cloud Functions samples still use PHP 7.4

 - https://github.com/GoogleCloudPlatform/php-docs-samples/blob/add395fbcfb6cc55f3f638b025ddc3d6b6d31f59/functions/helloworld_storage/composer.json#L3
 - https://github.com/GoogleCloudPlatform/php-docs-samples/blob/add395fbcfb6cc55f3f638b025ddc3d6b6d31f59/functions/typed_greeting/composer.json#L3
 - https://github.com/GoogleCloudPlatform/php-docs-samples/blob/add395fbcfb6cc55f3f638b025ddc3d6b6d31f59/functions/helloworld_pubsub/composer.json#L3
 - https://github.com/GoogleCloudPlatform/php-docs-samples/blob/add395fbcfb6cc55f3f638b025ddc3d6b6d31f59/functions/helloworld_http/composer.json#L3

As this version of PHP is now EOL, we need to upgrade these to the more current versions of PHP 8.2 or 8.3.

Issue #1961, opened 2024-02-08
feat: update Cloud Run sample to use non-EOL PHP


bshaffer:
Our [helloworld Dockerfile](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/HEAD/run/helloworld/Dockerfile) for Cloud Run currently uses PHP 8.0. As this version of PHP is now EOL, we need to upgrade this to the more current versions of PHP 8.2 or 8.3.

PR #1963, opened 2024-02-08
feat: upgrade all instances of PHP 7.4 and 8.0 to 8.1 or higher

bshaffer:
addresses https://github.com/GoogleCloudPlatform/php-docs-samples/issues/1962 and https://github.com/GoogleCloudPlatform/php-docs-samples/issues/1961

see cl/634137632


bigtable/src/update_app_profile.php
```diff
@@ -90,7 +90,7 @@ function update_app_profile(
         if ($operationResponse->operationSucceeded()) {
             $updatedAppProfile = $operationResponse->getResult();
             printf('App profile updated: %s' . PHP_EOL, $updatedAppProfile->getName());
-        // doSomethingWith($updatedAppProfile)
+            // doSomethingWith($updatedAppProfile)
```
Hectorhammett, line 93: Should we remove this line?

appengine/standard/slim-framework/README.md
```diff
@@ -34,7 +34,7 @@ The application consists of three components:
  3. An [`index.php`](index.php) which handles all the requests which get routed to your app.
 
 The `index.php` file is the most important. All applications running on App Engine
-for PHP 7.4 require use of a [front controller][front-controller] file.
+for PHP require use of a [front controller][front-controller] file.
```
Hectorhammett, line 37: Why did we drop the version in these files?

I could see this being redundant though. Are the minimum requirements stated elsewhere?

bigtable/src/update_app_profile.php
```diff
@@ -90,7 +90,7 @@ function update_app_profile(
...
         } else {
             $error = $operationResponse->getError();
             // handleError($error)
```
Hectorhammett, line 96: And this one?

bigtable/src/update_app_profile.php
bshaffer: I am ambivalent... someone put it here because they thought it illustrated something important. I only changed it because it was indented improperly.

bigtable/src/update_app_profile.php
bshaffer: Same with this one - the comment is meant as an example of something a user could implement, e.g. they could implement some error handling here with a fake function such as `handleError`

appengine/standard/slim-framework/README.md
bshaffer: Because the instructions are not specific to a version of PHP, so we can remove the "7.4" because it's an unnecessary qualifier.

bshaffer:
Looks like the `dlp` and `run/laravel` tests are failing for every version of PHP

PR #1966, opened 2024-02-14
feat(spanner): Replace Spanner Database Admin samples (1/3)

ajupazhamayil:


spanner/src/add_column.php
```diff
@@ -24,27 +24,31 @@
...
- * add_column($instanceId, $databaseId);
+ * add_column($projectId, $instanceId, $databaseId);
  * ```
  *
+ * @param string $projectId The Google Cloud project ID.
  * @param string $instanceId The Spanner instance ID.
  * @param string $databaseId The Spanner database ID.
  */
-function add_column(string $instanceId, string $databaseId): void
+function add_column(string $projectId, string $instanceId, string $databaseId): void
```
vishwarajanand, line 41: instead you can just do with full database name?
But either way works.. pls ensure this is consistent with other languages

spanner/src/add_column.php
```diff
@@ -24,27 +24,31 @@
 namespace Google\Cloud\Samples\Spanner;
 
 // [START spanner_add_column]
```
vishwarajanand, line 26: Will the tag remain the same? I mean the `src/add_column` **VS** `admin/archives/add_column`?

spanner/test/spannerBackupTest.php
```diff
@@ -283,6 +283,14 @@ private function runFunctionSnippet($sampleName, $params = [])
         );
     }
 
+    private function runAdminFunctionSnippet($sampleName, $params = [])
```
vishwarajanand, line 286: Can you try to make this function coherent with the other two implementation which are same but different from this one, and move it to a common place like `SpannerTestCase`?

vishwarajanand: I am not sure whether its doable entirely..

vishwarajanand (changes requested):
Left few questions for clarifications

spanner/src/add_column.php
ajupazhamayil: Yes, tags remains same

spanner/src/add_column.php
ajupazhamayil: I see `Java` used this way here: https://github.com/googleapis/java-spanner/blob/c58042375a216ec42330cc7a11b22f3b1d3c450b/samples/snippets/src/main/java/com/example/spanner/CreateSequenceSample.java#L43

So, we are in sync with Java

spanner/test/spannerBackupTest.php
ajupazhamayil: Yeah, I am also not sure, I have added it here since they already have other functions locally. Also, its just a function call from the trait. Let me know if it needs to be moved. I am okay with both.

ajupazhamayil:
Second PR raised here: https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1966

spanner/test/spannerBackupTest.php
ajupazhamayil: Also, I have removed this function from `spannerBackupTest` file as it is not required there. So, this is a function that required by two of the test files in a little different way.

spanner/src/admin/archived/list_database_operations.php
vishwarajanand, line 1: Seems this the admin-samples are supposed to be in `/src` as well as `/src/archived`.
But I don't see this being followed for following files:
```
src/enable_fine_grained_access.php
src/get_instance_config.php
src/list_database_operations.php
src/list_instance_configs.php
```

vishwarajanand: Okay, I see the three of these are covered in #1969 but not the `enable_fine_grained_access` can you check why its only in src and not in src/admin/archived as well?

vishwarajanand:
One more Q, why do we have `admin/archived` and not just `admin`?

vishwarajanand (approved):
I tested that the new samples are passing in tests, follow our style guide and they indeed run correct functionality. 
I also see that kokoro tests are running, so the changes lgtm.

spanner/src/admin/archived/list_database_operations.php
ajupazhamayil: Its because `enable_fine_grained_access` was already using the required auto generated admin client. I just modified the way the sample creates `databaseName` just for simplicity. But I still created a copy of it just for completion. Thank you!

ajupazhamayil:
Third PR raised here https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1973

ajupazhamayil:
Other two PRs are merged to this branch. We will wait for the green signal from Spanner team to merge this PR to `main`.

Other two PRs are:
PR1: https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1969
PR2: https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1973

PR #1968, opened 2024-02-16
chore(samples): Split PHP Datastore samples

shivgautam:


datastore/api/src/inequality_range.php
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\Samples\Datastore;
+
+use DateTime;
+use Google\Cloud\Datastore\DatastoreClient;
+use Google\Cloud\Datastore\Query\Query;
+
+// [END datastore_cursor_paging]
```
vishwarajanand, line 24: this seems misplaced

datastore/api/src/transactional_retry.php
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\Samples\Datastore;
+
+use Google\Cloud\Datastore\DatastoreClient;
+use Google\Cloud\Datastore\Key;
+
+// [END datastore_transactional_update]
```
vishwarajanand, line 23: this seems misplaced

datastore/api/src/ancestor_query.php
```diff
@@ -0,0 +1,53 @@
+<?php
+/**
+ * Copyright 2023 Google Inc.
```
bshaffer, line 3: nit: these should all be 2024

datastore/api/src/array_value_inequality_range.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+function array_value_inequality_range(DatastoreClient $datastore)
+{
+    // [START datastore_array_value_inequality_range]
+    $query = $datastore->query()
+        ->kind('Task')
+        ->filter('tag', '>', 'learn')
+        ->filter('tag', '<', 'math');
+    // [END datastore_array_value_inequality_range]
+    // return $query;
```
bshaffer, line 37: nit: remove commented code
```suggestion
```

datastore/api/src/basic_query.php
```diff
@@ -0,0 +1,54 @@
...
+namespace Google\Cloud\Samples\Datastore;
+
+use Google\Cloud\Datastore\DatastoreClient;
+use Google\Cloud\Datastore\Query\Query;
+
+/**
+ * Create a basic Datastore query.
+ *
+ * @param DatastoreClient $datastore
+ *
```
bshaffer, line 27: nit: remove extra line. It seems like this is the case for all samples.
```suggestion
```

datastore/api/src/batch_delete.php
```diff
@@ -0,0 +1,38 @@
...
+ * Delete multiple Datastore entities with the given keys.
+ *
+ * @param DatastoreClient $datastore
+ * @param array<Key> $keys
+ */
+function batch_delete(DatastoreClient $datastore, array $keys)
+{
+    // [START datastore_batch_delete]
+    $datastore->deleteBatch($keys);
+    // [END datastore_batch_delete]
```
bshaffer, line 33: should this print something? Looks like it returns a [response body](https://cloud.google.com/datastore/reference/rest/v1/projects/commit#response-body) that could be used to do something like 

```php
printf('Deleted %s items', count($response['mutationResults']));

File: datastore/api/src/get_or_create.php [hunk 7]
```diff
@@ -0,0 +1,44 @@
...
+function get_or_create(DatastoreClient $datastore, EntityInterface $task)
+{
+    // [START datastore_transactional_get_or_create]
+    $transaction = $datastore->transaction();
+    $existed = $transaction->lookup($task->key());
+    if ($existed === null) {
+        $transaction->insert($task);
+        $transaction->commit();
+    }
+    // [END datastore_transactional_get_or_create]
```
bshaffer, line 39: for consistency should this have print statements as well?  Something like:

```php
print(($existed ? 'Created task ' : 'Fetched existing task ') . $task->key());
```

datastore/api/src/cursor_paging.php
```diff
@@ -0,0 +1,67 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Cloud\Samples\Datastore;
+
+use Google\Cloud\Datastore\DatastoreClient;
+use Google\Cloud\Datastore\Query\Query;
+
+// [START datastore_cursor_paging]
```
bshaffer, line 23: This one also seems misplaced - I think they both should be inside the `cursor_paging` function.

bshaffer (approved):
this looks great! A bunch of nits but otherwise, approved

datastore/api/src/batch_delete.php
shivgautam: We are checking for deletion in the test by another lookup. The response doesn't contain the mutation, at least not the one we get. Direct REST call might. I'll leave these as is. There are a few other this way - for transactions. I'll leave them as is as well.

datastore/api/src/transactional_retry.php
shivgautam: Fixed.

datastore/api/src/ancestor_query.php
shivgautam: Done

datastore/api/src/array_value_inequality_range.php
shivgautam: Good catch. Done.

datastore/api/src/basic_query.php
shivgautam: Done

datastore/api/src/get_or_create.php
shivgautam: Added a print statement for the fetched entity.

PR #1975, opened 2024-02-22
chore: Datastore Samples - Revert deleted file 

shivgautam:
Added src/functions/concepts.php back till the docs are updated to refer to the new sample files.


yash30201 (approved):
Approving to unblock the samples.

More context: Right now the Datastore samples are read from a single hard coded filename in the docs page. Shiv will figure out how to add the config in the docs page so that php samples get read from per sample file using the region tags. 

Also, since the context on this mono sample file is same as all the files containing single samples, we can omit adding their tests otherwise repo would have two sets of identical tests for datastore.

PR #1978, opened 2024-02-24
chore: Removing old datastore samples file

shivgautam:


shivgautam:
The snippet-bot-check seems incorrect. The links on the pages that the bot checker mentions, refer to the new paths. So no region tags should be removed.

shivgautam:
So a few auto-generated pages are still referring to the old sample file. This does not show up in the snippet-bot-check. I'll wait for those doc pages to be re-generated and then rerun the snippet bot checl

shivgautam:
And, it's all set now. It can now be reviewed.

vishwarajanand (approved):
Lgtm, validated in logs that tests for Datastore are running.

PR #1983, opened 2024-03-06
feat: Add publisher with compression enabled sample

yash30201:
Fixes b/227480725

### DIFF
--- a/pubsub/api/test/pubsubTest.php
+++ b/pubsub/api/test/pubsubTest.php
@@ -182,6 +182,22 @@ class PubSubTest extends TestCase
         $this->assertMatchesRegularExpression('/Message published with retry settings/', $output);
     }
 
+    public function testTopicMessageWithCompressionEnabled()
+    {
+        $topic = $this->requireEnv('GOOGLE_PUBSUB_TOPIC');
+
+        $output = $this->runFunctionSnippet('publisher_with_compression', [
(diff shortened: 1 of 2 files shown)


pubsub/api/test/pubsubTest.php
```diff
@@ -182,6 +182,22 @@ public function testTopicMessageWithRetrySettings()
...
+        $topic = $this->requireEnv('GOOGLE_PUBSUB_TOPIC');
+
+        $output = $this->runFunctionSnippet('publisher_with_compression', [
+            self::$projectId,
+            $topic,
+            'This is a test message',
+        ]);
+
+        $this->assertMatchesRegularExpression(
+            '/Published a compressed message of message ID: /',
```
ajupazhamayil, line 196: You can use contains string check, just a suggestion. this also looks good to me.

```
$this->assertStringContainsString('Published a compressed message of message ID:', $output);
```

ajupazhamayil (approved):
LGTM

pubsub/api/test/pubsubTest.php
yash30201: I saw that the `assertMatchesRegularExpression` method is used much more that the contains string thing so just went ahead with it.

But...your argument makes sense! Let's just write clean code. Updated the code.

Issue #1899, opened 2023-07-27
Create samples for storageinsights


JesseLovelace:
Please create the following samples for the Storage Insights API, using [this](https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8460) as a reference 

storageinsights_create_inventory_report_config
storageinsights_delete_inventory_report_config
storageinsights_edit_inventory_report_config
storageinsights_get_inventory_report_names
storageinsights_list_inventory_report_configs

saranshdhingra:
Hey @yash30201, please have a look at this.

PR #1988, opened 2024-04-03
feat(StorageInsights): Adding samples

vishwarajanand:
Fixes https://github.com/GoogleCloudPlatform/php-docs-samples/issues/1899

Depends on https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1987

Tests pass on prod:
```
➜  storageinsights git:(storage_insights) ✗ ../testing/vendor/bin/phpunit -c phpunit.xml.dist                                                         
PHPUnit 9.6.18 by Sebastian Bergmann and contributors.

Warning:       XDEBUG_MODE=coverage or xdebug.mode=coverage has to be set

.....                                                               5 / 5 (100%)

Time: 00:19.591, Memory: 10.00 MB

OK (5 tests, 7 assertions)
➜  storageinsights git:(storage_insights) ✗
```

I managed to create an insights report: `projects/the-book-club-337319/locations/us-west1/reportConfigs/02f2961c-0e8d-4339-937c-f0bda263447c`


storageinsights/src/create_inventory_report_config.php
```diff
@@ -0,0 +1,77 @@
...
+    string $sourceBucket,
+    string $destinationBucket
+): void {
+    $storageInsightsClient = new StorageInsightsClient();
+
+    $reportConfig = (new ReportConfig())
+        ->setDisplayName('Example inventory report configuration')
+        ->setFrequencyOptions((new FrequencyOptions())
+            ->setFrequency(FrequencyOptions\Frequency::WEEKLY)
+            ->setStartDate((new Date())
```
saranshdhingra, line 48: Is this the best way to set the day?
Does the interface accept any other format, which doesn't ask the user to make such verbose calls?

storageinsights/src/get_inventory_report_names.php
```diff
@@ -0,0 +1,51 @@
...
+    string $projectId,
+    string $bucketLocation,
+    string $inventoryReportConfigUuid
+): void {
+    $storageInsightsClient = new StorageInsightsClient();
+
+    $reportConfigName = $storageInsightsClient->reportConfigName($projectId, $bucketLocation, $inventoryReportConfigUuid);
+    $reportConfig = $storageInsightsClient->getReportConfig($reportConfigName);
+    $extension = $reportConfig->hasCsvOptions() ? 'csv' : 'parquet';
+    printf('You can use the Google Cloud Storage Client '
```
saranshdhingra, line 38: nit: Use `print` instead?

storageinsights/src/list_inventory_report_configs.php
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Google\Cloud\Samples\StorageInsights;
+
+# [START storageinsights_list_inventory_report_configs]
+use Google\Cloud\StorageInsights\V1\StorageInsightsClient;
+
+/**
+ * Example:
+ * ```
+ * list_inventory_report_configs($projectId, $location);
```
saranshdhingra, line 26: nit: Let's add this call for other samples too?

saranshdhingra (commented):
Just noticed that there is an error regarding a PHP extension in the logs, can you check it?

storageinsights/src/create_inventory_report_config.php
vishwarajanand: [This code pointer](https://github.com/googleapis/google-cloud-php/blob/ec3c0442772d165bdfecba8eaceb60aa3677d139/StorageInsights/src/V1/FrequencyOptions.php#L116-L119) requires it to be `\Google\Type\Date::class`
So, I think we can't pass `DateTime` or a `string` here.

vishwarajanand:
@saranshdhingra  seems the warning is due to missing extension `pdo_srv` which is not used in this library. I will try to fix that in a separate PR.

vishwarajanand:
@saranshdhingra  I raised a PR to fix the missing PHP extension warning:
https://github.com/googleapis/testing-infra-docker/pull/373

ajupazhamayil (approved):
LGTM, Thank you!

PR #1990, opened 2024-05-15
feat: sample for authenticating GAPIC clients with an API key

bshaffer:


auth/src/auth_cloud_apikey.php
```diff
@@ -0,0 +1,75 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/main/auth/README.md
+ */
+
+# [START auth_api_explicit]
```
Sita04, line 23: This region tag is already taken to demonstrate auth with service accounts.
Can this be modified to `auth_api_key_explicit` ?

Sita04 (commented):
Thanks for the snippet! Can we add a test to this for verification?

Sita04 (approved):

PR #2012, opened 2024-06-03
chore: upgrade bigtable samples to v2

bshaffer:
depends on https://github.com/googleapis/google-cloud-php/pull/7317/

### DIFF
--- a/bigtable/composer.json
+++ b/bigtable/composer.json
@@ -1,7 +1,8 @@
 {
     "require": {
-        "google/cloud": "dev-bigtable-v2 as 2.0"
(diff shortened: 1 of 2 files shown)


bigtable/composer.json
```diff
@@ -1,7 +1,8 @@
 {
     "require": {
-        "google/cloud-bigtable": "^1.30"
+        "google/cloud-bigtable": "^2.0"
     },
+    "minimum-stability": "RC",
```
bshaffer, line 5: This is unnecessary once we remove the `RC1` from bigtable
```suggestion
```

PR #2021, opened 2024-07-01
chore: upgrade storageinsights to new surface

bshaffer:


bshaffer:
FYI @vishwarajanand I upgrade these to the new surface clients

PR #2037, opened 2024-07-25
feat(spanner): add samples for instance partitions

varuncnaik:
Add samples for instance partitions.

### DIFF
--- a/spanner/test/spannerTest.php
+++ b/spanner/test/spannerTest.php
@@ -50,6 +50,12 @@ class spannerTest extends TestCase
     /** @var string lowCostInstanceId */
     protected static $lowCostInstanceId;
 
+    /** @var string instancePartitionInstanceId */
+    protected static $instancePartitionInstanceId;
+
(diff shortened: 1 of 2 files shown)


spanner/test/spannerTest.php
```diff
@@ -50,6 +50,12 @@ class spannerTest extends TestCase
     /** @var string lowCostInstanceId */
     protected static $lowCostInstanceId;
 
+    /** @var string instancePartitionInstanceId */
+    protected static $instancePartitionInstanceId;
+
+    /** @var Instance instancePartitionInstance */
+    protected static $instancePartitionInstance;
```
bshaffer, line 57: nit, but you can typehint these directly now
```suggestion
    protected static Instance $instancePartitionInstance;
```

bshaffer (approved):
amazing job!

The tests are failing, but they're failing on a different test (`testDmlReturningUpdate`), while the test for this sample is passing. So I'll have to take a look at that other one later, but this one is good to go.

PR #2058, opened 2024-10-04
Use newer runtimes and let nginx serve static files

SarthakAjmera26:


compute/firewall/src/print_firewall_rule.php
kennethye1, line 1: why is this file even included in the PR?

SarthakAjmera26: To fix the failing presubmits. Check out these logs: https://github.com/GoogleCloudPlatform/php-docs-samples/actions/runs/11180011972/job/31080841794#step:4:65 . The said job is no longer failing after I made the changes.

SarthakAjmera26: I have removed the files which were unrelated to appengine samples from this change.

bshaffer (approved):

PR #2061, opened 2024-11-15
feat: update apikey sample for client option support

bshaffer:
Sample for https://github.com/googleapis/gax-php/pull/351


bshaffer:
note: kokoro tests are not running due to an ACL failure, but I have run the tests locally and confirmed they pass

PR #2078, opened 2025-04-07
Feat(Storage): add samples for support MoveObject

thiyaguk09:
add samples and test case for support MoveObject


thiyaguk09:
@bshaffer As the [feature](https://github.com/googleapis/google-cloud-php/pull/8213) has now been released, this PR is fully available for your review.

thiyaguk09:
@bshaffer  Just a friendly nudge to review that PR when you get a chance. Thanks! 😊

bshaffer:
@thiyaguk09 if you ping me on here and don't hear back, please message me directly on chat, as it's very easy for spammy github emails to get lost.

bshaffer (approved):

bshaffer:
@thiyaguk09 question - why didn't we update the sample in `move_object.php`? It doesn't makes sense to me to keep both of them.

thiyaguk09:
> @thiyaguk09 question - why didn't we update the sample in `move_object.php`? It doesn't makes sense to me to keep both of them.

It was a valid point to update `move_object.php`. 

Currently, `move_object.php` refers to the move feature, which is a manual operation involving the copy and delete features. The new version automates the copy and delete operations in the backend. In the end, both perform the same task. 

In other languages like Node, Ruby, and .NET, we are asked to keep both features and samples. That's why the same set of things is implemented here as well.

PR #2085, opened 2025-05-16
feat(storage): add samples for soft delete (objects)

thiyaguk09:
### Added Storage Samples:

- [ ] List Soft Deleted Objects
- [ ] List Versions of Soft Deleted Objects
- [ ] Restore Deleted Objects
- [ ] Disable Soft Delete Policy
- [ ] Get Soft Delete Policy
- [ ] Set Soft Delete Policy

### DIFF
--- a/storage/src/get_soft_delete_policy.php
+++ b/storage/src/get_soft_delete_policy.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Copyright 2025 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * For instructions on how to run the full sample:
+ *
+ * @see https://github.com/GoogleCloudPlatform/php-docs-samples/tree/main/storage/README.md
+ */
+
+namespace Google\Cloud\Samples\Storage;
+
+# [START storage_get_soft_delete_policy]
+use Google\Cloud\Storage\StorageClient;
+
+/**
+ * Gets a bucket soft delete policy.
+ *
+ * @param string $bucketName The name of your Cloud Storage bucket.
+ *        (e.g. 'my-bucket')
+ */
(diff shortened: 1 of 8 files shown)


storage/src/set_soft_delete_policy.php
```diff
@@ -0,0 +1,50 @@
...
+ * @param string $bucketName The name of your Cloud Storage bucket.
+ *        (e.g. 'my-bucket')
+ */
+function set_soft_delete_policy(string $bucketName): void
+{
+    $storage = new StorageClient();
+    $bucket = $storage->bucket($bucketName);
+    $bucket->update([
+        'softDeletePolicy' => [
+            'retentionDurationSeconds' => 864000,
```
TimothyLiuAtGoogle, line 41: What's the difference between the value `864000` and `604800s` (with a `s` suffix as below)?

thiyaguk09: You're absolutely right — the value `864000` is an integer, while `604800s` (with the `s` suffix) is a string. My bad on that — we’re using integers only here. I've updated it accordingly. Thanks for pointing it out!

storage/src/get_soft_delete_policy.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    $storage = new StorageClient();
+    $bucket = $storage->bucket($bucketName);
+    $bucket->reload();
+
+    if ($bucket->info()['softDeletePolicy']['retentionDurationSeconds'] === '0') {
+        printf('Bucket %s soft delete policy was disabled' . PHP_EOL, $bucketName);
+    } else {
+        printf('Soft delete Policy for ' . $bucketName . PHP_EOL);
+        printf('Soft delete Period: %d seconds' . PHP_EOL, $bucket->info()['softDeletePolicy']['retentionDurationSeconds']);
```
bshaffer, line 45:
```suggestion
        printf(
            'Soft delete Period: %d seconds' . PHP_EOL,
            $bucket->info()['softDeletePolicy']['retentionDurationSeconds']
        );
```

storage/src/get_soft_delete_policy.php
```diff
@@ -0,0 +1,56 @@
...
+        if ($bucket->info()['softDeletePolicy']['effectiveTime']) {
+            printf('Effective Time: ' . $bucket->info()['softDeletePolicy']['effectiveTime'] . PHP_EOL);
```
bshaffer, line 47:
```suggestion
            printf(
                'Effective Time: %s' . PHP_EOL,
                $bucket->info()['softDeletePolicy']['effectiveTime']
            );
```

storage/src/get_soft_delete_policy.php
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
```
bshaffer, line 50:
```suggestion
```

bshaffer (approved):

PR #2097, opened 2025-06-02
docs(PubSub): README improvements

cy-yun:
These variables are needed for testing multiple new samples.

### DIFF
--- a/pubsub/api/test/pubsubTest.php
+++ b/pubsub/api/test/pubsubTest.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Copyright 2016 Google Inc.
  *
@@ -32,6 +33,8 @@ class PubSubTest extends TestCase
     use EventuallyConsistentTestTrait;
 
     private static $eodSubscriptionId;
+    private static $awsRoleArn = 'arn:aws:iam::111111111111:role/fake-role-name';
+    private static $gcpServiceAccount = 'fake-service-account@project.iam.gserviceaccount.com';
 
     public static function setUpBeforeClass(): void
     {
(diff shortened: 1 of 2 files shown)


pubsub/api/test/pubsubTest.php
```diff
@@ -32,10 +33,17 @@ class PubSubTest extends TestCase
...
+    private static $gcpServiceAccount;
 
     public static function setUpBeforeClass(): void
     {
         self::$eodSubscriptionId = 'test-eod-subscription-' . rand();
+        $serviceAccountPath = self::requireEnv('GOOGLE_APPLICATION_CREDENTIALS');
+        $serviceAccountCredentials = new ServiceAccountCredentials(
+            'https://www.googleapis.com/auth/cloud-platform',
+            $serviceAccountPath);
+        self::$gcpServiceAccount = $serviceAccountCredentials->getClientName();
```
bshaffer, line 46: This shouldn't be needed, since the service will be found automatically from `GOOGLE_APPLICATION_CREDENTIALS` thanks to Application Default Credentials

cy-yun: I tried this. I'm getting "Uncaught Error: Cannot use object of type stdClass as array" on the second line.

pubsub/api/test/pubsubTest.php
```diff
@@ -17,6 +17,7 @@
 
 namespace Google\Cloud\Samples\PubSub;
 
+use Google\Cloud\Core\Exception\ServiceException;
```
cy-yun, line 20: this import statement isn't being used in this change but is needed for multiple tests that depend on this parent commit. Adding it here to avoid merge conflicts later on

bshaffer (approved):

PR #2098, opened 2025-06-02
feat(PubSub): Add update_topic_type sample

cy-yun:
b/322221088

### DIFF
--- a/pubsub/api/src/update_topic_type.php
+++ b/pubsub/api/src/update_topic_type.php
@@ -49,7 +49,9 @@ function update_topic_type(
         'projectId' => $projectId,
     ]);
 
-    $topic = $pubsub->topic($topicName, [
+    $topic = $pubsub->topic($topicName);
+
+    $topic->update([
         'ingestionDataSourceSettings' => [
             'aws_kinesis' => [
                 'stream_arn' => $streamArn,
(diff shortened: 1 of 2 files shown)


pubsub/api/src/update_topic_type.php
```diff
@@ -0,0 +1,71 @@
...
+                'consumer_arn' => $consumerArn,
+                'aws_role_arn' => $awsRoleArn,
+                'gcp_service_account' => $gcpServiceAccount
+            ]
+        ]
+    ]);
+
+    $topic->update([], [
+        'updateMask' => [ 'ingestionDataSourceSettings' ]
+    ]);
```
bshaffer: This is a valid way to do it, but I think when _updating_, the API would normally look a bit more like this:

```suggestion
    $topic = $pubsub->topic($topicName);
    $topic->update([
        'ingestionDataSourceSettings' => [
            'aws_kinesis' => [
                'stream_arn' => $streamArn,
                'consumer_arn' => $consumerArn,
                'aws_role_arn' => $awsRoleArn,
                'gcp_service_account' => $gcpServiceAccount
            ]
        ]
    ], [
        'updateMask' => [
            'ingestionDataSourceSettings'
        ]
    ]);

```

Then the `update` function will calculate the update mask for you.

bshaffer (approved):

PR #2101, opened 2025-06-03
fix(PubSub): Add locational endpoint to SubscribeExactlyOnceDelivery sample

cy-yun:
https://buganizer.corp.google.com/issues/361183840

### DIFF
--- a/pubsub/api/src/subscribe_exactly_once_delivery.php
+++ b/pubsub/api/src/subscribe_exactly_once_delivery.php
@@ -38,6 +38,8 @@ function subscribe_exactly_once_delivery(
 ): void {
     $pubsub = new PubSubClient([
         'projectId' => $projectId,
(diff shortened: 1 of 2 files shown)


pubsub/api/src/subscribe_exactly_once_delivery.php
```diff
@@ -38,6 +38,7 @@ function subscribe_exactly_once_delivery(
 ): void {
     $pubsub = new PubSubClient([
         'projectId' => $projectId,
+        'apiEndpoint' => 'us-west1-pubsub.googleapis.com:443'
```
bshaffer, line 41: Hmm, maybe a comment to explain what this is for? just something like
```suggestion
        // use the "apiEndpoint" option to set a regional endpoint
        'apiEndpoint' => 'us-west1-pubsub.googleapis.com:443'
```

PR #2103, opened 2025-06-03
feat(PubSub): Add CreateUnwrappedPushSubscription sample

cy-yun:
https://buganizer.corp.google.com/issues/285019494

### DIFF
--- a/pubsub/api/src/create_unwrapped_push_subscription.php
+++ b/pubsub/api/src/create_unwrapped_push_subscription.php
@@ -1,0 +1,57 @@
+<?php
+
+/**
+ * Copyright 2025 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


pubsub/api/src/create_unwrapped_push_subscription.php
```diff
@@ -0,0 +1,57 @@
...
+): void {
+
+    $pubsub = new PubSubClient([
+        'projectId' => $projectId,
+    ]);
+    $pubsub->subscribe($subscriptionId, $topicName, [
+        'pushConfig' => [
+            'no_wrapper' => [
+                'write_metadata' => true
+            ]
```
bshaffer, line 50: so, it's a bit confusing, but we accept both snake and camel-case for fields which are serialized into protobuf messages from handwritten client libraries

I am a bit torn because on one hand, we use camelCase across handwritten and GAPIC libraries for all options. On the other hand, protobuf uses snake_case for everything. However, I think it's important that we stay consistent, and for now I prefer camelCase to snake_case.

```suggestion
            'noWrapper' => [
                'writeMetadata' => true
            ]
```

cy-yun: Sounds good. Do you want this change to be applied for other tests as well?

bshaffer (approved):

PR #2143, opened 2025-07-03
samples(StorageControl): Add sample for anywhereCache

thiyaguk09:
This PR adds sample code for the anywhere cache feature, allowing users to:

- [ ] Create anywhere cache
- [ ] Get anywhere cache
- [ ] List anywhere cache
- [ ] Update anywhere cache
- [ ] Pause anywhere cache
- [ ] Disable anywhere cache
- [ ] Resume anywhere cache

### DIFF
--- a/storagecontrol/README.md
+++ b/storagecontrol/README.md
@@ -58,7 +58,3 @@ to [browse the source][google-cloud-php-source] and  [report issues][google-clou
 ## Licensing
 
 * See [LICENSE](../../LICENSE)
-
-## Note
-
-Tests for Anywhere cache Sample are not included due to the long operation times typical for cache operations.


thiyaguk09:
@bshaffer  Just a reminder—when you have time, please take a look at this PR.

storagecontrol/src/get_anywhere_cache.php
```diff
@@ -0,0 +1,58 @@
...
+ * @param string $bucketName The name of your Cloud Storage bucket.
+ *        (e.g. 'my-bucket')
+ * @param string $anywhereCacheId Uniquely identifies the cache.
+ *        (e.g. 'us-east1-b')
+ */
+function get_anywhere_cache(string $bucketName, string $anywhereCacheId): void
+{
+    $storageControlClient = new StorageControlClient();
+
+    // Set project to "_" to signify global bucket
```
bshaffer, line 43: is this standard? I've never seen this before - I thought we specified the string `global`.

thiyaguk09: Yes, In the error response, we receive the following comment, and in other language samples like C++ and Java, only '_' is used. [Reference](https://cloud.google.com/storage/docs/using-anywhere-cache#create-cache
)


`"message": "Invalid name 'projects\/projectID\/buckets\/sample-bucket'. The resource name must begin with 'projects\/_', with an underscore as the segment after projects.",
"reason": "GRPC_PROJECT_NOT_UNDERSCORE_INVALID_BUCKET_NAME"`

storagecontrol/src/create_anywhere_cache.php
```diff
@@ -0,0 +1,67 @@
...
+        'parent' => $formattedName,
+        'anywhere_cache' => $anywhereCache,
+    ]);
+
+    // Start a create operation and block until it completes. Real applications
+    // may want to setup a callback, wait on a coroutine, or poll until it
+    // completes.
+    $response = $storageControlClient->createAnywhereCache($request);
+
+    printf('Created anywhere cache: %s', $response->getName());
```
bshaffer, line 61: these responses are operations (because they are longrunning), so this is incorrect. 

I would suggest either calling the `$response` variable an `$operation`, or waiting for it to complete like this:

```php
$operation->pollUntilComplete();
$cache = $operation->getResult();

printf('Created anywhere cache: %s', $cache->getName());


```

bshaffer (changes requested):

storagecontrol/src/create_anywhere_cache.php
bshaffer: I ran these tests and they all passed. Great job!

bshaffer (approved):

